PR #29, opened 2020-06-10
CI: Increases the default optimizer-ortool version to use ortool v7.5

alain-andre:
@fonsecadeline @braktar This PR increases optimizer-ortools version to use ortools v7.5


fonsecadeline:
This infortunately does not seem to be enough to fix our problematic test.

senhalil:
@alain-andre we use 7.5 on dev/beta, not 7.4.

senhalil:
I checked the 2 failing tests: The one in real cases is already fixed a month or two ago but it is not merged to dev yet. It is due to performance change in or-tools (it was depending on the old "iterations without improvement" limit). So it is not due to CI settings but will be fixed by this commit

The other fail is due to ortools version, we suspect because @braktar implemeted this feature with 7.5 and this test passes on @fonsecadeline 's Travis with version 7.5

senhalil:
@alain-andre  this is the fix for the second failing test https://github.com/senhalil/optimizer-api/commit/6d9ff6531258981f0700277d7e42a631f2e5e14c

Can you add it to this pull request and the other failing test will be fixed when you increase 7.4 to 7.5

PR #31, opened 2020-06-16
Small fix + Quality changes

fonsecadeline:
@fab-girard can you please check changes on CSV output in particular?

### DIFF
--- a/models/vrp.rb
+++ b/models/vrp.rb
@@ -238,9 +238,20 @@ module Models
       end
     end
 
+    def self.convert_route_indice_into_index(hash)
+      hash[:routes].to_a.each{ |route|
+        next unless route[:indice]
+
+        log 'Route indice was used instead of route index', level: :warn
+        route[:index] = route[:indice]
+        route.delete(:indice)
+      }
+    end
+
     def self.ensure_retrocompatibility(hash)
       self.convert_position_relations(hash)
       self.deduce_solver_parameter(hash)
+      self.convert_route_indice_into_index(hash)
     end
 
     def self.filter(hash)
@@ -281,18 +292,7 @@ module Models
     end
 
     def self.generate_schedule_indices_from_date(hash)
-      return hash if !hash[:configuration]
-
-      if hash[:configuration][:preprocessing] && hash[:configuration][:preprocessing][:partitions].to_a.size.positive?
-        hash[:routes]&.each{ |route|
-          route[:mission_ids].each{ |id|
-            corresponding = [hash[:services], hash[:shipments]].compact.flatten.find{ |s| s[:id] == id }
-            corresponding[:sticky_vehicle_ids] = [route[:vehicle_id]]
-          }
-        }
-      end
-
-      return hash if !hash[:configuration][:schedule] ||
+      return hash if !hash[:configuration] || !hash[:configuration][:schedule] ||
                      hash[:configuration][:schedule][:range_indices]
 
       start_indice = hash[:configuration][:schedule][:range_date][:start].to_date.cwday - 1
@@ -350,9 +350,9 @@ module Models
 
       # convert route dates into indices
       hash[:routes]&.each{ |route|
-        next if route[:indice]
+        next if route[:index]
 
-        route[:indice] = (route[:date].to_date - hash[:configuration][:schedule][:range_date][:start].to_date).to_i + start_indice
+        route[:index] = (route[:date].to_date - hash[:configuration][:schedule][:range_date][:start].to_date).to_i + start_indice
         route.delete(:date)
       }
 
--- a/optimizer_wrapper.rb
+++ b/optimizer_wrapper.rb
@@ -233,6 +233,7 @@ module OptimizerWrapper
       vrp.compute_matrix(&block)
 
       sub_unfeasible_services = config[:services][service].check_distances(vrp, sub_unfeasible_services)
+      vrp.clean_according_to(sub_unfeasible_services)
 
       vrp = config[:services][service].simplify_constraints(vrp)
 
--- a/optimizer_wrapper.rb
+++ b/optimizer_wrapper.rb
@@ -256,8 +256,7 @@ module OptimizerWrapper
       unfeasible_services += sub_unfeasible_services
       if vrp.resolution_solver && !vrp.preprocessing_first_solution_strategy.to_a.include?('periodic')
         # TODO: Move select best heuristic in each solver
-        block&.call(nil, nil, nil, "process heuristic choice : #{vrp.preprocessing_first_solution_strategy}", nil, nil, nil) if vrp.preprocessing_first_solution_strategy&.include?('self_selection') || vrp.preprocessing_first_solution_strategy&.size.to_f > 1
-        Interpreters::SeveralSolutions.custom_heuristics(service, vrp)
+        Interpreters::SeveralSolutions.custom_heuristics(service, vrp, block)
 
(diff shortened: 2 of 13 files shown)


models/vrp.rb
```diff
@@ -350,9 +362,9 @@ def self.generate_schedule_indices_from_date(hash)
 
       # convert route dates into indices
       hash[:routes]&.each{ |route|
-        next if route[:indice]
+        next if route[:indice] || route[:index]
```
senhalil, line 365: If the retrocompatibility function is called before this function then we don't need the `route[:indice] ||` part. 

If not, we should move the `convert_route_indice_into_index` earlier. 

After a parameter is deprecated, it should not appear anywhere in the API except the location where we handle "switch"

optimizer_wrapper.rb
```diff
@@ -491,16 +491,22 @@ def self.job_remove(api_key, id)
...
     elsif activity['point_id']
-      'store'
+      I18n.t('export_file.type_store')
     end
   end
 
+  def self.basic_header
+    [I18n.t('export_file.route'),
+     I18n.t('export_file.activity_id'), I18n.t('export_file.activity_point_id'), I18n.t('export_file.lat'), I18n.t('export_file.lon'), I18n.t('export_file.activity_type'), I18n.t('export_file.setup_duration'), I18n.t('export_file.duration'),
+     I18n.t('export_file.additional_value'), I18n.t('export_file.skills'), I18n.t('export_file.tags'), I18n.t('export_file.total_travel_time'), I18n.t('export_file.total_travel_distance')]
```
senhalil, line 505:
```suggestion
    %w[
      route
      activity_id
      activity_point_id
      lat
      lon
      activity_type
      setup_duration
      duration
      additional_value
      skills
      tags
      total_travel_time
      total_travel_distance
    ].collect{ |label| I18n.t("export_file.#{label}") }
```

I think like this it is easier to add/remove labels and keep count of them and their order (otherwise it is easier to make mistakes while counting multiple items per line), what do you think?

test/real_cases_test.rb
```diff
@@ -228,7 +228,7 @@ def test_ortools_global_with_identical_route_duration_and_vehicle_window
       assert_equal 29, (result[:routes].count{ |r| r[:activities].count{ |a| a[:service_id] }.positive? })
 
       # Check total travel time
-      assert result[:routes].map{ |r| r[:total_travel_time] }.reduce(&:+) < 176000, "Too long travel time:#{result[:routes].map{ |r| r[:total_travel_time] }.reduce(&:+)}"
+      assert result[:routes].map{ |r| r[:total_travel_time] }.reduce(&:+) < 178000, "Too long travel time:#{result[:routes].map{ |r| r[:total_travel_time] }.reduce(&:+)}"
```
senhalil, line 231: Did you have an issue with this test @fonsecadeline ? 

We haven't touch optim-ortools recently, it is weird :thinking: @braktar do you have any problems with it?

config/locales/en.yml
```diff
@@ -1,4 +1,4 @@
-# Copyright © Mapotempo, 2016
+# Copyright © Mapotempo, 2020
```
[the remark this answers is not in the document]
fonsecadeline: Since file was empty I assumed I should change year but I guess we can keep 2016.

models/vrp.rb
```diff
@@ -238,9 +238,20 @@ def self.deduce_solver_parameter(hash)
...
+        log 'Route indice was used instead of route index', level: :warn
+        route[:index] = route[:indice]
+        route.delete(:indice)
+      }
+    end
+
     def self.ensure_retrocompatibility(hash)
       self.convert_position_relations(hash)
       self.deduce_solver_parameter(hash)
+      self.convert_route_indice_into_index(hash)
```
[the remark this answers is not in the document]
fonsecadeline: Sure, they are all related to deprecated parameters.

config/locales/en.yml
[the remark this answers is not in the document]
senhalil: Yeah file creation date is the important one because it demonstrates that "you already had this in mind". To be honest, I have no knowledge about copyrights but I know this one because I made a similar change which was rejected due to this rule :)

test/real_cases_test.rb
fonsecadeline: It appeared when I removed vehicles useless sort but I could not figure out why...

models/vrp.rb
```diff
@@ -238,9 +238,20 @@ def self.deduce_solver_parameter(hash)
       end
     end
 
+    def self.convert_route_indice_into_index(hash)
+      hash[:routes].to_a.each{ |route|
+        next unless route[:indice]
```
fonsecadeline, line 243:
```suggestion
        next unless route[:index]
```

test/real_cases_test.rb
braktar: OR-Tools v7.5+ find a solution one less unassigned. which increase the total_travel_time.
So that's ok to increase this value. But additonnaly, the total number of unassigned must be decreased accordingly. Maybe with a note to indicate that the new conditions only pass with OR-Tools v7.5+

fonsecadeline: I don't think the reason here is not related to OR-Tools version, but really to vehicles order. 

Indeed, at previous commit we get same number of unassigned stops but with a lower travel_time.

Every vehicle is identical except from skills. I think first solution strategy is vehicles order dependent and can slightly change solution.

fonsecadeline: Ok :) (that would rather be <= 30 for 7.4 and 7.1, <= 29 for 7.5, I will use 30).

fonsecadeline: Well, in fact there already is a check for unassigned number. How can it be 1 unassigned more with 7.4 ? :p

fonsecadeline: oups, sorry, I did everything at the opposite, 29 is number of assigned stops.

fonsecadeline: Still, checking number of stops in route is equivalent to checking number of unassigned.

senhalil: That's the number of routes I think.

senhalil: If there are more than one service in each vehicle, then checking the number of vehicles is not enough.

fonsecadeline: Yep, I looked too fast :)

test/real_cases_test.rb
```diff
@@ -219,16 +219,17 @@ def test_ortools_global_with_identical_route_duration_and_vehicle_window
       assert(result[:unassigned].one? { |unassigned| unassigned[:service_id] == 'service84' })
       assert(result[:unassigned].one? { |unassigned| unassigned[:service_id] == 'R1169' })
       assert(result[:unassigned].one? { |unassigned| unassigned[:service_id] == 'R1183' })
+      assert_operator result[:unassigned].size, :<=, 6
 
-      assert_operator(check_vrp_services_size - 6, :<=, result[:routes].sum{ |r| r[:activities].count{ |a| a[:service_id] } })
+      assert_operator(check_vrp_services_size - result[:unassigned].size, :<=, result[:routes].sum{ |r| r[:activities].count{ |a| a[:service_id] } })
```
senhalil, line 224: We can remove this check completely because the next condition checks it.

config/environments/development.rb
```diff
@@ -51,8 +51,8 @@ module OptimizerWrapper
   OptimizerLogger.with_datetime = true
   OptimizerLogger.caller_location = :relative
 
-  I18n.available_locales = [:en, :fr]
-  I18n.default_locale = :en
+  I18n.available_locales = [:en, :fr, :unspecified]
```
fab-girard, line 54: avoid using `:unspecified`

fonsecadeline:
@fab-girard can you please merge this request ?

PR #33, opened 2020-06-18
[OO] Minimum duration lapse and improve when shipment in relation

fonsecadeline:


fonsecadeline:
Les tests cassent car les commits correspondants sur optimizer-ortools ne sont pas sur mapotempo/dev

senhalil:
We need to wait for that PR to be merged then :thinking:

PR #35, opened 2020-06-23
sentry fixes + tests + output corrections

senhalil:


lib/interpreters/compute_several_solutions.rb
```diff
@@ -202,7 +203,7 @@ def self.find_best_heuristic(service_vrp)
         vrp.resolution_batch_heuristic = nil
         vrp.preprocessing_first_solution_strategy = [best_heuristic]
         vrp.preprocessing_heuristic_synthesis = synthesis
-        vrp.resolution_duration = vrp.resolution_duration ? (vrp.resolution_duration - times.sum).floor : nil
+        vrp.resolution_duration = vrp.resolution_duration ? [(vrp.resolution_duration.to_f * (1 - percent_allocated_to_heur_selection)).round, 1000].max : nil
```
fonsecadeline, line 206: I agree this fixes the case where times.sum is a little higher than vrp.resolution_duration.

But we are slightly changing behavior : if we did not use all _total_time_allocated_for_heuristic_selection_ to select best heuristic then we do not use spare time for resolution like we used to.

Is it voluntary ? Could we rather consider :
`vrp.resolution_duration = [vrp.resolution_duration ? (vrp.resolution_duration - times.sum).floor, 1000].max : nil` ?

PR #37, opened 2020-07-02
Various edits

braktar:


api/v01/entities/vrp_result.rb
```diff
@@ -44,6 +44,13 @@ class VrpResultSolutionRouteActivityDetails < Grape::Entity
       expose :timewindows, using: VrpResultSolutionRouteActivityDetailTimewindows, documentation: { is_array: true, desc: '' }
     end
 
+    class VRPResultSolutionRouteCosts < Grape::Entity
+      expose :fixed, documentation: { type: String, desc: 'Cost associated to the use of the vehicle' }
+      expose :time, documentation: { type: String, desc: 'Cost associated to the time dimension' }
+      expose :distance, documentation: { type: String, desc: 'Cost associated to the distance dimension' }
+      expose :value, documentation: { type: String, desc: 'Cost associated to the value dimension' }
+    end
```
senhalil, line 52: We don't want the total cost for ease of comparison (otherwise comparing two results becomes a bit of a chore) ?

api/v01/entities/vrp_result.rb
```diff
@@ -66,6 +73,7 @@ class VrpResultSolutionRoute < Grape::Entity
       expose :end_time, documentation: { type: Integer, desc: 'Give the actual end time of the current route if provided by the solver' }
       expose :geometry, documentation: { type: String, desc: 'Contains the geometry of the route, if asked in first place' }
       expose :initial_loads, using: VrpResultDetailQuantities, documentation: { is_array: true, desc: 'Give the actual initial loads of the route' }
+      expose :costs, using: VRPResultSolutionRouteCosts, documentation: { desc: 'The impact of the current route within the solution cost' }
```
senhalil, line 76: Do we want this cost for each activity or each route?

The costs does not make much sense for each activity (for example distance cost will be less for services clump together but it will be high for the first and the last services (depart/arrive depot) -- which affects the time cost as well). Again fixed cost is a cost that is shared by each activity on the route.

wrappers/ortools.rb
```diff
@@ -75,6 +75,7 @@ def solver_constraints
         :assert_route_date_or_indice_if_periodic,
         :assert_not_too_many_visits_in_route,
         :assert_no_route_if_schedule_without_periodic_heuristic,
+        # :assert_no_overall_duration, # Requires a complete rework
```
senhalil, line 78: So that if "one day" we attack the todo's it'll be visible to make us unhappy :)  
```suggestion
        # :assert_no_overall_duration, # TODO: Requires a complete rework
```

wrappers/wrapper.rb
```diff
@@ -727,7 +732,6 @@ def check_distances(vrp, unfeasible)
 
         add_unassigned(unfeasible, vrp, service, 'Service cannot be reached within its timewindows') if service_unreachable_within_its_tw
       }
-
       log "Following services marked as infeasible:\n#{unfeasible.group_by{ |u| u[:reason] }.collect{ |g, set| "#{(set.size < 20) ? set.collect{ |s| s[:service_id] }.join(', ') : "#{set.size} services"}\n with reason '#{g}'" }.join("\n")}", level: :debug unless unfeasible.empty?
```
senhalil, line 735: Can we get the space back please? It is my favorite space :) 

Jokes aside it is better to see the end of the loop without clutter for rapid comprehension.

api/v01/entities/vrp_result.rb
```diff
@@ -44,6 +44,13 @@ class VrpResultSolutionRouteActivityDetails < Grape::Entity
       expose :timewindows, using: VrpResultSolutionRouteActivityDetailTimewindows, documentation: { is_array: true, desc: '' }
     end
 
+    class VRPResultDetailCosts < Grape::Entity
```
senhalil, line 47: I think instead of  `VRPResultDetailCosts` we can use one of the following. `DetailCosts` sounds wrong
```suggestion
    class VRPResultDetailedCosts < Grape::Entity
```
```suggestion
    class VRPResultCostDetails < Grape::Entity
```

api/v01/entities/vrp_result.rb
senhalil: @braktar  it looks like, you forgot to add the `total`

wrappers/ortools.rb
senhalil: I have the impression that you pushed before finishing the edits :D

wrappers/wrapper.rb
```diff
@@ -367,6 +367,11 @@ def assert_no_vehicle_overall_duration_if_heuristic(vrp)
       vrp.vehicles.none?{ |vehicle| vehicle[:overall_duration] } || vrp.preprocessing_first_solution_strategy.to_a.first != 'periodic'
     end
 
+    def assert_no_overall_duration(vrp)
+      vrp.vehicles.none?{ |vehicle| vehicle[:overall_duration] } &&
+        vrp.relations.none?{ |relation| ['vehicle_group_duration', 'vehicle_group_duration_on_weeks', 'vehicle_group_duration_on_months'].include?(relation.type) }
```
senhalil, line 372:
```suggestion
        vrp.relations.none?{ |relation| %w[vehicle_group_duration vehicle_group_duration_on_weeks vehicle_group_duration_on_months].include?(relation.type) }
```
I don't know if it is better in terms of perf but it is definitely more ruby syntax and it is easier to modify

api/v01/entities/vrp_result.rb
fonsecadeline: I like _VRPResultCostDetails_ more. It sounds more meaningfull for me.

PR #41, opened 2020-07-17
Fix generate rests

fonsecadeline:


lib/interpreters/periodic_visits.rb
```diff
@@ -513,24 +513,24 @@ def generate_routes(vrp)
     end
 
     def generate_rests(vehicle, day_index, rests_durations)
-      new_rests = vehicle.rests.collect{ |rest|
+      vehicle.rests.collect{ |rest|
         new_rest = nil
-        if rest.timewindows.empty? || rest.timewindows.any?{ |timewindow| timewindow[:day_index] == day_index }
+        if rest.timewindows.empty? || rest.timewindows.any?{ |timewindow| timewindow.day_index.nil? || timewindow.day_index == day_index }
```
senhalil, line 518: may this is missing a modulo? because `timewindow.day_index` (i don't know if this is true for `day_index`)

lib/interpreters/periodic_visits.rb
```diff
@@ -513,24 +513,24 @@ def generate_routes(vrp)
...
                 {
-                  id: ("#{timewindow[:id]} #{day_index}" if timewindow[:id] && !timewindow[:id].nil?),
-                  start: timewindow[:start] ? timewindow[:start] + day_index.to_i * 86400 : nil,
-                  end: timewindow[:end] ? timewindow[:end] + day_index.to_i * 86400 : nil
+                  id: ("#{timewindow.id} #{day_index}" if timewindow.id),
+                  start: timewindow.start ? timewindow.start + day_index.to_i * 86400 : nil,
+                  end: timewindow.end ? timewindow.end + day_index.to_i * 86400 : nil
                 }.delete_if { |_k, v| !v }
             end
           }.flatten.sort_by{ |tw| tw[:start] }.compact.uniq
```
senhalil, line 536:
```suggestion
          }.flatten.sort_by(&:start).compact.uniq
```
the forgotten one :)

lib/interpreters/periodic_visits.rb
```diff
@@ -513,24 +513,24 @@ def generate_routes(vrp)
...
+                start: timewindow.start ? timewindow.start + timewindow.day_index * 86400 : nil,
+                end: timewindow.end ? timewindow.end + timewindow.day_index * 86400 : nil
               }.delete_if { |_k, v| !v }
             else
                 {
-                  id: ("#{timewindow[:id]} #{day_index}" if timewindow[:id] && !timewindow[:id].nil?),
-                  start: timewindow[:start] ? timewindow[:start] + day_index.to_i * 86400 : nil,
-                  end: timewindow[:end] ? timewindow[:end] + day_index.to_i * 86400 : nil
+                  id: ("#{timewindow.id} #{day_index}" if timewindow.id),
+                  start: timewindow.start ? timewindow.start + day_index.to_i * 86400 : nil,
```
senhalil, line 532: Do we really need `to_i` here? Can `day_index` be nil?

lib/interpreters/periodic_visits.rb
senhalil: We need a test which passes through this function.

lib/interpreters/periodic_visits.rb
fonsecadeline: You are right ! If day_index is nil for start then it should be okay because day_index becomes 0 and start is as soon as possible.
But when day_index is nil for timewindow end then we are making a mistake...

fonsecadeline: In fact if day_index is nil we should generate as many timewindows as available days.

senhalil: This got more complex than I expected :D but yeah it looks like this needs a correction

fonsecadeline:
Replaced by : https://github.com/Mapotempo/optimizer-api/pull/55 because there were too many changes to make.

PR #43, opened 2020-07-20
Fix Travis stopping without output and test_delete_job

senhalil:
travis_wait doesn't show any output until the test finishes and sometimes even after test is finished, it doesn't show the output correctly (we cant see how many and which tests failed, etc...).

instead of travis_wait, this commit uses a small subprocess to output something every 60 seconds, launches the tests, kills the subprocess and returns the exist value of the test.


senhalil:
@alain-andre can you mainly check the `Travis: fix stopping without output` commit, the others are optim-api related..

ci-utils/tests.sh
```diff
@@ -43,4 +43,15 @@ done
...
+
+# Run the tests and register the exit value
+docker exec -it ${CONTAINER} bundle exec rake test ${TEST_ENV}
+test_exit_status=$?
+
+# Kill background sleep loop
+kill %1
+
+# Return the exit value of the tests
+exit $test_exit_status
```
alain-andre, line 57: Please add a new line at end of file.

test/api/v01/request_helper.rb
```diff
@@ -25,18 +25,23 @@ def app
   end
 
   def wait_status(job_id, status, options)
+    puts "#{job_id} #{Time.now} waiting #{status} status"
     loop do
       get "0.1/vrp/jobs/#{job_id}.json", options
       sleep 1
 
+      puts "Empty resonse body: #{JSON.parse(last_response.body)}" if JSON.parse(last_response.body).nil? || JSON.parse(last_response.body)['job'].nil?
```
fonsecadeline, line 33:
```suggestion
      puts "Empty response body: #{JSON.parse(last_response.body)}" if JSON.parse(last_response.body).nil? || JSON.parse(last_response.body)['job'].nil?
```

PR #46, opened 2020-07-21
Various edits

braktar:


test/lib/interpreters/split_clustering_test.rb
```diff
@@ -702,5 +702,34 @@ def test_basic_split
       result = OptimizerWrapper.wrapper_vrp('ortools', { services: { vrp: [:ortools] }}, TestHelper.create(problem), nil)
       assert result
     end
+
+    def test_scheduling_partitions_without_occurency
```
senhalil, line 706: What do we mean by `_without_occurency`, maybe `without_lapse` or `without_recurrence` might be better?

wrappers/ortools.rb
```diff
@@ -606,6 +627,7 @@ def parse_output(vrp, _services, points, _matrix_indices, _cost, _iterations, ou
           earliest_start = route_start
           {
             vehicle_id: vehicle.id,
+            costs: route_costs,
             activities: route['activities'].collect.with_index{ |activity, activity_index|
```
senhalil, line 631: It is better if you can replace all these `A['X']` with `A.X` if protobuf has the functions otherwise you break the uniformity which will bite us in the back in the future.

wrappers/ortools.rb
```diff
@@ -491,6 +492,20 @@ def build_rest(rest, day_index, job_load)
...
+      cost_hash = {
+         fixed: costs&.fixed || 0,
+         time: costs && (costs.time + costs.time_fake + costs.time_without_wait) || 0,
+         distance: costs && (costs.distance + costs.distance_fake) || 0,
+         value: costs&.value || 0,
+         lateness: costs&.lateness || 0,
+         overload: costs&.overload || 0
+      }
+      total = cost_hash.values.reduce(&:+)
+      cost_hash[:total] = total
```
senhalil, line 505: Didn't tested but this should work right?
```suggestion
      cost_hash[:total] = cost_hash.values.reduce(&:+)
```

lib/helper.rb
```diff
@@ -63,11 +63,32 @@ def self.euclidean_distance(loc_a, loc_b)
...
+    }
+    total_costs
+  end
+
   def self.merge_results(results, merge_unassigned = true)
     results.flatten!
     {
       solvers: results.flat_map{ |r| r && r[:solvers] }.compact,
       cost: results.map{ |r| r && r[:cost] }.compact.reduce(&:+),
+      costs: merge_costs(results.map{ |r| r && r[:costs] }.compact),
```
senhalil, line 91: Instead of a `merge_costs` function,  we should overload the `+=` and `+` operators for the `cost` class so that we can do `results.sum{ |r| r[:costs] }` to merge -- this is more idiomatic and we don't need to remember a function name.

test/lib/interpreters/split_clustering_test.rb
```diff
@@ -702,5 +702,34 @@ def test_basic_split
...
+        service.maximum_lapse = nil
+      }
+
+      vrp.schedule_range_indices = {
+        start: 0,
+        end: 6
+      }
+
+      result = OptimizerWrapper.wrapper_vrp('demo', { services: { vrp: [:demo] }}, vrp, nil)
+      assert result
```
fonsecadeline, line 732: Shouldn't we ensure we did not use generate_route function ? With stub for instance.

PR #47, opened 2020-07-24
Small fixes in multi_trips expand

fonsecadeline:


fonsecadeline:
Fix for : https://sentry.mapotempo.com/mapotempo/api-optimizer-beta/issues/12801/

lib/interpreters/multi_trips.rb
```diff
@@ -22,23 +22,31 @@ module Interpreters
...
-          (0..vehicle.trips - 1).collect{ |index|
-            new_vehicle = Marshal.load(Marshal.dump(vehicle))
-            new_vehicle.id += "_trip_#{index}"
-            new_vehicle.trips = 1
-            new_vehicle
-          }
-        else
-          vehicle
-        end
+        new_ids = (0..vehicle.trips - 1).collect{ |index| vehicle.id + '_trip_' + index.to_s }
```
senhalil, line 27:
```suggestion
        new_ids = vehicle.trips.times.collect{ |index| vehicle.id + '_trip_' + index.to_s }
```

models/vrp.rb
```diff
@@ -191,6 +191,7 @@ def self.check_consistency(hash)
     def self.expand_data(vrp)
       vrp.add_sticky_vehicle_if_routes_and_partitions
       vrp.adapt_relations_between_shipments
+      Interpreters::MultiTrips.new.expand(vrp)
```
senhalil, line 194:
```suggestion
```

senhalil: May be in the future we can move this logic out of `Interpreters::MultiTrips` into vrp` class because it is just transformation of the input. However while this logic is in the class ` Interpreters::MultiTrips`, I think it is better to keep it in the old location (unless there is a specific reason to not to) because all other expands are inside solve at the moment.

optimizer_wrapper.rb
```diff
@@ -190,8 +190,6 @@ def self.solve(service_vrp, job = nil, block = nil)
 
     tic = Time.now
 
-    Interpreters::MultiTrips.new.expand(vrp)
-
```
senhalil, line 194: (c/p) May be in the future we can move this logic out of Interpreters::MultiTrips into vrpclass because it is just transformation of the input. However while this logic is in the class Interpreters::MultiTrips`, I think it is better to keep it in the old location (unless there is a specific reason to not to) because all other expands are inside solve at the moment.

lib/interpreters/multi_trips.rb
```diff
@@ -22,23 +22,31 @@ module Interpreters
...
     def expand(vrp)
       vrp.vehicles = vrp.vehicles.collect{ |vehicle|
-        if vehicle.trips && vehicle.trips >= 1
-          new_ids = (0..vehicle.trips - 1).collect{ |index| vehicle.id + '_trip_' + index.to_s }
-          containing_relations = vrp.relations.select{ |relation| relation.linked_vehicle_ids.include?(vehicle.id) }.each{ |relation|
-            relation.linked_vehicle_ids -= vehicle.id
-            relation.linked_vehicle_ids += new_ids
-          }
-          vrp.relations += [Models::Relation.new(type: 'vehicle_trips', linked_vehicle_ids: new_ids)]
+        return vehicle unless vehicle.trips && vehicle.trips > 1
```
senhalil, line 25: This looks like a bug :thinking: `return` returns from the function `expand` but if I understand correctly your intention was to collect the vehicles in an array without duplicating them if they don't satisfy the condition. I think the if condition was the right way.

If this didn't break any tests, this means we don't have unit test for `MultiTrips.expand()`. Maybe we should add a test (which fails in this state of the code) with "mixed" vehicles (i.e., with vehicle.trips = {nil, 1 and 2}) that checks that we return 4 vehicles and relations are there, etc..

braktar (approved):

lib/interpreters/multi_trips.rb
```diff
@@ -22,20 +22,30 @@ module Interpreters
...
+          # vrp.relations += [Models::Relation.new(type: 'vehicle_group_duration', linked_vehicle_ids: new_ids, lapse: vehicle.overall_duration)] if vehicle.overall_duration # TODO: Requires a complete rework of overall_duration
 
-          (0..vehicle.trips - 1).collect{ |index|
-            new_vehicle = Marshal.load(Marshal.dump(vehicle))
-            new_vehicle.id += "_trip_#{index}"
-            new_vehicle.trips = 1
-            new_vehicle
+          vrp.services.select{ |service| service.sticky_vehicles.any?{ |sticky_vehicle| sticky_vehicle.id == vehicle.id } }.each{ |service|
+            service.sticky_vehicles -= [vehicle.id]
+            service.sticky_vehicles += new_vehicles
```
senhalil, line 45: looks like a typo
```suggestion
            service.sticky_vehicles += new_ids
```

senhalil (changes requested):

fonsecadeline:
Waiting for https://github.com/Mapotempo/optimizer-api/pull/65 to fix tests

lib/interpreters/multi_trips.rb
```diff
@@ -22,20 +22,30 @@ module Interpreters
   class MultiTrips
     def expand(vrp)
       vrp.vehicles = vrp.vehicles.collect{ |vehicle|
-        if vehicle.trips && vehicle.trips >= 1
-          new_ids = (0..vehicle.trips - 1).collect{ |index| vehicle.id + '_trip_' + index.to_s }
-          containing_relations = vrp.relations.select{ |relation| relation.linked_vehicle_ids.include?(vehicle.id) }.each{ |relation|
-            relation.linked_vehicle_ids -= vehicle.id
+        if vehicle.trips > 1
+          new_ids = Array.new(vehicle.trips) { |index| vehicle.id + '_trip_' + index.to_s }
```
braktar, line 26: Fulfil the original_id field too ? Maybe before the Model create?

senhalil (approved):

PR #49, opened 2020-07-28
Start using balanced_vrp_clustering gem

senhalil:


lib/interpreters/split_clustering.rb
```diff
@@ -303,35 +304,24 @@ def self.kmeans_process(centroids, nb_clusters, data_items, unit_symbols, limits
       while restart < options[:restarts]
         block&.call() # in case job is killed during restarts
         log "Restart #{restart}/#{options[:restarts]}", level: :debug
-        c = BalancedKmeans.new
+        c = Ai4r::Clusterers::BalancedVRPClustering.new
         c.max_iterations = options[:max_iterations]
-        c.centroid_indices = centroids if centroids && centroids.size == nb_clusters
-        c.on_empty = 'random'
```
fonsecadeline, line 309: this was removed but it shouldn't. I had to fix it on my side, I think we should fix it here.

wrappers/ortools.rb
```diff
@@ -881,6 +881,7 @@ def run_ortools(problem, vrp, services, points, matrix_indices, thread_proc = ni
       end
     ensure
       input&.unlink
+      output&.close
```
fonsecadeline, line 884: Discussed already, do not forget to remove this

PR #51, opened 2020-07-29
direct shipment

fonsecadeline:


models/vrp.rb
```diff
@@ -188,6 +189,24 @@ def self.check_consistency(hash)
       raise OptimizerWrapper::DiscordantProblemError, 'Rests are not available with periodic heuristic.' unless hash[:vehicles].all?{ |vehicle| vehicle[:rests].to_a.empty? }
     end
 
+    def self.deduce_data_from_hash(hash)
```
senhalil, line 192: This function only generates relations from shipments but its name is very generic. If we can move it inside `expand_data` then we can name it `create_relations_of_direct_shipments`.

models/vrp.rb
```diff
@@ -188,6 +189,24 @@ def self.check_consistency(hash)
...
+      return hash if hash[:shipments].to_a.empty?
+
```
senhalil, line 194:
```suggestion
```
It is not necessary to check if it is empty because if the `hash[:shipments]` is empty/nil then the for loop does nothing.

models/vrp.rb
```diff
@@ -188,6 +189,24 @@ def self.check_consistency(hash)
...
+      hash[:relations] ||= []
+      hash[:shipments].each{ |shipment|
```
senhalil, line 196:
```suggestion
      hash[:shipments]&.each{ |shipment|
```

models/vrp.rb
```diff
@@ -123,6 +123,7 @@ def self.create(hash, delete = true)
       vrp = super({})
       self.check_consistency(hash)
       self.ensure_retrocompatibility(hash)
+      self.deduce_data_from_hash(hash)
```
senhalil, line 126: Is it possible to move this logic inside `expand_data`?

models/vrp.rb
```diff
@@ -188,6 +189,24 @@ def self.check_consistency(hash)
...
+      hash[:relations] ||= []
+      hash[:shipments].each{ |shipment|
+        if shipment[:direct]
+          hash[:relations] << {
+            type: 'sequence',
+            linked_ids: [shipment[:id] + 'pickup', shipment[:id] + 'delivery']
+          }
+        end
+
+        shipment.delete(:direct)
```
senhalil, line 204: Do we want to delete this? I imagine it can be useful in heuristics :thinking:

models/vrp.rb
fonsecadeline: The idea was not to have _direct_ field within shipments model. In expand_data we would not have this information (_direct_) anymore so it is not possible to move this logic if we keep it like this.

I am not sure it is worth to have _direct_ field in _shipments_ model because we can just generate corresponding relation if we need it. But, as you said it, in heuristics it could be useful to have a "shortcut" to know wether it is a direct shipment or not. 

What do you think ?

models/shipment.rb
```diff
@@ -31,6 +31,8 @@ class Shipment < Base
 
     field :visits_number, default: 1
```
braktar, line 33: stick parameters together

braktar (changes requested):

senhalil (changes requested):
we forgot to add an assert for this option (like `assert_no_direct_shipments`) for the other solvers vroom, jsprit

senhalil (approved):

braktar (approved):

PR #53, opened 2020-08-06
Fix: multiple gets generates error

senhalil:
When there is a solution dump, `job` is nil..

### DIFF
--- a/test/api/v01/vrp_test.rb
+++ b/test/api/v01/vrp_test.rb
@@ -201,6 +201,21 @@ class Api::V01::VrpTest < Api::V01::RequestHelper
     delete_job @job_id, api_key: 'demo'
   end
 
+  def test_get_completed_job_with_solution_dump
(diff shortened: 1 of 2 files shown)


test/api/v01/vrp_test.rb
```diff
@@ -201,6 +201,20 @@ def test_cannot_delete_vrp
...
+  def test_get_completed_job_with_solution_dump
+    old_config_dump_solution = OptimizerWrapper.config[:dump][:solution]
+    OptimizerWrapper.config[:dump][:solution] = true
+    TestHelper.solve_asynchronously do
+      @job_id = submit_vrp api_key: 'demo', vrp: VRP.toy
+      wait_status @job_id, 'completed', api_key: 'demo'
+    end
+
+    get "/0.1/vrp/jobs/#{@job_id}", api_key: 'demo'
+    assert_equal 200, last_response.status, last_response.body
```
fab-girard, line 213: can you check status is `completed` in addition ?

PR #55, opened 2020-08-12
Fix generate rests

fonsecadeline:
Fix disappearance of rests.
Fix overall logic in periodic expand : if no rest/shipment/service/vehicle has day_index then we do not need to expand timewindows (duplicating them and adding 24h x corresponding day) but if any of these elements has day index then they should all have their timewindows expanded.

### DIFF
--- a/test/lib/interpreters/interpreter_test.rb
+++ b/test/lib/interpreters/interpreter_test.rb
@@ -100,7 +100,7 @@ class InterpreterTest < Minitest::Test
     expanded_vrp = periodic.send(:expand, vrp, nil)
     assert_equal 2, expanded_vrp[:vehicles].size
     assert_equal 2 * (size - 1), expanded_vrp[:services].size
-    assert(expanded_vrp[:services].all?{ |service| service.activity.timewindows.size == 2 })
+    assert expanded_vrp[:services].all?{ |service| service.activity.timewindows.size == 4 }, 'We expect 2 timewindows per days. There are two days. Services should all have 4 timewindows.'
     assert_equal expanded_vrp[:services][0].activity.timewindows[0][:start], expanded_vrp[:services][1].activity.timewindows[0][:start]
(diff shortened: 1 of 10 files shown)


braktar (approved):

test/lib/interpreters/interpreter_test.rb
```diff
@@ -19,719 +19,128 @@
...
-    }
+  def periodic_expand(problem)
     vrp = TestHelper.create(problem)
     periodic = Interpreters::PeriodicVisits.new(vrp)
-    expanded_vrp = periodic.send(:expand, vrp, nil)
-    assert_equal 2, expanded_vrp[:vehicles].size
-    assert_equal 2 * (size - 1), expanded_vrp[:services].size
-    assert(expanded_vrp[:services].all?{ |service| service.activity.timewindows.size == 2 })
-    assert_equal expanded_vrp[:services][0].activity.timewindows[0][:start], expanded_vrp[:services][1].activity.timewindows[0][:start]
+    periodic.send(:expand, vrp, nil)
```
braktar, line 25: Interpreters::PeriodicVisits is initialized using the vrp, then we send the vrp a second time, cant we just send it a single time ?

braktar (changes requested):

test/lib/interpreters/interpreter_test.rb
fonsecadeline: You are right, it does not seem so logical to send vrp twice.

However this looks like a big change in periodic_visits.rb. I suggest that we create an issue for this.

braktar (approved):

senhalil:
@fonsecadeline  it looks like this branch is in conflict.. rebase error?

senhalil:
It looks like you rebased over my dev instead of just cherry-picking the specific commit, no?

senhalil (approved):

PR #58, opened 2020-08-21
fix dichotomious approach clustering

fonsecadeline:
I am not sure about this modification but I had an error while running some tests and I am wondering if this is a necessary fix ?


senhalil (approved):
I think you are right @fonsecadeline.. I don't know how and why we had that second inner `each` loop but  this fix looks necessary, otherwise, we are duplicating the data inside clustering.

braktar (approved):

PR #68, opened 2020-09-25
Bump ruby version and standardize rubocop

fab-girard:


.rubocop.yml
```diff
@@ -78,40 +83,25 @@ Style/ClassAndModuleChildren:
...
 Style/WordArray:
   Enabled: false
 
-Style/ParallelAssignment:
-  Enabled: false
-
-Style/PreferredHashMethods:
-  EnforcedStyle: verbose
+Style/RaiseArgs:
+  EnforcedStyle: compact
```
senhalil, line 102: Except 4 out of 69, all our `raise`s are in the default form. If  you want to switch the style, we need to replace all the raises so that the code will not be "red-lined" everywhere.

.rubocop.yml
```diff
@@ -78,40 +83,25 @@ Style/ClassAndModuleChildren:
...
 
-Style/RedundantException:
+Style/ParallelAssignment:
   Enabled: false
```
senhalil, line 105: Since we had this cop in opposite, we have many `raise RuntimeError, ...` in the api, if you want to switch this cop, we need to clean them in this PR.

test/rubocop_test.rb
```diff
@@ -18,10 +18,21 @@
...
 
 class RubocopTest < Minitest::Test
-  def test_no_offenses_in_tests
-    output = `rubocop -f c -S --parallel ./test/*`
-    /[0-9]* files inspected,\ (.*)\ offenses\ detected/ =~ output
-    n_offeses = Regexp.last_match(1)
-    assert_equal('no', n_offeses, "Following new Rubocop offenses added to the tests:\n\n#{output}")
+  def test_no_error_in_files
+    parallel = ENV['RUBOCOP_PARALLEL'] || ENV['TRAVIS'] ? '--parallel' : nil
+    # parallel option could cause not to use rubocop from bundle
```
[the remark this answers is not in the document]
senhalil: In any case we can erase `--parallel` completely because on local it doesn't make any difference, since there is a cache and it only runs the modified files and on Travis there is only one thread..

senhalil (changes requested):
- Some cops need to be discussed
- Code needs cleaning vis-a-vis activated cops (otherwise it creates too many "red-lines")

test/rubocop_test.rb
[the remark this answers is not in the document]
fab-girard: You should not have "red lines" for only convention offenses: check your linter to have a lighter visual indication for convention/refactor offenses. Other levels (Warning, Error, Fatal) should be linted with a more visual indication (for instance "red line" for Error/Fatal) and none offenses from theses types should be accepted. By definition, conventional offenses are less serious.

.rubocop.yml
fab-girard: `raise RuntimeError, ...` instead of `raise RuntimeError,.new` is just convention, but if we follow this convention all our raise are not homogeous

.rubocop.yml
```diff
@@ -3,20 +3,23 @@ require:
...
-  Enabled: false
+  TargetRubyVersion: 2.5
 
 Layout/SpaceBeforeBlockBraces:
   Enabled: false
 
+Layout/LineLength:
+  Max: 120
+
 Metrics/MethodLength:
```
senhalil, line 14: do we really don't want this  cop? It is the "dual" of Layout/LineLength ?

.rubocop.yml
```diff
@@ -3,20 +3,23 @@ require:
...
   Enabled: false
 
 Metrics/AbcSize:
   Enabled: false
 
+Naming/MethodParameterName:
+  MinNameLength: 1
```
senhalil, line 21: if it is for matrix indices, we can use `AllowedNames` parameter for allowing `i, j, k` but otherwise we think this cop is useful.

.rubocop.yml
```diff
@@ -26,10 +29,10 @@ Metrics/PerceivedComplexity:
 Metrics/CyclomaticComplexity:
   Enabled: false
 
-Layout/ParameterAlignment:
+Style/Documentation:
   Enabled: false
 
-Style/Documentation:
+Style/DoubleNegation:
```
senhalil, line 35: We find this dangerous inside an algorithm because `== true` is a lot more expressive.

.rubocop.yml
```diff
@@ -78,43 +83,25 @@ Style/ClassAndModuleChildren:
...
-
 Layout/EndAlignment:
-  Enabled: false
-
-Layout/ElseAlignment:
-  Enabled: false
-
-Performance/ChainArrayAllocation:
-  Enabled: true
+  EnforcedStyleAlignWith: start_of_line
```
senhalil, line 87: We didn't know this cop was dis activated! We would like to just activate this cop with the default setting if it is alright because the assignments are more visible (especially if they are hidden inside an algorithmic logic)

If the variable name and the condition is short enough, it is very clear
```ruby
@clusters[index].data_items.each{ |d_i|
  distance_weight = if d_i[4][:moved_up]
                      0.1
                    elsif d_i[4][:moved_down]
                      10
                    else
                      1
                    end
}
```

If the variable name is too long and/or the condition is too long we find the following assignment more visible
```ruby
clusters.each_with_index{ |cluster, index|
  data_sticky = cluster.data_items.collect{ |data| data[4] }.compact
  data_skills = cluster.data_items.collect{ |data| data[5] }.compact
  sub_problem[index][:vrp][:vehicles] =
    if !data_sticky.empty?
      [vehicle_list.find{ |vehicle| vehicle[:id] == data_sticky.flatten.first }].compact
    elsif data_sticky.empty? && !data_skills.empty? && vehicle_list.any?{ |vehicle| vehicle[:skills] && !vehicle[:skills].first.empty? }
      [vehicle_list.find{ |vehicle| vehicle[:skills] && !vehicle[:skills].first.empty? && ([data_skills.flatten.first] & vehicle[:skills].first).size > 0 }]
    else
      [vehicle_list.last] # TODO : function that return the best vehicle for this cluster
    end
  vehicle_list -= sub_problem[index][:vrp][:vehicles]
}
```

instead of the following because the if condition and the assignment is lost in the clutter and the logic is not visible immediately:

```ruby
clusters.each_with_index{ |cluster, index|
  data_sticky = cluster.data_items.collect{ |data| data[4] }.compact
  data_skills = cluster.data_items.collect{ |data| data[5] }.compact
  sub_problem[index][:vrp][:vehicles] = if !data_sticky.empty?
    [vehicle_list.find{ |vehicle| vehicle[:id] == data_sticky.flatten.first }].compact
  elsif data_sticky.empty? && !data_skills.empty? && vehicle_list.any?{ |vehicle| vehicle[:skills] && !vehicle[:skills].first.empty? }
    [vehicle_list.find{ |vehicle| vehicle[:skills] && !vehicle[:skills].first.empty? && ([data_skills.flatten.first] & vehicle[:skills].first).size > 0 }]
  else
    [vehicle_list.last] # TODO : function that return the best vehicle for this cluster
  end
  vehicle_list -= sub_problem[index][:vrp][:vehicles]
}
```

.rubocop.yml
senhalil: effectively this is the case, I misunderstood this cop apparently! I like the new setting better because of homogeneity too!

.rubocop.yml
```diff
@@ -78,43 +83,25 @@ Style/ClassAndModuleChildren:
...
-
-Style/PreferredHashMethods:
-  EnforcedStyle: verbose
-
-Style/RedundantException:
-  Enabled: false
+Style/RaiseArgs:
+  EnforcedStyle: compact
 
-Style/FrozenStringLiteralComment:
```
senhalil, line 117: We don't know how we can handle this cop, this is why we disactivated it. Otherwise, all it does is to give the same warning for all files. If we can find a way to freeze the strings then we are all for doing it. But how?

.rubocop.yml
```diff
@@ -44,12 +47,11 @@ Style/TrailingCommaInArguments:
...
   Enabled: false
 
-Style/TernaryParentheses:
-  EnforcedStyle: require_parentheses_when_complex
+Style/MultilineTernaryOperator:
+  Enabled: false
 
 Layout/SpaceInsideHashLiteralBraces:
-  Enabled: true
-  EnforcedStyle: compact
```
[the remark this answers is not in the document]
fab-girard: you will add this rule in a local rubocop file for optim

.rubocop.yml
fab-girard: I prefere keep this rule

senhalil: me too (i don't understand my comment)

senhalil: Just understood my comment... it is not about `Layout/LineLength`! We want to keep `Layout/LineLength` as well. 

However, my comment is about `Metrics/MethodLength`: "don't we want this MethodLength cop?" 

Because at the moment, we define a limit on "horizontal" with LineLength but we dis-activate the limit on "vertical" with MethodLength: false. So we risk having very long methods. 

These two limits work together as duals to limit the complexity and increase the readability, activating only one without the other just distorts the "shape" of the code (i.e., either it become really long vertically or horizontally) and we might want to consider activating MethodLength as well.

fab-girard: we are going to keep `Metrics/MethodLength:` with `Max: 30`

.rubocop.yml
fab-girard: The second block is the most elegant, whatever value of `EnforcedStyleAlignWith` it's authorized, this cop is not very relevant

PR #71, opened 2020-10-12
v1: use function to reduce memory usage

fonsecadeline:


models/service.rb
```diff
@@ -53,5 +54,14 @@ class Service < Base
     has_many :activities, class_name: 'Models::Activity'
     has_many :sticky_vehicles, class_name: 'Models::Vehicle'
     has_many :quantities, class_name: 'Models::Quantity'
+
+    def duplicate_safe(options = {})
```
senhalil, line 58: I thought we decided to keep all this duplicate logic inside periodic_visits until we have the generic clone function ready so that it wouldn't be used somewhere else, do I remember wrong?

I find this function dangerous because it has periodic_visits logic in it (for example a generic duplicate function wouldn't skip `:sticky_vehicle_ids, :quantity_ids`) but it is inside service.rb (the same for vehicles, shipments)

fonsecadeline: You are right, if I move it in periodic_visits.rb we make it harder to use this function and we will be forced to implement a better way. I will move it.

Still, I do think this was a generic function : you can have sticky_vehicle_ids and quantity_ids even if you do not have a schedule. But you are right, this would probably come from a scheduling optimization since it implies using scheduling front (which works well even if you do not provide any schedule in configuration).

senhalil: I wasn't clear enough about the thing that is not generic.

The function is not generic, not because of the existence of `sticky_vehicle_ids` or `quantity_ids` (of course we can have them or not) but it is not generic because of the fact that we are skipping these in this duplicate function :
```ruby
next if [:sticky_vehicle_ids, :quantity_ids].include?(k)
```

This "skipping" logic makes it specific in my opinion and it wouldn't be right to make it a generic `Service` class function. A generic duplicate function would duplicate the objects faithfully.

fonsecadeline: I see. Then I might need @braktar 's help and yours : 

self.attributes iterates over every field defined within grape model. How can we rather iterate over active_hash model attributes ? 

Or maybe we can remove attributes with key **k** such that :
`key.include?("_id") || key.include?("_ids") || key.include?("y_ids")`
after model is created.

What would be the best practice here ?

senhalil: I don't know immediately but in any case, this PR is for the non-generic version, so lets move this function inside periodic_visits and we can finalize this PR. 

We shall discuss how to implement the generic version on the other ticket when we are  working on it. I think it'll be clearer how to handle all of this, one we are inside the base.rb.

I just wanted to make myself clear about why I didn't see this function a generic function that's it.

braktar: fields with '_id' or '_ids' are sometimes part of the model.

I think that iterate over the fields define within the model would be more generic.

https://github.com/zilkey/active_hash/blob/master/lib/active_hash/base.rb
We could probably explore the possibility to use `field_names` methods

fonsecadeline: field_names seems undefined so I will just move functions in periodic.rb and we will discuss later as @senhalil suggested

Thank you for your help :)

fonsecadeline: @senhalil we still need one function per model (vehicle/service/shipment) because we call Model::new. Do we still want to put it in periodic_visits.rb ?

senhalil: It's normal that we need different functions for each of these, even if we move it to `periodic_visits.rb`. The reason why we want to move it to there is that they shouldn't be used. But basically what I had in mind was a function (lets call it `periodic_visits::duplicate_model`) which duplicates the necessary parts the way periodic visits needs. Lets discuss this on the meeting via screen sharing on Monday.

lib/heuristics/concerns/scheduling_data_initialisation.rb
```diff
@@ -192,8 +190,18 @@ def adapt_services_data(vrp)
...
-    @services_data.collect{ |_id, data| data[:points_ids] }.flatten.uniq.sort.each{ |point_id|
-      @indices[point_id] = vrp.points.find{ |pt| pt.id == point_id }&.matrix_index
+    vrp.vehicles.each{ |vehicle|
+      @indices[vehicle.start_point.id] = vehicle.start_point.matrix_index if vehicle.start_point
+      @indices[vehicle.end_point.id] = vehicle.end_point.matrix_index if vehicle.end_point
+    }
+
+    vrp.services.each{ |service|
+      [service.activity ? [service.activity] : service.activities].flatten.each{ |activity|
+        @indices[activity.point.id] = activity.point.matrix_index
```
braktar, line 200: Isn't this formulation possible ?
(service.activity ? [service.activity] : service.activities).each

lib/interpreters/periodic_visits.rb
```diff
@@ -118,157 +114,113 @@ def generate_relations(vrp)
       }.compact.flatten
     end
 
+    def generate_relations_between_visits(vrp, mission)
+      # TODO : need to uniformize generated relations whether mission has minimum AND maximum lapse or only one of them
+      return unless mission.visits_number > 1
+
+      if mission.minimum_lapse && mission.maximum_lapse
```
braktar, line 121: could you factorise the relation definition ? and give the specific mission_ids in a single place. It is a bit difficult to see the distinction between the cases.

models/service.rb
```diff
@@ -27,15 +27,16 @@ class Service < Base
     # validates_numericality_of :exclusion_cost, allow_nil: true
     field :type, default: :service
 
-    field :first_possible_day, default: nil
-    field :last_possible_day, default: nil
+    field :first_possible_days, default: []
+    field :last_possible_days, default: []
+    field :first_possible_day, default: nil # this should be a VISIT attribute
```
braktar, line 32: Did you mean, that each duplicated visit, should have only one first_possible_day, while the original on has an array ?

braktar (changes requested):

lib/heuristics/concerns/scheduling_data_initialisation.rb
fonsecadeline: you are right, that would be better

models/service.rb
fonsecadeline: In fact this should rather be on _visit_ model only but we don't have it yet. Avery visit would have only one first_possible day.

In the meantime, one service has as many first_possible_days as visits before expand, and one first_possibleçday after expand. We could indeed keep only _first_possible_days_ even after expand.

lib/interpreters/periodic_visits.rb
fonsecadeline: For now the goal was to group code that was identical for shipment and services.

I would indeed like to factorise but it is not that easy because of https://gitlab.com/mapotempo/optimizer-api/-/issues/339 . There can be an impact we need to test.

wrappers/ortools.rb
```diff
@@ -472,7 +472,7 @@ def build_costs(costs)
 
     def check_services_compatible_days(vrp, vehicle, service)
       if vrp.schedule_range_indices && (service.minimum_lapse || service.maximum_lapse)
-        (vehicle.global_day_index >= service[:first_possible_day] && vehicle.global_day_index <= service[:last_possible_day]) ? true : false
+        (vehicle.global_day_index >= service[:first_possible_days].first && vehicle.global_day_index <= service[:last_possible_days].first) ? true : false
```
braktar, line 475: The condition already returns true or false ?

braktar: The content of the function could be replaced with:
```
!vrp.schedule_range_indices || (!service.minimum_lapse && !service.maximum_lapse) ||
  (vehicle.global_day_index >= service[:first_possible_days].first && vehicle.global_day_index <= service[:last_possible_days].first)
```

braktar (changes requested):

braktar (approved):

PR #72, opened 2020-10-14
Various small fixes

braktar:


lib/interpreters/split_clustering.rb
```diff
@@ -277,7 +277,7 @@ def self.build_partial_service_vrp(service_vrp, partial_service_ids, available_v
       sub_vrp.rests = sub_vrp.rests.select{ |r| sub_vrp.vehicles.flat_map{ |v| v.rests.map(&:id) }.include? r.id }
       sub_vrp.relations = sub_vrp.relations.select{ |r| r.linked_ids.all? { |id| sub_vrp.services.any? { |s| s.id == id } || sub_vrp.shipments.any? { |s| id == s.id + 'delivery' || id == s.id + 'pickup' } } }
       sub_vrp.points = sub_vrp.points.select{ |p| points_ids.include? p.id }.compact
-      sub_vrp.points += sub_vrp.vehicles.flat_map{ |vehicle| [vehicle.start_point, vehicle.end_point] }.compact
+      sub_vrp.points += sub_vrp.vehicles.flat_map{ |vehicle| [vehicle.start_point, vehicle.end_point] }.uniq.compact
```
senhalil, line 280: I suggest the following because uniq is slightly more expensive then compact; otherwise nice catch!  

```suggestion
      sub_vrp.points += sub_vrp.vehicles.flat_map{ |vehicle| [vehicle.start_point, vehicle.end_point] }.compact.uniq
```

lib/interpreters/split_clustering.rb
```diff
@@ -608,7 +608,7 @@ def build_service_like(shipment, activity)
                             quantities: shipment.quantities)
       end
 
-      def build_services_from_shipments(shipments)
+      def build_services_from_shipments(depot_ids, shipments)
```
senhalil, line 611: It is suprising that we didn't see this bug before, this means there is no test that passes through this function. Can you add one please?

models/vehicle.rb
```diff
@@ -118,8 +118,8 @@ def self.create(hash)
     end
 
     def need_matrix_time?
-      cost_time_multiplier != 0 || cost_late_multiplier && cost_late_multiplier != 0 || cost_setup_time_multiplier != 0 ||
-        !rests.empty? || maximum_ride_time || duration || overall_duration
+      cost_time_multiplier.positive? || (timewindow.end && (cost_late_multiplier.nil? || cost_late_multiplier&.zero?)) || cost_late_multiplier&.positive? ||
```
senhalil, line 121: why `(timewindow.end && (cost_late_multiplier.nil? || cost_late_multiplier&.zero?))` instead of just `timewindow.end` like follows because if `cost_late_multiplier&.positive?` doesn't hold the extra bit `(cost_late_multiplier.nil? || cost_late_multiplier&.zero?)` always holds

```suggestion
      cost_time_multiplier.positive? || cost_late_multiplier&.positive? || timewindow.end
```

senhalil (changes requested):
smaller fixes

models/vehicle.rb
braktar: I still got the exception case for vroom in mind. But, Indeed, it shouldn't impact the need of time dimension.

models/vehicle.rb
```diff
@@ -118,8 +118,8 @@ def self.create(hash)
     end
 
     def need_matrix_time?
-      cost_time_multiplier != 0 || cost_late_multiplier && cost_late_multiplier != 0 || cost_setup_time_multiplier != 0 ||
-        !rests.empty? || maximum_ride_time || duration || overall_duration
+      cost_time_multiplier.positive? || timewindow.end || cost_late_multiplier&.positive? ||
+        cost_setup_time_multiplier.positive? || !rests.empty? || maximum_ride_time || duration || overall_duration
     end
```
fonsecadeline, line 123: Don't we always need time_matrix ?

lib/interpreters/split_clustering.rb
```diff
@@ -377,7 +377,7 @@ def self.split_balanced_kmeans(service_vrp, nb_clusters, options = {}, &block)
       options = default_options.merge(options)
       vrp = service_vrp[:vrp]
       # Split using balanced kmeans
-      if vrp.services.all?{ |service| service[:activity] && service[:activity][:point][:location] } && nb_clusters > 1
+      if vrp.services.all?{ |service| service&.activity&.point&.location } && nb_clusters > 1
```
fonsecadeline, line 380: We should check the same for shipments too.

fonsecadeline (approved):

models/vehicle.rb
braktar: VROOM only tolerate a single dimension.

If we don't need one dimension with HERE matrix, we can save a large amount of time

PR #73, opened 2020-10-19
Fix ids without scheduling

fonsecadeline:
Homogenization meanwhile https://gitlab.com/mapotempo/optimizer-api/-/issues/603

### DIFF
--- a/lib/interpreters/periodic_visits.rb
+++ b/lib/interpreters/periodic_visits.rb
@@ -233,19 +233,16 @@ module Interpreters
 
     def generate_vehicles(vrp)
       rests_durations = Array.new(vrp.vehicles.size, 0)
-      vrp.vehicles.each{ |vehicle|
-        vehicle.id = vehicle.original_id if vehicle.original_id # if we used work_day clustering
-        vehicle.original_id = vehicle.id if vehicle.original_id.nil?
-      }
       new_vehicles = vrp.vehicles.collect{ |vehicle|
         @equivalent_vehicles[vehicle.id] = []
+        @equivalent_vehicles[vehicle.original_id] = []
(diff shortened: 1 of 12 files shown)


optimizer_wrapper.rb
```diff
@@ -690,11 +690,25 @@ def self.route_details(vrp, route, vehicle)
     details
   end
 
+  def self.provide_original_ids(data)
```
braktar, line 693: Shall we initialize the original id when the vrp is constructed for the first time ? And use this value when the result is under construction ?

braktar (changes requested):

lib/interpreters/periodic_visits.rb
```diff
@@ -172,6 +172,7 @@ def generate_services(vrp)
             new_service = nil
             if !service.unavailable_visit_indices || service.unavailable_visit_indices.none?{ |unavailable_index| unavailable_index == visit_index }
               new_service = Marshal.load(Marshal.dump(service))
+              new_service.original_id = new_service.id
```
braktar, line 175: It has to be defined in models/vrp.rb within the create method. This way, we are sure it is defined for every service or shipment.

Here, it is only defined in a periodic context

wrappers/ortools.rb
```diff
@@ -587,6 +587,7 @@ def parse_output(vrp, _services, points, _matrix_indices, _cost, _iterations, ou
                   current_matrix_index = point.matrix_index
                   route_data = build_route_data(vehicle_matrix, previous_matrix_index, current_matrix_index)
                   current_activity = {
+                    original_service_id: service.original_id,
```
braktar, line 590: Currently It cannot be used if we are not in a periodic context ? It will return a nil value

braktar (changes requested):

braktar (approved):

braktar (approved):

PR #74, opened 2020-10-22
Use equi

fonsecadeline:
Mostly refactoring + implementing an option to always try to balance schedule days by default.
Slight modification of the choice of the first point to insert if current route is empty (needed when refactoring).


fonsecadeline:
Based on : https://github.com/Mapotempo/optimizer-api/pull/71 because it made it easier to test impact.

PR #76, opened 2020-10-23
Ensure consistent lapses when same point day

fonsecadeline:


lib/heuristics/concerns/scheduling_data_initialisation.rb
```diff
@@ -311,4 +309,17 @@ def compute_capacities(quantities, vehicle, available_units = [])
 
     capacities
   end
+
+  def exist_common_multiplexer(lapses)
```
braktar, line 313: Could you add some comments on the purpose of the method ?

if a boolean is returned, it would be more natural to add a question mark to the method name : exist_common_multiplexer?

Does it return true if  multiplexers.size is equal to zero ?

braktar (changes requested):

braktar (approved):

lib/heuristics/concerns/scheduling_data_initialisation.rb
```diff
@@ -311,4 +309,22 @@ def compute_capacities(quantities, vehicle, available_units = [])
 
     capacities
   end
+
+  def exist_common_multiplexer?(lapses)
```
senhalil, line 313: [multiplexer](https://en.wikipedia.org/wiki/Multiplexer) has a very specific meaning on electronics :thinking:  do we mean "multiple", "divisor" something like that here?

If we also name the function `has_common_X?` or `common_X_exists?`, it would be less Yodish, what do you think?

senhalil: I didn't mark every appearance of the word `multiplexer` but can you replace them with the correct word as well?

lib/heuristics/concerns/scheduling_data_initialisation.rb
```diff
@@ -311,4 +309,22 @@ def compute_capacities(quantities, vehicle, available_units = [])
...
+    lapses = lapses.compact.uniq
+    min_lapse = lapses.min
+
+    multiplexers = lapses.collect{ |lapse|
+      (2..min_lapse).select{ |multiplexer|
+        (lapse % multiplexer).zero?
+      }
+    }
+
+    multiplexers.size < 2 || multiplexers.first.any?{ |value| multiplexers.all?{ |set| set.include?(value) } }
```
senhalil, line 328: It feels like `multiplexers.size < 2` case only happen if the lapses array has 1 uniq element. In that case, we can `return true` earlier (just after `lapses = lapses.compact.uniq`), right ?

lib/heuristics/concerns/scheduling_data_initialisation.rb
```diff
@@ -311,4 +309,22 @@ def compute_capacities(quantities, vehicle, available_units = [])
...
     capacities
   end
+
+  def exist_common_multiplexer?(lapses)
+    # If lapses have common multiplexer then any of these
+    # services will be assigned at the same time as service
+    # with smallest lapse (and normaly biggest number of visits).
+    # If this is not the case, we can not guarantee
+    # same_point_day constraint.
+    lapses = lapses.compact.uniq
```
senhalil, line 319: This looks like a side-effect, are we sure we want to override the original list? If we call the function with a proper variable in the future, it will modify the variable.

lib/heuristics/concerns/scheduling_data_initialisation.rb
senhalil: After reading the function in detail, I think the word you are looking for is "divisor" but not sure.

lib/heuristics/concerns/scheduling_data_initialisation.rb
senhalil: Actually IF the goal of this function is to check the existence a common divisor for all values in the `lapses` array then we don't need to do any of these, we already have such a function in the `Filters` class 

```ruby
def has_common_divisor?(lapses)
  uniq_lapses = lapses.compact.uniq
  return true if uniq_lapses.size < 2 

  gcd = Filters.gcd_of_array(uniq_lapses)
  return gcd > 1 # if gcd is 1 that means there is no common divisor
end
```

Since I don't know the exact usage, I can't be sure but it looks like to me this should do what you want. What do you think @fonsecadeline ?

senhalil (changes requested):
we have a function to find greatest common divisor of an array of value, I think we can use that here.

lib/heuristics/concerns/scheduling_data_initialisation.rb
fonsecadeline: I did not know about this function, perfect for me !

lib/heuristics/concerns/scheduling_data_initialisation.rb
```diff
@@ -152,32 +152,30 @@ def adapt_services_data(vrp)
...
-        }
-      }.sort_by(&:visits_number)
+      same_located_set = @services_data.select{ |_id, data| data[:points_ids].include?(point.id) }.sort_by{ |_id, data| data[:raw].visits_number }
 
       next if same_located_set.empty?
 
-      raise OptimizerWrapper.UnsupportedProblemError, 'Same_point_day is not supported if a set has one service with several activities' if same_located_set.any?{ |s| !s.activities.empty? }
+      raise OptimizerWrapper.UnsupportedProblemError, 'Same_point_day is not supported if a set has one service with several activities' if same_located_set.any?{ |id, data| data[:points_ids].size > 1 }
+
+      raise OptimizerWrapper::UnsupportedProblemError, 'Same_point_day is not supported if frequences of a set have no common multiplexer' unless common_divisor?(same_located_set.collect{ |_id, data| data[:heuristic_period] })
```
senhalil, line 161:
```suggestion
      raise OptimizerWrapper::UnsupportedProblemError, 'Same_point_day is not supported if frequencies of a set have no common divisor' unless has_common_divisor?(same_located_set.collect{ |_id, data| data[:heuristic_period] })
```

braktar (approved):

lib/heuristics/concerns/scheduling_data_initialisation.rb
```diff
@@ -305,4 +303,18 @@ def compute_capacities(quantities, vehicle, available_units = [])
 
     capacities
   end
+
+  def common_divisor?(lapses)
+    # If lapses have common multiplexer then any of these
```
senhalil, line 308: I think `common_divisor?` naming is more appropriate for a function that checks if the caller has a common divisor -- like `5.common_divisor?(3)` 
```suggestion
  def has_common_divisor?(lapses)
    # If lapses have a common divisor then any of these
```

lib/heuristics/concerns/scheduling_data_initialisation.rb
senhalil, line 161: Wouldn't it be better to move these checks inside the `self.check_consistency(hash)` function. Because otherwise we accept the VRP and return `200` but then raise an error.

senhalil (changes requested):
Wouldn't it be better to move these checks inside the `check_consistency(hash)` function on the api side? 

Because otherwise we accept the VRP (and return 201) but then we raise an error and when the get job is called we return 400.

senhalil (approved):
Thanks @fonsecadeline it is :ok_hand:

PR #82, opened 2020-11-16
Detect negative value returned by router

fonsecadeline:
https://sentry.mapotempo.com/mapotempo/api-optimizer-prod/issues/14086/?query=is:unresolved

### DIFF
--- a/models/concerns/distance_matrix.rb
+++ b/models/concerns/distance_matrix.rb
@@ -72,6 +72,8 @@ module DistanceMatrix
           distance: (router_matrices[dimensions.index(:distance)] if dimensions.index(:distance)),
           value: (router_matrices[dimensions.index(:value)] if dimensions.index(:value))
         )
+        raise RouterError, 'Negative value provided by router' if [m.time, m.distance, m.value].any?{ |matrix| matrix&.any?{ |row| row.any?(&:negative?) } }
+
         self.matrices += [m]
         [[mode, dimensions, options], m]
       }]


fonsecadeline:
@braktar @senhalil I am not sure if this should return an error or a warning. It seems that router returns negative values more than what we expect, but this does not always produce an error on optimizer-api side.

fonsecadeline:
Could be usefull for https://gitlab.com/mapotempo/router-api/-/issues/81#note_456839979

senhalil:
We cannot leave negative values in the matrix because or-tools has network-algorithm-based logic... so a negative distance (even if it doesn't lead to a "loud" error) it can generate bugs that would be a pain to debug.

I suggest, if this only happens for very very close points, we can either take the absolute value or consider them as 0.

@fab-girard  this can be an issue even for mt-web while calculating the route.. I think a correction on the router-api (either setting such values to 0 or taking the absolute value) would be the best option (if osrm is not willing to fix it).

fab-girard:
The correction must be in router project.
Is it really useful for you to raise a specific exception when this error form router occurs?

senhalil:
I just checked the ticket on the router-api side.  This is an OSRM problem and their suggestion is to mitigate it on the user side so it is better to fix this on the router-api side and then we can merge this `raise RouterError` so that it will not cause a problem on the optimiser-api side.

fonsecadeline:
I think it would save us debugging time. When receiving sentry https://sentry.mapotempo.com/mapotempo/api-optimizer-prod/issues/14086/?query=is:unresolved it was not clear to me that the problem came from router computation. I suspect more sentry could be produced by this error.

senhalil:
> The correction must be in router project.
> Is it really useful for you to raise a specific exception when this error form router occurs?

There is a risk that negative values would break ortools optimisation logic but also when we develop inside optimizer-api we always (or inherently) assume that distances are non-negative so it is better if we do this raise I think (but only after the router-api is fixed). What do you say @fab-girard?

fab-girard:
ok @fonsecadeline can you benchmarck a test on a 5000x5000 x3 (time/dist/value) matrices please?

fonsecadeline:
Do you want me to test router and see if it returns a negative value for 5000x5000 x3 (time/dist/value) matrices or add a test to see if raise works ?

senhalil:
@fab-girard apparently these negative values can rarely happen under very specific conditions (https://github.com/Project-OSRM/osrm-backend/issues/5855#issuecomment-708058935) very close coordinates but also some other technical conditions.

I think it is hard to make it happen randomly. But a real user instance (with lots of very close delivery/pickup points in the same building structure/street) might generate it. Like you said, this is going to be fixed on the router-api side and this check would just serve as a protection of optimizer-api against incoherent distance data.

fab-girard:
I wasn't clear: @fonsecadeline I've mentioned a benchmark. I want just to ensure checking this negative value in 3x 5000x5000 matrices is insignifiant in time consuming as wrote with a `flatten` (you don't need to launch an optim or have negative values in matrices, a simple ruby script will do the job! ;) )

senhalil:
@fonsecadeline while you are at it, can you also test the following as well (no compact, no flatten)
```ruby
[m.time, m.distance, m.value].any?{ |matrix| matrix&.any?{ |row| row.any?(&:negative?) } }
```
I think `compact` also creates a new matrix, like `flatten`.

You can use `benchmark/ips` for this 
```ruby
require 'benchmark/ips'

Benchmark.ips do |x|
  x.config(time: 30, warmup: 5)
  x.report('1') {
  }
  x.report('2') {
  }
  x.compare!
end
```

models/concerns/distance_matrix.rb
```diff
@@ -72,6 +72,8 @@ def compute_need_matrix(&block)
           distance: (router_matrices[dimensions.index(:distance)] if dimensions.index(:distance)),
           value: (router_matrices[dimensions.index(:value)] if dimensions.index(:value))
         )
+        raise RouterError, 'Negative value provided by router' if [m.time, m.distance, m.value].compact.any?{ |matrix| matrix.flatten.any?(&:negative?) }
```
senhalil, line 75:
```suggestion
        raise RouterError, 'Negative value provided by router' if [m.time, m.distance, m.value].any?{ |matrix| matrix&.any?{ |row| row.any?(&:negative?) } }
```
I did the benchmark during the meeting, it turns out without flatten/compact, the routine is ~2.4x faster if there are no negative values and if there are negative values, it is thousands times (~5K) faster depending on the location of the negative value..

senhalil (changes requested):
```bash
#No negative values in the matrices.
Warming up --------------------------------------
     compact-flatten     1.000  i/100ms
             nothing     1.000  i/100ms
Calculating -------------------------------------
     compact-flatten      3.309  (± 0.0%) i/s -     34.000  in  10.280770s
             nothing      7.860  (± 0.0%) i/s -     79.000  in  10.054909s

Comparison:
             nothing:        7.9 i/s
     compact-flatten:        3.3 i/s - 2.38x  (± 0.00) slower

#Some negative values in one of the matrices.
Warming up --------------------------------------
     compact-flatten     1.000  i/100ms
             nothing     9.319k i/100ms
Calculating -------------------------------------
     compact-flatten     17.201  (± 5.8%) i/s -    172.000  in  10.020032s
             nothing     92.711k (± 2.9%) i/s -    931.900k in  10.061909s

Comparison:
             nothing:    92711.2 i/s
     compact-flatten:       17.2 i/s - 5389.73x  (± 0.00) slower
```
```ruby
require 'benchmark/ips'

# define sizexsize matrices
size = 1000
time = Array.new(size){ Array.new(size){ rand(100) - 1 }} # matrix with ~1% negatif values (replace 1 with 0 for no negative values)
distance = Array.new(size){ Array.new(size){ 1 }}
value = Array.new(size){ Array.new(size){ 1 }}


Benchmark.ips do |x|
  x.config(time: 10, warmup: 1)
  x.report('compact-flatten') {
      [time, distance, value].compact.any?{ |matrix| matrix.flatten.any?(&:negative?) }
  }
  x.report('nothing') {
      [time, distance, value].any?{ |matrix| matrix&.any?{ |row| row.any?(&:negative?) } }
  }
  x.compare!
end
```
The check without compact/flatten takes 0.1 seconds for 1000x1000 matrix and it is ~3 seconds for 5000x5000 matrix. I think this is a small price to pay for the security against negative values, what do you think @fab-girard?

senhalil (approved):

braktar (approved):

PR #83, opened 2020-11-17
More data in result

fonsecadeline:
Makes result documentation and result output consistent, for every solver.

New fields : idle_time, begin_time, end_time and total_idle_time.
(original_vehicle_id and original_id come from fix https://github.com/Mapotempo/optimizer-api/pull/73, approved but not merged yet).

https://gitlab.com/mapotempo/optimizer-api/-/issues/616
old headers | day_week_num | day_week | vehicle_id | id | point_id | lat | lon | type |   |   |   | setup_duration | duration | additional_value | skills | tags | total_travel_time | total_travel_distance |   | quantity_kg | quantity_l | quantity_qte | timewindow_start_0 | timewindow_end_0 | timewindow_start_1 | timewindow_end_1 | unassigned_reason
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
new headers | day_week_num | day_week | vehicle_id | id | point_id | lat | lon | type | waiting_time | begin_time | end_time | setup_duration | duration | additional_value | skills | tags | total_travel_time | total_travel_distance | total_waiting_time | quantity_kg | quantity_l | quantity_qte | timewindow_start_0 | timewindow_end_0 | timewindow_start_1 | timewindow_end_1 | unassigned_reason

### DIFF
--- a/api/v01/entities/vrp_result.rb
+++ b/api/v01/entities/vrp_result.rb
@@ -102,25 +102,25 @@ module Api
     end
 
     class VrpResultJobGraphItem < Grape::Entity
-      expose :iteration, documentation: { type: Integer, desc: 'Iteration number.' }
-      expose :time, documentation: { type: Integer, desc: 'Time in ms since resolution begin.' }
-      expose :cost, documentation: { type: Float, desc: 'Current best cost at this iteration.' }
+      expose :iteration, documentation: { type: Integer, desc: 'Iteration number' }
+      expose :time, documentation: { type: Integer, desc: 'Time in ms since resolution begin' }
+      expose :cost, documentation: { type: Float, desc: 'Current best cost at this iteration' }
     end
 
     class VrpResultJob < Grape::Entity
       expose :id, documentation: { type: String, desc: 'Job uniq ID' }
-      expose :status, documentation: { type: String, desc: 'One of queued, working, completed, killed or failed.' }
-      expose :avancement, documentation: { type: String, desc: 'Free form advancement message.' }
-      expose :graph, using: VrpResultJobGraphItem, documentation: { is_array: true, desc: 'Items to plot cost evolution.' }
+      expose :status, documentation: { type: String, desc: 'One of queued, working, completed, killed or failed' }
+      expose :avancement, documentation: { type: String, desc: 'Free form advancement message' }
+      expose :graph, using: VrpResultJobGraphItem, documentation: { is_array: true, desc: 'Items to plot cost evolution' }
     end
 
     class VrpResult < Grape::Entity
-      expose :solutions, using: VrpResultSolution, documentation: { is_array: true, desc: 'The current best solution.' }
-      expose :job, using: VrpResultJob, documentation: { desc: 'The Job status.' }
+      expose :solutions, using: VrpResultSolution, documentation: { is_array: true, desc: 'The current best solution' }
+      expose :job, using: VrpResultJob, documentation: { desc: 'The Job status' }
     end
 
     class VrpJobsList < Grape::Entity
-      expose :jobs, using: VrpResultJob, documentation: { is_array: true, desc: 'The Jobs.' }
+      expose :jobs, using: VrpResultJob, documentation: { is_array: true, desc: 'The Jobs' }
     end
   end
 end
--- a/api/v01/entities/vrp_result.rb
+++ b/api/v01/entities/vrp_result.rb
@@ -61,6 +61,7 @@ module Api
       expose :travel_value, documentation: { type: Integer, desc: 'Travel value from previous point' }
       expose :waiting_time, documentation: { type: Integer, desc: '' }
       expose :begin_time, documentation: { type: Integer, desc: 'Time visit starts' }
+      expose :end_time, documentation: { type: Integer, desc: 'Time visit ends' }
       expose :departure_time, documentation: { type: Integer, desc: '' }
       expose :service_id, documentation: { type: String, desc: 'Internal reference of the service' }
       expose :pickup_shipment_id, documentation: { type: String, desc: 'Internal reference of the shipment' }
@@ -77,6 +78,7 @@ module Api
       expose :total_travel_time, documentation: { type: Integer, desc: 'Sum of every travel time within the route' }
       expose :total_distance, documentation: { type: Integer, desc: 'Sum of every distance within the route' }
       expose :total_time, documentation: { type: Integer, desc: 'Sum of every travel time and activity duration of the route' }
+      expose :total_waiting_time, documentation: { type: Integer, desc: 'Sum of every idle time within the route' }
       expose :start_time, documentation: { type: Integer, desc: 'Give the actual start time of the current route if provided by the solve' }
       expose :end_time, documentation: { type: Integer, desc: 'Give the actual end time of the current route if provided by the solver' }
       expose :geometry, documentation: { type: String, desc: 'Contains the geometry of the route, if asked in first place' }
--- a/api/v01/entities/vrp_result.rb
+++ b/api/v01/entities/vrp_result.rb
@@ -56,20 +56,25 @@ module Api
 
     class VrpResultSolutionRouteActivities < Grape::Entity
       expose :point_id, documentation: { type: String, desc: 'Linked spatial point' }
-      expose :travel_distance, documentation: { type: Integer, desc: 'travel distance from the previous point' }
-      expose :travel_duration, documentation: { type: Integer, desc: 'travel time from the previous point' }
-      expose :waiting_duration, documentation: { type: Integer, desc: '' }
-      expose :arrival_time, documentation: { type: Integer, desc: '' }
+      expose :travel_distance, documentation: { type: Integer, desc: 'Travel distance from previous point' }
+      expose :travel_time, documentation: { type: Integer, desc: 'Travel time from previous point' }
+      expose :travel_value, documentation: { type: Integer, desc: 'Travel value from previous point' }
+      expose :waiting_time, documentation: { type: Integer, desc: '' }
+      expose :begin_time, documentation: { type: Integer, desc: 'Time visit starts' }
       expose :departure_time, documentation: { type: Integer, desc: '' }
-      expose :service_id, documentation: { type: String, desc: '' }
-      expose :pickup_shipment_id, documentation: { type: String, desc: '' }
-      expose :delivery_shipment_id, documentation: { type: String, desc: '' }
+      expose :service_id, documentation: { type: String, desc: 'Internal reference of the service' }
+      expose :pickup_shipment_id, documentation: { type: String, desc: 'Internal reference of the shipment' }
+      expose :delivery_shipment_id, documentation: { type: String, desc: 'Internal reference of the shipment' }
       expose :detail, using: VrpResultSolutionRouteActivityDetails, documentation: { desc: '' }
+      expose :type, documentation: { type: String, desc: 'depot, rest, service, pickup or delivery' }
+      expose :current_distance, documentation: { type: Integer, desc: 'Travel distance from route start to current point' }
+      expose :alternative, documentation: { type: Integer, desc: 'When one service has alternative activities, index of the chosen one' }
     end
 
     class VrpResultSolutionRoute < Grape::Entity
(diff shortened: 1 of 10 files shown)


wrappers/demo.rb
```diff
@@ -33,11 +33,15 @@ def build_route_activity(mission, activity)
...
       {
         point_id: activity.point.id,
+        travel_time: 0,
         travel_distance: 0,
         travel_start_time: 0,
-        waiting_duration: 0,
+        waiting_time: 0,
         arrival_time: 0,
         departure_time: 0,
+        type: 'service',
```
braktar, line 42: are we sure it is a service ? It may be a pickup or delivery too

api/v01/entities/vrp_result.rb
```diff
@@ -56,71 +53,83 @@ class VRPResultDetailedCosts < Grape::Entity
...
-      expose :service_id, documentation: { type: String, desc: '' }
-      expose :pickup_shipment_id, documentation: { type: String, desc: '' }
-      expose :delivery_shipment_id, documentation: { type: String, desc: '' }
+      expose :service_id, documentation: { type: String, desc: 'Internal reference of the service' }
+      expose :original_service_id, documentation: { type: String, desc: '' }
+      expose :pickup_shipment_id, documentation: { type: String, desc: 'Internal reference of the shipment' }
+      expose :delivery_shipment_id, documentation: { type: String, desc: 'Internal reference of the shipment' }
+      expose :original_shipment_id, documentation: { type: String, desc: '' }
       expose :detail, using: VrpResultSolutionRouteActivityDetails, documentation: { desc: '' }
+      expose :type, documentation: { type: String, desc: 'Start, end, break, service, pickup or delivery' }
```
braktar, line 69: Start -> start

Have a 'depot' type instead of 'start' or 'end' could allow to have multiple trips within the same route ?

braktar: break should be rest to be consistent with the API and the internal model ?

braktar (changes requested):

api/v01/entities/vrp_result.rb
fonsecadeline: Today API returns 'break', 'start' and 'end' (ortools.rb) that is why I just adapted documentation accordingly. But you are right, we should change these attributes names in version 2.

braktar: I see no previous mention of the field type in the solution returned by ortools.rb before this commit ?
These elements are not part of the solution returned by the API, they are specific to the model used with OR-Tools.
For OR-Tools a route always have a start and a end regardless the vehicle has depots or not.
break is the terminology used by OR-Tools. the parse_output method return a rest_id.

api/v01/entities/vrp_result.rb
```diff
@@ -56,71 +53,83 @@ class VRPResultDetailedCosts < Grape::Entity
...
-      expose :travel_duration, documentation: { type: Integer, desc: 'travel time from the previous point' }
-      expose :waiting_duration, documentation: { type: Integer, desc: '' }
-      expose :arrival_time, documentation: { type: Integer, desc: '' }
+      expose :travel_distance, documentation: { type: Integer, desc: 'Travel distance from previous point' }
+      expose :travel_time, documentation: { type: Integer, desc: 'Travel time from previous point' }
+      expose :travel_value, documentation: { type: Integer, desc: 'Travel value from previous point' }
+      expose :waiting_time, documentation: { type: Integer, desc: '' }
+      expose :begin_time, documentation: { type: Integer, desc: 'Time visit starts' }
+      expose :end_time, documentation: { type: Integer, desc: 'Time visit ends' }
       expose :departure_time, documentation: { type: Integer, desc: '' }
```
senhalil, line 62: do we want to keep `departure_time`? I thought we replace arrival/departure with begin/end, is it not?

api/v01/entities/vrp_result.rb
```diff
@@ -56,71 +53,83 @@ class VRPResultDetailedCosts < Grape::Entity
...
-      expose :service_id, documentation: { type: String, desc: '' }
-      expose :pickup_shipment_id, documentation: { type: String, desc: '' }
-      expose :delivery_shipment_id, documentation: { type: String, desc: '' }
+      expose :service_id, documentation: { type: String, desc: 'Internal reference of the service' }
+      expose :original_service_id, documentation: { type: String, desc: '' }
+      expose :pickup_shipment_id, documentation: { type: String, desc: 'Internal reference of the shipment' }
+      expose :delivery_shipment_id, documentation: { type: String, desc: 'Internal reference of the shipment' }
+      expose :original_shipment_id, documentation: { type: String, desc: '' }
```
senhalil, line 67: Do we have to do this "original" trick? This looks like to me a design problem. 

Why not keeping the `service_id` the way it sent by the user and inside the activity details we give the frequency information like `visit: 5` "5th visit".

api/v01/entities/vrp_result.rb
```diff
@@ -56,71 +53,83 @@ class VRPResultDetailedCosts < Grape::Entity
...
       expose :detail, using: VrpResultSolutionRouteActivityDetails, documentation: { desc: '' }
+      expose :type, documentation: { type: String, desc: 'start, end, break, service, pickup or delivery' }
+      expose :current_distance, documentation: { type: Integer, desc: 'Travel distance from route start to current point' }
+      expose :alternative, documentation: { type: Integer, desc: 'When one service has alternative activities, index of the chosen one' }
     end
 
     class VrpResultSolutionRoute < Grape::Entity
-      expose :vehicle_id, documentation: { type: String, desc: 'Reference of the vehicule used for the current route' }
+      expose :vehicle_id, documentation: { type: String, desc: 'Internal reference of the vehicule used for the current route' }
+      expose :original_vehicle_id, documentation: { type: String, desc: '' }
```
senhalil, line 76: Similar to `service_id`, can't we keep the vehicle_id intact and add the necessary info's (such as day and others) as details. @fab-girard I think this way with the new mt-web date interface, we can upload multiple routes (on different days) of the same vehicle, what do you think?

senhalil (changes requested):
Need to discuss

api/v01/entities/vrp_result.rb
fonsecadeline: It was initially removed but @braktar says this can have an impact for Keesing when they generate their classes.

api/v01/entities/vrp_result.rb
fonsecadeline: I am sorry, you are right ! I got confused because I know I documented _type_ because we already return a type for some elements of _unassigned_ but I did not remember that we never return 'break', 'start' nor 'end'.

Therefore you are right, we could change 'break', 'start' and 'end' with 'rest' and 'store'. To be consistent. But we won't be consistent with ORtools returned values anymore. But I guess the most important is to be consistent for users.

api/v01/entities/vrp_result.rb
fonsecadeline: I think you are right but I am not sure we can do this all at one time because if we change returned ID this could have an impact for users. 

We can and definetly should return _day_index_ for vehicles and _visit_number_ for services. We could also return an additional field _internal_id_ which would be equal to _id_ for now but it would allow us to change returned ids in a v2.

More or less like https://gitlab.com/mapotempo/optimizer-api/-/issues/536

api/v01/entities/vrp_result.rb
braktar: Yes, the API doesn't have to be consistent with OR-Tools, we wrap our own model to the format expected by OR-Tools and convert back the answer to our format.

api/v01/entities/vrp_result.rb
```diff
@@ -56,71 +53,78 @@ class VRPResultDetailedCosts < Grape::Entity
...
+      expose :total_waiting_time, documentation: { type: Integer, desc: 'Sum of every idle time within the route' }
       expose :start_time, documentation: { type: Integer, desc: 'Give the actual start time of the current route if provided by the solve' }
       expose :end_time, documentation: { type: Integer, desc: 'Give the actual end time of the current route if provided by the solver' }
       expose :geometry, documentation: { type: String, desc: 'Contains the geometry of the route, if asked in first place' }
       expose :initial_loads, using: VrpResultDetailQuantities, documentation: { is_array: true, desc: 'Give the actual initial loads of the route' }
       expose :costs, using: VRPResultDetailedCosts, documentation: { desc: 'The impact of the current route within the solution cost' }
     end
 
-    class VrpResultSolutionUnassigneds < Grape::Entity
+    class VrpResultSolutionUnassigned < Grape::Entity
```
braktar, line 86: Changing an object name which is exposed, is always akward as it could be used. Nevertheless, as the documentation is badly generated, and these objects are the unassigned ones, it seems quite ok.

wrappers/ortools.rb
```diff
@@ -551,6 +551,7 @@ def parse_output(vrp, _services, points, _matrix_indices, _cost, _iterations, ou
                   previous_matrix_index = points[vehicle.start_point.id].matrix_index
                   current_activity = {
                     point_id: vehicle.start_point.id,
+                    type: 'store',
```
braktar, line 554: That's the first time, the word 'store' is used to define the depots. Is there a particular reason to change the naming ?
depots are mentioned indirectly multiple times within the documentation and the API code

braktar (changes requested):

api/v01/entities/vrp_result.rb
```diff
@@ -56,71 +53,78 @@ class VRPResultDetailedCosts < Grape::Entity
...
       expose :point_id, documentation: { type: String, desc: 'Linked spatial point' }
-      expose :service_id, documentation: { type: String, desc: '' }
-      expose :pickup_shipment_id, documentation: { type: String, desc: '' }
-      expose :delivery_shipment_id, documentation: { type: String, desc: '' }
+      expose :service_id, documentation: { type: String, desc: 'Internal reference of the service' }
+      expose :shipment_id, documentation: { type: String, desc: 'Internal reference of the shipment' }
```
braktar, line 89: We should keep 'pickup_shipment_id' and 'delivery_shipment_id' to be consistent with the current behaviour of the assigned shipments

api/v01/entities/vrp_result.rb
```diff
@@ -32,14 +32,11 @@ class VrpResultDetailQuantities < Grape::Entity
     end
 
     class VrpResultSolutionRouteActivityDetails < Grape::Entity
-      expose :router_mode, documentation: { type: String, desc: 'Means of transport used to reach this activity, it may vary within a route if subtours are defined' }
```
braktar, line 35: These fields are usefull in the case of multi_modal routes : https://github.com/Mapotempo/optimizer-api/blob/c3e0b1732e85c95e21c0bf8982cc0b4f1704a1e5/lib/interpreters/multi_modal.rb#L152

api/v01/entities/vrp_result.rb
```diff
@@ -32,14 +32,11 @@ class VrpResultDetailQuantities < Grape::Entity
...
-      expose :speed_multiplier, documentation: { type: String, desc: 'Speed multiplier applied to the current means of transport, it may vary within a route if subtours are defined' }
       expose :lat, documentation: { type: Float, desc: '' }
       expose :lon, documentation: { type: Float, desc: '' }
       expose :skills, documentation: { type: Array[String], desc: '' }
       expose :setup_duration, documentation: { type: Integer, desc: '' }
       expose :duration, documentation: { type: Integer, desc: '' }
-      expose :additional_value, documentation: { type: Integer, desc: '' }
```
braktar, line 42: This field is returned by ortools wrapper

braktar (changes requested):

api/v01/entities/vrp_result.rb
fonsecadeline: Shouldn't this rather be in VrpResultSolutionRoute ?

braktar: In the multi modal case, the router_mode can change within the same route

fonsecadeline: Right, I will fix this then !

braktar (approved):

senhalil (approved):

PR #86, opened 2020-11-23
fix wrong number of visits returned

fonsecadeline:


lib/heuristics/concerns/scheduling_end_phase.rb
```diff
@@ -119,7 +119,7 @@ def add_missing_visits
 
   def update_costs(costs, best_cost)
     # update costs for inserted id, available_days changed
-    uninserted_set = @uninserted.keys.select{ |key| key.include?(best_cost[0] + '_') }
+    uninserted_set = @uninserted.keys.select{ |key| key.split('_').slice(0..-3).join('_') == best_cost[0] }
```
senhalil, line 122: If we add more parts to the key `0..-3` might lead to an error, it might be better to do it directly `0..X` where X is the value of `key.size-3`

senhalil (commented):
a suggestion

senhalil (approved):

braktar (approved):

PR #69, opened 2020-09-28
Update Vroom, Grape, Active hash & co

braktar:

PR #89, opened 2020-12-03
Differenciate workload spread and same_point_day option

fonsecadeline:


lib/heuristics/concerns/scheduling_end_phase.rb
```diff
@@ -473,18 +469,18 @@ def reaffect_prohibiting_partial_assignment(still_removed)
         }
       }
 
-      sequences = available_sequences.select{ |sequence| !sequence[:seq].empty? && sequence[:seq].all?{ |day| @candidate_routes[sequence[:vehicle]][day][:current_route].size.positive? } }
+      sequences = available_sequences.select{ |sequence| !sequence[:seq].empty? && sequence[:seq].all?{ |day| @candidate_routes[sequence[:vehicle]][day][:stops].size.positive? } }
```
senhalil, line 472:
```suggestion
      sequences = available_sequences.select{ |sequence| !sequence[:seq].empty? && sequence[:seq].none?{ |day| @candidate_routes[sequence[:vehicle]][day][:stops].empty? } }
```

lib/heuristics/scheduling_heuristic.rb
```diff
@@ -299,34 +298,34 @@ def update_route(full_route, first_index, first_start = nil)
...
+        unassigned << get_unassigned_info(service, service_in_vrp, @uninserted[service][:reason])
       }
 
       unassigned
     end
 
     def provide_group_tws(services, day)
       services.each{ |service|
-        next if !@services_unlocked_by.collect{ |id, set| [id, set] }.flatten.include?(service.id) || # not in a same_point_day_relation
+        next if !@services_unlocked_by.flat_map{ |id, set| [id, set].flatten }.include?(service.id) || # not in a same_point_day_relation
```
senhalil, line 328: Instead of repeatedly creating this `flat_map` for each service can't we create it before the loop and use it inside the loop (even better if this function is called many times during the time where the array `@services_unlocked_by.flat_map{ |id, set| [id, set].flatten }` doesn't change we can create it outside and keep it in the data-structure of the algo)

lib/heuristics/scheduling_heuristic.rb
```diff
@@ -1086,22 +1062,23 @@ def prepare_output_and_collect_routes(vrp)
     def select_point(insertion_costs, empty_route)
       ### chose the most interesting point to insert according to [insertion_costs] ###
       if empty_route
-        if @routes_referents.empty?
+        if @candidate_routes.all?{ |_vehicle_id, all_routes| all_routes.all?{ |_day, route_data| route_data[:stops].empty? } }
           # chose closest with highest frequency
-          to_consider_set = insertion_costs.group_by{ |s| @services_data[s[:id]][:visits_number] }.sort_by{ |nb, _set| nb}.reverse.first[1]
+          to_consider_set = insertion_costs.group_by{ |s| @services_data[s[:id]][:raw].visits_number }.sort_by{ |nb, _set| nb }.reverse.first[1]
```
senhalil, line 1067: we can order correctly instead of `reverse` it
```suggestion
          to_consider_set = insertion_costs.group_by{ |s| @services_data[s[:id]][:raw].visits_number }.sort_by{ |nb, _set| -nb }.first[1]
```

lib/heuristics/scheduling_heuristic.rb
```diff
@@ -1086,22 +1062,23 @@ def prepare_output_and_collect_routes(vrp)
...
         else
           # chose distant service with highest frequency
           # max_priority + 1 so that priority never equal to max_priority and no multiplication by 0
-          return insertion_costs.max_by{ |s| (1 - (@services_data[s[:id]][:priority].to_f + 1) / @max_priority + 1) * @routes_referents.collect{ |ref| matrix(@candidate_routes[@candidate_routes.keys.first].first[1], ref, s[:point]) }.min * @services_data[s[:id]][:visits_number]**2 }
+          referents = @candidate_routes.collect{ |_vehicle_id, all_routes| all_routes.collect{ |_day, route_data| route_data[:stops].empty? ? nil : route_data[:stops].max_by{ |stop| matrix(route_data, route_data[:start_point_id], stop[:point_id]) + matrix(route_data, stop[:point_id], route_data[:end_point_id]) }[:point_id] } }.flatten.compact
+          return insertion_costs.max_by{ |s| (1 - (@services_data[s[:id]][:priority].to_f + 1) / @max_priority + 1) * referents.collect{ |ref| matrix(@candidate_routes[@candidate_routes.keys.first].first[1], ref, s[:point]) }.min * @services_data[s[:id]][:raw].visits_number**2 }
```
senhalil, line 1073: it is really hard to see what's going on here, can we format these two lines so that they are easier to read

senhalil (changes requested):
some performance suggestions and description improvements

lib/heuristics/concerns/scheduling_end_phase.rb
```diff
@@ -30,20 +30,20 @@ def refine_solution(&block)
       add_missing_visits
     end
 
-    unless @services_data.all?{ |_id, data| data[:exclusion_cost].zero? }
+    unless @services_data.all?{ |_id, data| (data[:raw].exclusion_cost || 0).zero? }
```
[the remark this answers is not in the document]
fonsecadeline: Default exclusion_cost is nil. I did not want to change it to zero because I assumed there would be an impact on ortools/dicho side. @braktar is there no impact ?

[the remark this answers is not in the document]
senhalil: With "What would be the consequence?" I was asking you if you can check there is an impact or not in fact..  Anyways, yes there is unfortunately :/ apparently tsp_simple expects "-1" for "no exclusion cost" so we cannot replace it with 0

```ruby
exclusion_cost: service.exclusion_cost && service.exclusion_cost.to_i || -1,
```

So I mark this resolved, this fix would need its own separate PR

lib/heuristics/concerns/scheduling_end_phase.rb
```diff
@@ -337,14 +338,16 @@ def reaffect_in_non_empty_route(still_removed)
...
-        if day_data[:current_route].empty?
-          [vehicle,
+      all_routes.collect{ |day, route_data|
+        if route_data[:stops].empty?
+          [vehicle_id,
            day,
-           stores: [day_data[:start_point_id], day_data[:end_point_id]], time_range: day_data[:tw_end] - day_data[:tw_start]]
+           stores: [
+             route_data[:start_point_id], route_data[:end_point_id]], time_range: route_data[:tw_end] - route_data[:tw_start]
+            ]
```
senhalil, line 350: This formatting looks a bit incorrect :thinking: or it is github making it look incorrect... 

What I meant was something like the following (if that is what we want to collect)
```ruby
[
  vehicle_id,
  day,
  {
    stores: [route_data[:start_point_id], route_data[:end_point_id]], 
    time_range: route_data[:tw_end] - route_data[:tw_start]
  }
]
``` 
it is easier to see the collected items

lib/heuristics/scheduling_heuristic.rb
senhalil: @fonsecadeline  did you forget this one or is there something wrong with what I suggested?

lib/heuristics/scheduling_heuristic.rb
```diff
@@ -1086,22 +1062,23 @@ def prepare_output_and_collect_routes(vrp)
...
+          to_consider_set = insertion_costs.group_by{ |s| @services_data[s[:id]][:raw].visits_number }.max_by{ |nb, _set| nb }[1]
           return to_consider_set.min_by{ |s| ((@services_data[s[:id]][:priority].to_f + 1) / @max_priority) * s[:additional_route_time] }
         else
           # chose distant service with highest frequency
           # max_priority + 1 so that priority never equal to max_priority and no multiplication by 0
-          return insertion_costs.max_by{ |s| (1 - (@services_data[s[:id]][:priority].to_f + 1) / @max_priority + 1) * @routes_referents.collect{ |ref| matrix(@candidate_routes[@candidate_routes.keys.first].first[1], ref, s[:point]) }.min * @services_data[s[:id]][:visits_number]**2 }
+          referents = @candidate_routes.collect{ |_vehicle_id, all_routes| all_routes.collect{ |_day, route_data| find_referent(route_data) } }.flatten.compact
+          return insertion_costs.max_by{ |s| (1 - (@services_data[s[:id]][:priority].to_f + 1) / @max_priority + 1) * distance_to_referents(referents, s).min * @services_data[s[:id]][:raw].visits_number**2 }
```
senhalil, line 1073: this `distance_to_referents` function is used only here.  @fonsecadeline do you plan to use this array somewhere else?

senhalil (commented):
Some followup questions

lib/heuristics/scheduling_heuristic.rb
fonsecadeline: I could rename it to make it more global. `distance_to_set` would be a usefull function soon.

api/v01/vrp.rb
```diff
@@ -184,8 +184,8 @@ def self.vrp_request_vehicle(this)
...
         this.optional :speed_multiplier, type: Float, default: 1.0, desc: 'Multiplies the vehicle speed, default : 1.0. Specifies if this vehicle is faster or slower than average speed.'
-        this.optional :area, type: Array, coerce_with: ->(c) { c.is_a?(String) ? c.split(/;|\|/).collect{ |b| b.split(',').collect{ |f| Float(f) } } : c }, desc: 'List of latitudes and longitudes separated with commas. Areas separated with pipes (only available for truck mode at this time).'
-        this.optional :speed_multiplier_area, type: Array[Float], coerce_with: ->(c) { c.is_a?(String) ? c.split(/;|\|/).collect{ |f| Float(f) } : c }, desc: 'Speed multiplier per area, 0 to avoid area. Areas separated with pipes (only available for truck mode at this time).'
+        this.optional :area, type: Array, coerce_with: ->(c) { c.is_a?(String) ? c.split(/;|\|/).collect{ |b| b.split(',').collect{ |f| Float(f) } } : c }, desc: 'List of latitudes and longitudes separated with commas. Areas separated with pipes (available only for truck mode at this time).'
```
braktar, line 187: Every change here imply both a conflict and a potential omission with this PR : https://github.com/Mapotempo/optimizer-api/pull/69

lib/output_helper.rb
```diff
@@ -120,6 +133,56 @@ def insert_visits(days, inserted_id, nb_visits)
       @scheduling_file << "#{line}\n"
     end
 
+    def compute_route(route_data)
```
braktar, line 136: Can't we use the route_details function with the call to compute_batch to generate the route trace if the geometry is needed ? 
https://github.com/Mapotempo/optimizer-api/blob/c3e0b1732e85c95e21c0bf8982cc0b4f1704a1e5/optimizer_wrapper.rb#L676

braktar (changes requested):

lib/output_helper.rb
fonsecadeline: That is a good idea. However we would need to re-work this function a little bit because, for instance, scheduling would not want to send a route put rather a list of point_ids (by the way I think this function needs a rework anyway in order to use _activity.point.id_ instead of looking for corresponding service/shipment). Also, scheduling algo does not have access to _vrp_ at this point.

I have other modifications regarding output, maybe I can push them all later and taking these comments into account.

api/v01/vrp.rb
fonsecadeline: Unfortunately we are adding an option so you will have to be careful when rebasing anyway. I think there are relatively few modifications here so we should be abble to handle them when rebasing.

lib/heuristics/scheduling_heuristic.rb
senhalil: I think the naming was okay, my comment was related to the fact that we create this array of distances to take the `min` and immediately discard the array. (this is why I asked if this array will be used somewhere else) IF it is not going to be useful somewhere else then instead of doing `collect` inside the `distance_to_referents` function and taking the `min` here, we can make the function so that it returns the `min` directly -- like `min_referent_distance` for example.

fonsecadeline: First I thought it would be useful only here, but then I realized it will be useful for my potential next developments : for instance, we talked about trying to detect if a service is far from others. I do not know yet if I will want to compute average/min or max distance to others. So I guess this is a most generic function for now.

And I did not want to keep _referent_ in function name because the set can contain other services than only referents.

senhalil (approved):

senhalil (approved):

PR #91, opened 2020-12-14
Changes in clustering preprocessing and work_time computation

fonsecadeline:


test/models/vehicle_test.rb
```diff
@@ -25,19 +25,52 @@ def test_work_duration
...
+    def test_total_work_time_in_range
+      vrp = VRP.lat_lon_scheduling
+      vrp = TestHelper.create(vrp)
+      assert_equal 40, vrp.vehicles.first.total_work_time_in_range(0, 3)
+
+      vrp.vehicles.first.ignore_computed_data
+      vrp.vehicles.first.duration = 5
+      assert_equal 20, vrp.vehicles.first.total_work_time_in_range(0, 3)
+
+      vrp.vehicles.first.ignore_computed_data
```
braktar, line 59: The method should be named _reset_computed_data_. It seems more meaningful to me

braktar (changes requested):

lib/interpreters/split_clustering.rb
```diff
@@ -378,7 +378,7 @@ def self.split_balanced_kmeans(service_vrp, nb_clusters, options = {}, &block)
       vrp = service_vrp[:vrp]
       # Split using balanced kmeans
       if vrp.shipments.all?{ |shipment| shipment&.pickup&.point&.location && shipment&.delivery&.point&.location} &&
```
senhalil, line 380:
```suggestion
      if vrp.shipments.all?{ |shipment| shipment&.pickup&.point&.location && shipment&.delivery&.point&.location } &&
```

models/vehicle.rb
```diff
@@ -203,31 +203,48 @@ def total_work_time_in_range(range_start, range_end)
...
+
+        @total_work_time_in_range[[range_start, range_end]] = [total_work_time, 2**32].min
       end
 
       @total_work_time_in_range[[range_start, range_end]]
     end
 
     def work_duration
-      return 2**32 if self.timewindow.nil? && self.sequence_timewindows.empty?
+      raise OptimizerWrapper::DiscordantProblemError, 'Vehicle should not have sequence timewindow if there is no schedule' unless self.sequence_timewindows.empty?
```
senhalil, line 236: why do we need a raise here, to prevent wrong usage of the function during dev? If that is the case are we sure that we have a test that would trip this raise, otherwise, it might slip into prod.

models/vehicle.rb
```diff
@@ -203,31 +203,48 @@ def total_work_time_in_range(range_start, range_end)
...
-          occurence = group[1].size
-          # TODO : fix case where serveral timewindows correspond to the same day
-          tw_index = working_week_days.find_index(week_day_index)
-          tw = self.timewindow
-          tw ||= self.sequence_timewindows[tw_index]
-          total_work_time += (tw.end - tw.start) * occurence
-        }
-        @total_work_time_in_range[[range_start, range_end]] = total_work_time
+        if self.timewindow.nil? && self.sequence_timewindows.empty?
+          total_work_time = working_day_indices_in_range(range_start, range_end).size * [2**32, self.duration].compact.min
```
senhalil, line 207: This would work right?
```suggestion
          total_work_time = working_day_indices_in_range(range_start, range_end).size * (self.duration || 2**32)
```

braktar (approved):

senhalil (approved):

models/vehicle.rb
fonsecadeline: Yes this was only meant to be for us during dev. You are right about the test. We do have one in vehicle_trips.rb .

PR #95, opened 2020-12-18
Small changes in test

fonsecadeline:
Fixes one test that frequently breaks because of too hard limit and adds a test.

### DIFF
--- a/test/lib/heuristics/scheduling_functions_test.rb
+++ b/test/lib/heuristics/scheduling_functions_test.rb
@@ -536,5 +536,25 @@ class HeuristicTest < Minitest::Test
         assert_equal 0, vrp.services.first.last_possible_days.first, "There are #{visits_number} working days, hence a service with #{visits_number} visits can only start at day 0"
       }
     end
+
+    def test_compute_latest_authorized_day_2
+      vrp = VRP.scheduling
+      vrp[:vehicles].first.delete(:timewindow)
+      vrp[:vehicles].first[:sequence_timewindows] = [
+        { start: 0, end: 20, day_index: 0 },
+        { start: 0, end: 20, day_index: 1 },
+        { start: 0, end: 20, day_index: 2 },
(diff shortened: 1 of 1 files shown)


braktar (approved):

test/wrapper_test.rb
```diff
@@ -3060,7 +3060,7 @@ def test_detecting_unfeasible_services_can_not_take_too_long
       }
     end
 
-    assert_operator total_time, :<=, 1.5, 'check_distances function took longer than expected'
+    assert_operator total_time, :<=, 15, 'check_distances function took longer than expected'
```
senhalil, line 3063: @fonsecadeline did you check the performance of this test? I suspect this modification might be incorrect. If we are losing 10x perf, there is a regression somewhere.

Maybe you didn't realize the test has changed? We now set the repetition to 1 and set the log level to fatal. So the performance is stable and 1 repeat decreases the time. This test passed regularly on local and on travis for mapotempo/dev when I was pushing for maj-fixes. 
```
    OptimizerLogger.level = :fatal # turn off output completely no matter the setting
    OptimizerWrapper.config[:solve][:repetition] = 1 # fix repetition to measure the perf correctly
```

senhalil (changes requested):
It looks like you are also pushing the commit of PR93 but that commit is already merged :thinking: Forgot to rebase maybe?

Also I think there is problem with the limit of the performance test -- test_detecting_unfeasible_services_can_not_take_too_long

test/wrapper_test.rb
fonsecadeline: I realized the test changed but I did not realize limit has also changed (which makes sense ...)! This is an old PR and we used to have an issue on travis because previous version (with previous limit) was too strict so I thought I still needed to fix it. 
But you are right, the tests passes everytime on local so great ! 

I will revert this modif then :)

senhalil (approved):

PR #97, opened 2020-12-22
Bump to ruby 2.5 using passenger image

alain-andre:
This PR makes Docker use passenger instead of rake. It fixes the warning message about some libs incompatibilities with optimizer-ortools images since this one is now based on.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,5 +1,5 @@
 source 'https://rubygems.org'
-ruby '2.3.3'
+ruby '~> 2.5.5'
 
 gem 'require_all'
 
--- a/ci-utils/tests.sh
+++ b/ci-utils/tests.sh
@@ -3,7 +3,6 @@ TEST_ENV=''
 TEST_LOG_LEVEL='info'
 TEST_COVERAGE='false'
 DOCKER_SERVICE_NAME=optimizer_api
-CONTAINER=${DOCKER_SERVICE_NAME}.1.$(docker service ps -f "name=${DOCKER_SERVICE_NAME}.1" ${DOCKER_SERVICE_NAME} -q --no-trunc | head -n1)
 
 case "$1" in
   'basis')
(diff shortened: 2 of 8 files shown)


ci-utils/tests.sh
```diff
@@ -31,21 +30,37 @@ case "$1" in
...
+  fi
+
+  nbps=$(docker service ls | grep 1/1 | awk '{print $4}' | wc -l)
+  if [ ${nbps} -eq ${total_dockercompose_services} ];
+  then
+    echo "All services up, starting tests"
+    break;
+  fi
+
+  echo "Waiting for all services (${nbps}/4) to start ($cpt secondes)."
```
senhalil, line 56: why do we need this logic?

docker/docker-compose.yml
```diff
@@ -1,5 +1,23 @@
...
+
+  redis-cache:
+    image: redis:${REDIS_VERSION:-3.2-alpine}
+    hostname: travis-redis-cache
+    volumes:
+      - ../redis:/data
+    command: redis-server --save ""
+    networks:
+      - redis_cache
+
```
senhalil, line 20: As far as I can see, there is no change, do we need this cut and paste?  Is docker-compose.yml order dependent?

senhalil (commented):
Some questions

ci-utils/tests.sh
alain-andre: This logic handles all stack errors and gives the process status and logs of failing service. Before this, we only was checking api start. I had test running for a long time without results because of redis that was not able to launch.

docker/docker-compose.yml
alain-andre: Since some days, Redis is taking around 50 secondes to start in Travis, it's better for the stack to launch them first.

Gemfile
```diff
@@ -1,5 +1,5 @@
 source 'https://rubygems.org'
-ruby '2.3.3'
+ruby '~>2.5.5'
```
senhalil, line 2:
```suggestion
ruby '~> 2.5.5'
```

senhalil (changes requested):

Gemfile
alain-andre: Sorry, I did not see it was a suggestion, I pushed the modified commit.

senhalil (approved):
Thanks Alain! :pray:

PR #98, opened 2020-12-24
[optim-ortools] fixes for maj

senhalil:
Needs https://github.com/Mapotempo/optimizer-ortools/pull/19 to be merged first

Remaining fixes for maj

- The cost error is corrected by moving the quantity multiplication logic to optim-ortools so that both final and intermediate costs will always be correct on optim-api side and inside ortools
- Other small fixes for tests


senhalil:
Tests are cancelled because without this PR needs https://github.com/Mapotempo/optimizer-ortools/pull/19 to be merged first

PR #100, opened 2021-01-05
Empty result with scheduling

fonsecadeline:


fonsecadeline:
Closes https://gitlab.com/mapotempo/optimizer-api/-/issues/657

braktar (approved):

senhalil (approved):

braktar (approved):

braktar:
I will probably generate some conflicts with #107 due to the introduction of scheduling?

models/vrp.rb
```diff
@@ -571,5 +575,13 @@ def calculate_service_exclusion_costs(type = :time, force_recalc = false)
     def transactions
       vehicles.count * points.count
     end
+
+    def scheduling?
+      !self.schedule_range_indices.nil?
+    end
+
+    def periodic?
```
fab-girard, line 583: should we call this `periodic_heuristic?` instead? because we can have periodic vehicle routing problem without this heuristic, only if a user specifies `visits_number` in vrp...

PR #102, opened 2021-01-11
Provide & respect specified units in result

fonsecadeline:
One of our customers notices than, from scheduling heuristic, routes' _start_time_ and _end_time_ could be either integers or floatting numbers.

### DIFF
--- a/optimizer_wrapper.rb
+++ b/optimizer_wrapper.rb
@@ -653,13 +653,17 @@ module OptimizerWrapper
     end
   end
 
-  def self.compute_result_total_dimensions(result)
+  def self.compute_result_total_dimensions_and_round_route_stats(result)
     [:total_time, :total_travel_time, :total_travel_value, :total_distance, :total_waiting_time].each{ |stat_symbol|
       next unless result[:routes].all?{ |r| r[stat_symbol] }
 
       result[stat_symbol] = result[:routes].collect{ |r|
         r[stat_symbol]
       }.reduce(:+)
+    }
+
+    result[:routes].each{ |r|
+      round_route_stats(r)
     }
   end
 
@@ -800,10 +804,8 @@ module OptimizerWrapper
         details = route_details(vrp, r, v) if details.nil?
         r[:geometry] = details.map(&:last) if details
       end
-
-      round_route_stats(r)
     }
-    compute_result_total_dimensions(result)
+    compute_result_total_dimensions_and_round_route_stats(result)
 
     log "result - unassigned rate: #{result[:unassigned].size} of (ser: #{vrp.visits}, ship: #{vrp.shipments.size}) (#{(result[:unassigned].size.to_f / (vrp.visits + 2 * vrp.shipments.size) * 100).round(1)}%)"
     used_vehicle_count = result[:routes].count{ |r| r[:activities].any?{ |a| a[:service_id] || a[:pickup_shipment_id] } }
--- a/optimizer_wrapper.rb
+++ b/optimizer_wrapper.rb
@@ -603,6 +603,23 @@ module OptimizerWrapper
     end
   end
 
+  def self.round_route_stats(route)
+    [:end_time, :start_time].each{ |key|
+      next unless route[key]
+
+      route[key] = route[key].round
+    }
+
+    route[:activities].each{ |activity|
+      [:begin_time, :current_distance, :departure_time, :end_time,
+       :travel_distance, :travel_time, :travel_value, :waiting_time].each{ |key|
+        next unless activity[key]
+
+        activity[key] = activity[key].round
+      }
+    }
+  end
+
   def self.compute_route_total_dimensions(vrp, route, matrix)
     previous = nil
     dimensions = []
@@ -612,18 +629,8 @@ module OptimizerWrapper
 
     total = dimensions.collect.with_object({}) { |dimension, hash| hash[dimension] = 0 }
     route[:activities].each{ |activity|
-      # TODO: This next operation is expensive for big instances. Is there a better way?
-      point_id = activity[:point_id]
-      point_id ||= if activity[:service_id]
-        vrp.services.find{ |s| s.id == activity[:service_id] }.activity.point_id
-      elsif activity[:pickup_shipment_id]
-        vrp.shipments.find{ |s| s.id == activity[:pickup_shipment_id] }.pickup.point_id
-      elsif activity[:delivery_shipment_id]
-        vrp.shipments.find{ |s| s.id == activity[:delivery_shipment_id] }.pickup.point_id
-      end
-
-      if point_id
-        point = vrp.points.find{ |p| p.id == point_id }.matrix_index
+      if activity[:point_id]
+        point = vrp.points.find{ |p| p.id == activity[:point_id] }.matrix_index
         if previous && point
           dimensions.each{ |dimension|
             activity[('travel_' + dimension.to_s).to_sym] = matrix.send(dimension)[previous][point]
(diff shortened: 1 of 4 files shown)


senhalil (changes requested):
round suits better than to_i for float -> integer conversions.

--

@fonsecadeline  I wonder if we have the same issue with other solvers and if there is a more "centralized" location (closer to the "end" of the API) to make this conversion so that it will be corrected for all current and future solvers.  (this is just an idea but), inside merge_results for example :thinking:

fonsecadeline:
@senhalil you are right. I first thought _ortools_result_pb.rb_ would ensure each type to correspond to expected type but looking closer to this file it is not the case. There are post-precossing to try to ensure this.

I will then move this logic to _parse_result_ function in _optimizer_wrapper.rb_ file which I think is the most consistent. What do you think ?

fonsecadeline:
This would help https://gitlab.com/mapotempo/optimizer-api/-/issues/655 even though it will be better when we use a Result model.

fonsecadeline:
_cleanse_ function could also be helpfull for that. What do you think @braktar  ?

senhalil:
> @senhalil you are right. I first thought _ortools_result_pb.rb_ would ensure each type to correspond to expected type but looking closer to this file it is not the case. There are post-precossing to try to ensure this.
> 
> I will then move this logic to _parse_result_ function in _optimizer_wrapper.rb_ file which I think is the most consistent. What do you think ?

If you are sure that parse_result is always called for each independent vrp then I think it's perfect :ok_hand:  (I wonder what happens with the for intermediate results? If they "touch" this function or not :thinking: )

senhalil:
I think cleanse might not be the best place for this type of operation, especially considering the fact that it is there specifically for managing Empty-Fill type services. Its name is unfortunately too generic considering its specificity.

fonsecadeline:
@senhalil perfect, I will work on _parse_result_ then. This function is called every time solve (from _optimizer_wrapper.rb_) is called. Therefore every intermediate result will call that if by intermediate result we think about the same thing. I will only need to add it in the case we return empty_result (which is where these modifications will be the more usefull).

Thanks!

braktar:
I agree with Halil, cleanse is dedicated to the management of empty and fills. parse_result is then the most fitted place to round these values

senhalil:
> @senhalil perfect, I will work on _parse_result_ then. This function is called every time solve (from _optimizer_wrapper.rb_) is called. Therefore every intermediate result will call that if by intermediate result we think about the same thing. 

Perfect! Cheers.

> I will only need to add it in the case we return empty_result (which is where these modifications will be the more usefull).

Hmm, I might have misunderstood this sentence but if you are thinking about a conditional rounding (i.e., only for empty results) then maybe we have different ideas. What I had in mind was a more generic (non-conditional) rounding in a centralized place so that all solutions (empty, non-empty, intermediate, complete, etc) touch this place before reaching the user.

fonsecadeline:
@senhalil I think we are on the same page . I meant I will need to add a call to _parse_result_ in the case we generate _empty_result_. But rounding will be in _parse_result_.

senhalil:
We are definitely very close I think.  Do you think that call to parse_result (for empty_results) is necessary on its own right? In that case :ok_hand: 

If that is not the case and if you think that empty_result actually doesn't need parse_result then what I had in mind was more like for example (even though it is not the best place) merge_results function. Because we know that every result touches this function just before exiting the define_main_process, even if there is one single vrp or multiple, even if the result is empty or full, the logic is generic and aggregates the result and it is the last function so we are sure that the user will get the correct type.

But like I said if empty_results was the source of the problem then fixing there would definitely make sense!

fonsecadeline:
I would say it is safer if we do call _parse_result_ on _empty_result_.

Indeed, this function provides fields that other non-empty results will have : total_time, total_distance ... (even though those are going to be zero, I think it is better to provide those fields even for empty results). 

Also, _parse_result_ outputs some infos (such as unassigned rate at this level). If we don't call parse_result then we would need to move this output so we don't get lost in logs while debugging.

I think it makes sense like this, but let me know if you don't agree after reviewing my last changes (I will ask your review again when it is ready).

lib/heuristics/scheduling_heuristic.rb
```diff
@@ -976,20 +976,20 @@ def get_stop(day, vrp, type, data = {})
...
-        travel_time: data[:travel_time],
-        travel_distance: data[:travel_distance],
-        begin_time: data[:arrival],
-        end_time: data[:end],
-        departure_time: data[:end],
+        travel_time: data[:travel_time].to_i,
+        travel_distance: data[:travel_distance].to_i,
+        begin_time: data[:arrival].to_i,
+        end_time: data[:end].to_i,
+        departure_time: data[:end].to_i,
```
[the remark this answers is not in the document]
fonsecadeline: In fact I wanted to deal with the case where one of those values where nil : nil.round does not work whereas nil.to_i returns 0. Put I moved everything in parse_result and dealt with this case properly :)

optimizer_wrapper.rb
```diff
@@ -596,6 +598,23 @@ def self.formatted_duration(duration)
     end
   end
 
+  def self.unify_field_units(route)
```
senhalil, line 601: maybe `round_route_stats` ?

optimizer_wrapper.rb
```diff
@@ -729,27 +751,30 @@ def self.route_details(vrp, route, vehicle)
...
+
   def self.parse_result(vrp, result)
     tic_parse_result = Time.now
     result[:routes].each{ |r|
       details = nil
-      v = vrp.vehicles.find{ |vehicle| vehicle.id == r[:vehicle_id] }
-      if r[:end_time] && r[:start_time]
-        r[:total_time] = r[:end_time] - r[:start_time]
-      end
+      unify_field_units(r)
```
senhalil, line 769: We need to calculate the totals first and then do the rounding of individual routes and stops of these routes. Otherwise, the rounding errors will accumulate.

optimizer_wrapper.rb
```diff
@@ -198,10 +198,12 @@ def self.solve(service_vrp, job = nil, block = nil)
...
       multi_modal = Interpreters::MultiModal.new(vrp, service)
-      optim_result = multi_modal.multimodal_routes
-    elsif vrp.vehicles.empty? || (vrp.services.empty? && vrp.shipments.empty?)
-      optim_result = config[:services][service].empty_result(
+      multi_modal_result = multi_modal.multimodal_routes
+      optim_result = parse_result(vrp, multi_modal_result)
+    elsif vrp.vehicles.empty? || (vrp.services.empty? && vrp.shipments.empty?) # parse result here too
+      empty_result = config[:services][service].empty_result(
         service.to_s, vrp, 'No vehicle available for this service', false)
+      optim_result = parse_result(vrp, empty_result)
```
senhalil, line 206: Can't we make sure that `multimodal_routes` and `empty_result` call `parse_result` so that they generate conforming results and we don't need to "remember calling parse_result each time we call these result functions"

senhalil (changes requested):

braktar (approved):

optimizer_wrapper.rb
senhalil: I have found my comment  @fonsecadeline.  You gave a thumbs up and resolved conversation so I assume you already did the modifications. Lets verify it next week as you suggested.

fonsecadeline: It is okay now :)

senhalil (approved):

PR #104, opened 2021-01-11
Improve scheduling validation

fonsecadeline:


lib/heuristics/scheduling_heuristic.rb
```diff
@@ -1194,5 +1218,17 @@ def restore
       @uninserted = @previous_uninserted
       @candidate_services_ids = @previous_candidate_service_ids
     end
+
+    def reject_all_visits(id, visits_number, specified_reason)
+      visits_number.times.each{ |visit_index|
+        @uninserted["#{id}_#{visit_index + 1}_#{visits_number}"] = {
+          original_service: id,
```
braktar, line 1225: original_service_id ?

lib/heuristics/concerns/scheduling_data_initialisation.rb
```diff
@@ -159,11 +165,13 @@ def adapt_services_data(vrp)
 
       group_tw = best_common_tw(same_located_set)
       if group_tw.empty? && !same_located_set.all?{ |_id, data| data[:tws_sets].first.empty? }
-        reject_group(same_located_set)
+        reject_group(same_located_set, 'Same_point_day conflict : services at this geografical point have no compatible timewindow')
```
braktar, line 168: this geografical point -> this geographical point or these coordinates

lib/heuristics/scheduling_heuristic.rb
```diff
@@ -303,13 +304,41 @@ def update_route(route_data, first_index, first_start = nil)
       route
     end
 
+    def exist_possible_first_route_according_to_same_point_day(service_id, point_id)
+      # TODO : eventually consider unavailable_days here
+      return true unless @unlocked.include?(service_id) || @services_data[service_id][:raw].visits_number == 1
```
braktar, line 309: Did you want to return true if visits_number > 1 ?

braktar (changes requested):

lib/heuristics/scheduling_heuristic.rb
fonsecadeline: you are right, I wrote it wrong here !

lib/heuristics/scheduling_heuristic.rb
braktar: I was pointing `original_service` and not `id`

fonsecadeline: This is named _original_service_ all over the code. I will make the clean in a new PR in order to check these modifications do not break anything.

lib/heuristics/scheduling_heuristic.rb
```diff
@@ -303,13 +304,42 @@ def update_route(route_data, first_index, first_start = nil)
       route
     end
 
+    def exist_possible_first_route_according_to_same_point_day(service_id, point_id)
```
braktar, line 307:
```suggestion
    def exist_possible_first_route_according_to_same_point_day?(service_id, point_id)
```
Whenever the returned value is a boolean I would suggest to add a question mark to the method name

braktar (approved):

senhalil (approved):

PR #105, opened 2021-01-13
move tests unconsistent with api tests

fonsecadeline:
This should help @alain-andre with integration.

test/api/v01/vrp_test.rb fail if we do call
`require './api/root'`

Something has been introduced and breaks this call. This should also be fixed.

### DIFF
--- a/test/api/v01/vrp_test.rb
+++ b/test/api/v01/vrp_test.rb
@@ -271,11 +271,11 @@ class Api::V01::VrpTest < Minitest::Test
         TestHelper.solve_asynchronously do
           vrp = VRP.lat_lon_scheduling_two_vehicles
           vrp[:configuration][:preprocessing][:partitions] = TestHelper.vehicle_and_days_partitions
-          @job_id = submit_vrp(api_key: 'ortools', vrp: vrp)
-          wait_status @job_id, 'completed', api_key: 'ortools'
+          @job_id = submit_vrp(api_key: 'demo', vrp: vrp)
+          wait_status @job_id, 'completed', api_key: 'demo'
           output.flush
         end
-        delete_completed_job @job_id, api_key: 'ortools' if @job_id
+        delete_completed_job @job_id, api_key: 'demo' if @job_id
 
         lines_with_avancement = output.grep(/avancement/)
       ensure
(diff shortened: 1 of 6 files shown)


test/models/active_hash_test.rb
```diff
@@ -0,0 +1,43 @@
...
+# or FITNESS FOR A PARTICULAR PURPOSE.  See the Licenses for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with Mapotempo. If not, see:
+# <http://www.gnu.org/licenses/agpl.html>
+#
+require './test/test_helper'
+
+module Models
+  class ActiveHashTest < Minitest::Test
```
senhalil, line 21: I find ActiveHashTest name and filename a bit misleading, because we are not testing ActiveHash itself, plus we may change the gem in the future but the tests of the model will stay. So maybe ModelTest?

senhalil: After close inspection I realized that both of these tests can be moved to `test/models/vrp_test.rb` (not the api but the models) because the tests related to Models are already there.

senhalil: After continuing the PR, I realized that we are splitting the `models/vrp_test.rb` as well... I am not sure if this split is the split we want though because all other tests are still related to vrp.. 

Still these two tests can go into VrpRejectionTest because we are refusing the vrp here.

test/models/vrp_rejection_test.rb
```diff
@@ -0,0 +1,159 @@
...
+# or FITNESS FOR A PARTICULAR PURPOSE.  See the Licenses for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with Mapotempo. If not, see:
+# <http://www.gnu.org/licenses/agpl.html>
+#
+require './test/test_helper'
+
+module Models
+  class VrpRejectionTest < Minitest::Test
```
senhalil, line 21: I am not sure splitting VRP tests into `VrpRejectionTest` and `VrpTest` is a good idea. I believe they belong to one place but I might be wrong. I don't know what the best practice is but I suspect tests belonging the same class usually stay together.

But if we want to go down that road, I think we should move at least tests like `test_reject_if_shipments_and_periodic_heuristic` and `test_reject_if_rests_and_periodic_heuristic` and possibly others to here as well because the logic is exactly the same.

braktar (approved):

senhalil (changes requested):
I think it is good that we move the non-api tests out from `test/api/v01/vrp_test.rb` but I am not sure of the split of `test/models/vrp_test.rb` and the location choice of some tests.

test/models/active_hash_test.rb
fonsecadeline: @fab-girard should we split `models/vrp_test.rb` from `test/models` section ?

fab-girard: Yes as I explained you, the best practice is to have a lot of very small files instead of one big file. Rubocop default class length is 100 for remember! We have set it to 1000 because of too many offenses! There are many advantages:  easier to navigate in text editor, avoid git conflicts, structure code by features, etc...

test/models/vrp_rejection_test.rb
fab-girard: As method on model is called `check_consistency` I suggest you to call this file `vrp_consistency_test.rb`. But as said by Halil, you had to merge `scheduling_instance_validity_test.rb ` in this file.

fonsecadeline: This makes sense but at the same time it is contradictory to what we discussed : dividing files in order to have smaller tests files. I think this is a big subject to discuss, to make things properly and with one defined guide line.

As @senhalil suggested I think this PR should only move non api tests from test/api/v01/vrp_tests.rb.

fab-girard: I don't think it's a BIG subject. Please don't waste time on it. You can split multiple files in this one PR.
In addition (but this can be performed in another PR), `check_consistency` method should be split in a concern (separated file) from the model.

senhalil: I tried to demonstrate in my analysis that we have bigger fishes to catch. That's why I suggested that we fix the error of non-api tests in this PR quickly (since it is an error) then decide what to do with other files (in a general fashion) because this is clearly not an issue with one single test file or one single class. It is a prevalent problem in our codebase. Dividing test files hastily will not resolve the main problem and it will not help us in the long run. That's why I suggest a separate proper rework PR.

I think our main problem is that our models are "empty", we can't have useful tests "for" them. For example models/vehicles.rb, models/services.rb or models/units.rb, they don't have "their" validation logic in them as member functions. So when we need to test "detect_unfeasible_service" or "check_consistency" we are forced to create a VRP and either post it or call wrapper_vrp on it. THIS is the source of the issue which leads to big test files.

braktar (approved):

senhalil (approved):

braktar (approved):

PR #106, opened 2021-01-13
Ci default values

alain-andre:
Set default value in `Dockerfile` instead of `.travis.yml` It allows to propagate the versions to all branches.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -4,10 +4,10 @@ ARG REGISTRY=${REGISTRY:-registry.mapotempo.com/}
 ARG VROOM_VERSION=${VROOM_VERSION}
 
 # Install Vroom
-FROM ${REGISTRY}mapotempo/vroom:${VROOM_VERSION} as vroom
+FROM ${REGISTRY}mapotempo/vroom:${VROOM_VERSION:-v1.5.0} as vroom
 
 # Rake
-FROM ${REGISTRY}mapotempo-${BRANCH}/optimizer-ortools:${OPTIMIZER_ORTOOLS_VERSION}
+FROM ${REGISTRY}mapotempo-${BRANCH}/optimizer-ortools:${OPTIMIZER_ORTOOLS_VERSION:-v1.4.0}
 
 ENV LANG C.UTF-8
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ services:
 before_install:
   - export REGISTRY=${REGISTRY:-registry.mapotempo.com/}
   - source ci-utils/version.sh
-  - docker build --build-arg VROOM_VERSION=${VROOM_VERSION:-v1.5.0} --build-arg OPTIMIZER_ORTOOLS_VERSION=$OPTIMIZER_ORTOOLS_VERSION -f docker/Dockerfile -t registry.test.com/mapotempo/optimizer-api:latest .
+  - docker build --build-arg VROOM_VERSION=${VROOM_VERSION} --build-arg OPTIMIZER_ORTOOLS_VERSION=${OPTIMIZER_ORTOOLS_VERSION} -f docker/Dockerfile -t registry.test.com/mapotempo/optimizer-api:latest .
   - docker swarm init
   - mkdir -p ./redis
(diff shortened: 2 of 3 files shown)


fonsecadeline (approved):

docker/Dockerfile
```diff
@@ -4,10 +4,10 @@ ARG REGISTRY=${REGISTRY:-registry.mapotempo.com/}
 ARG VROOM_VERSION=${VROOM_VERSION}
 
 # Install Vroom
-FROM ${REGISTRY}mapotempo/vroom:${VROOM_VERSION} as vroom
+FROM ${REGISTRY}mapotempo/vroom:${VROOM_VERSION:-v1.5.0} as vroom
```
fonsecadeline, line 7: This will depend on https://github.com/Mapotempo/optimizer-api/pull/107 integration

senhalil: #107 uses the new functionality but we can do this without #107 for the existing Vroom functionality. Actually it was already the case in Travis.yml -- that is everything is tested with 1.5.0 since quite some time -- and 1.5.0 is validated for existing functionality.

docker/Dockerfile
```diff
@@ -4,10 +4,10 @@ ARG REGISTRY=${REGISTRY:-registry.mapotempo.com/}
...
 
 # Rake
-FROM ${REGISTRY}mapotempo-${BRANCH}/optimizer-ortools:${OPTIMIZER_ORTOOLS_VERSION}
+FROM ${REGISTRY}mapotempo-${BRANCH}/optimizer-ortools:${OPTIMIZER_ORTOOLS_VERSION:-v1.4.0}
```
senhalil, line 10: There is a file called `version.sh` which also has the default for `OPTIMIZER_ORTOOLS_VERSION` environment variable. 

So the default of optimizer-ortools will be in two places at the moment. I wonder if there is an easy way to simplify this while keeping the functionality coming from there (i.e., devs can change the OPTIMIZER_ORTOOLS_VERSION if they want)..

@alain-andre if we modify the content of `version.sh` as follows, Docker image generation and Travis should continue to work correctly, right ?
```bash
#!/usr/bin/env bash

if ! { [[ $TRAVIS_BRANCH == "master" ]] || [[ $TRAVIS_TAG != "" ]]; }; then
  export OPTIMIZER_ORTOOLS_VERSION=${OPTIMIZER_ORTOOLS_VERSION:-Mapotempo};
fi
```

senhalil (changes requested):

docker/Dockerfile
alain-andre: This incoming PR will change the version. The Important thing is that it's allocated here.

docker/Dockerfile
alain-andre: Yes @senhalil you are right. But I would revert the conditions to set default `OPTIMIZER_ORTOOLS_VERSION` to `Mapotempo` only when *$TRAVIS_BRANCH* is not *master* or *$TRAVIS_TAG* is empty. It was written by @braktar he could have an idea.

senhalil: I just forgot to inverse the middle OR :facepalm: ! 

The correct condition is as follows
```bash
if ! { [[ $TRAVIS_BRANCH == "master" ]] && [[ $TRAVIS_TAG != "" ]]; }; then
  export OPTIMIZER_ORTOOLS_VERSION=${OPTIMIZER_ORTOOLS_VERSION:-Mapotempo};
fi
```

which keeps the same functionality of the current code (below) while moving the default (1.4.0) to docker file.

```bash
if [[ $TRAVIS_BRANCH == "master" ]] || [[ $TRAVIS_TAG != "" ]]; then
  export OPTIMIZER_ORTOOLS_VERSION="v1.4.0";
else
  export OPTIMIZER_ORTOOLS_VERSION=${OPTIMIZER_ORTOOLS_VERSION:-Mapotempo};
fi
```

alain-andre: wow, I did not noticed `! { ... }`. I change my PR.

senhalil (approved):
I am really happy that the defaults are now in one place and automatic for the image generation :partying_face:  Thanks @alain-andre!

braktar (approved):

PR #107, opened 2021-01-13
Update Vroom, Grape, Active hash & co

braktar:
Replace PR #69


test/lib/heuristics/scheduling_test.rb
```diff
@@ -720,5 +720,22 @@ def test_global_formula_to_find_original_id_back
...
+        start_point_id: 'point_0',
+        timewindow: {
+          start: 0,
+          end: 20
+        }
+      }
+      vrp[:services].each{ |s| s[:sticky_vehicle_ids] = ['vehicle_0'] }
+      vrp[:matrices].first[:distance] = vrp[:matrices].first[:time]
+      result = OptimizerWrapper.wrapper_vrp('ortools', { services: { vrp: [:ortools] }}, TestHelper.create(vrp), nil)
+      assert result[:costs] # TODO: Verify costs content whenever it is correctly returned by scheduling heuristic
```
braktar, line 738: should be cost_details

wrappers/demo.rb
```diff
@@ -72,6 +72,7 @@ def build_route_depot(point)
     def solve(vrp, _job = nil, _thread_proc = nil, &_block)
       {
         cost: 0,
+        costs: Models::CostDetails.new({}),
```
braktar, line 75: should be cost_details

wrappers/jsprit.rb
```diff
@@ -389,6 +389,7 @@ def parse_output(path, iterations, fleet, vrp)
       if solution
         {
           cost: Float(solution.at_xpath('cost').content),
+          costs: Models::CostDetails.new({}), # TODO: fulfill with solution costs
```
braktar, line 392: should be cost_details

braktar (commented):
cost_details is missing from some wrappers and vrp_result

api/v01/entities/vrp_input.rb
```diff
@@ -0,0 +1,522 @@
...
+    optional(:timewindow_start_day_shift_number, documentation: { hidden: true }, type: Integer, desc: '[ DEPRECATED ]')
+    requires(:point_id, type: String, allow_blank: false, desc: 'Reference to the associated point')
+    optional(:timewindow_ids, type: String, documentation: { hidden: true }, desc: 'Timewindows to consider, CSV front only')
+    optional(:timewindows, type: Array, desc: 'Time slot while the activity may start') do
+      use :vrp_request_timewindow
+    end
+    mutually_exclusive :timewindow_ids, :timewindows
+  end
+
+  params :vrp_request_capacity do
```
fonsecadeline, line 386: Move to VrpVehicles ?

api/v01/entities/vrp_input.rb
```diff
@@ -0,0 +1,522 @@
...
+      use :vrp_request_quantity
+    end
+    mutually_exclusive :quantity_ids, :quantities
+  end
+end
+
+module VrpShared
+  extend Grape::API::Helpers
+
+  params :vrp_request_activity do
```
fonsecadeline, line 371: Move to VrpMissions ?

api/v01/entities/vrp_input.rb
```diff
@@ -0,0 +1,522 @@
...
+    requires(:id, type: String, allow_blank: false)
+    optional(:matrix_index, type: Integer, allow_blank: false, desc: 'Index within the matrices, required if the matrices are already given')
+    optional(:location, type: Hash, allow_blank: false, documentation: { desc: 'Location of the point if matrices are not given' }) do
+      self.requires(:lat, type: Float, allow_blank: false, desc: 'Latitude coordinate')
+      self.requires(:lon, type: Float, allow_blank: false, desc: 'Longitude coordinate')
+    end
+    at_least_one_of :matrix_index, :location
+  end
+
+  params :vrp_request_quantity do
```
fonsecadeline, line 404: Move to VrpMissions ?

api/v01/entities/vrp_input.rb
```diff
@@ -0,0 +1,522 @@
...
+    optional(:direct, type: Boolean, default: false, desc: 'When activated, vehicle should go directly to delivery point after pick up')
+    optional(:quantity_ids, type: String, documentation: { hidden: true }, desc: 'Quantities to consider, CSV front only')
+    optional(:quantities, type: Array, desc: 'Define the entities which are taken and dropped') do
+      use :vrp_request_quantity
+    end
+    mutually_exclusive :quantity_ids, :quantities
+  end
+end
+
+module VrpShared
```
fonsecadeline, line 368:
```suggestion
module VrpCommon
```

api/v01/entities/vrp_input.rb
```diff
@@ -0,0 +1,522 @@
...
+    optional(:cost_distance_multiplier, type: Float, desc: 'Cost applied to the distance performed')
+
+    optional(:cost_time_multiplier, type: Float, desc: 'Cost applied to the total amount of time of travel (Jsprit) or to the total time of route (ORtools). Not taken into account within periodic heuristic.')
+    optional(:cost_value_multiplier, type: Float, desc: 'Multiplier applied to the value matrix and additional activity value. Not taken into account within periodic heuristic.')
+    optional(:cost_waiting_time_multiplier, type: Float, desc: 'Cost applied to the waiting time in the route. Not taken into account within periodic heuristic.')
+    optional(:cost_late_multiplier, type: Float, desc: 'Cost applied if a point is delivered late (ORtools only). Not taken into account within periodic heuristic.')
+    optional(:cost_setup_time_multiplier, type: Float, desc: 'Cost applied on the setup duration (Jsprit only). Not taken into account within periodic heuristic.')
+  end
+
+  params :vehicle_model_related do
```
fonsecadeline, line 475: Why is this specific ?

docker/Dockerfile
```diff
@@ -1,12 +1,11 @@
 ARG BRANCH=${BRANCH:-ce}
 ARG OPTIMIZER_ORTOOLS_VERSION=${OPTIMIZER_ORTOOLS_VERSION}
 ARG REGISTRY=${REGISTRY:-registry.mapotempo.com/}
-ARG VROOM_VERSION=${VROOM_VERSION}
+ARG VROOM_VERSION=${VROOM_VERSION:-v1.8.0}
 
 # Install Vroom
-FROM ${REGISTRY}mapotempo/vroom:${VROOM_VERSION} as vroom
+FROM vroomvrp/vroom-docker:${VROOM_VERSION} as vroom
```
fonsecadeline, line 7: This will have a conflict with https://github.com/Mapotempo/optimizer-api/pull/106 if it is merged before.

We need to chose which of these PR we want to merge first.

test/api/v01/output_test.rb
```diff
@@ -265,21 +265,21 @@ def test_returned_keys_json_result
         problem: VRP.lat_lon,
         solver_name: 'vroom',
         expected_route_keys: %w[vehicle_id original_vehicle_id activities total_travel_time total_distance total_time start_time end_time],
-        expected_activities_keys: %w[point_id travel_distance travel_time travel_value service_id original_service_id detail current_distance type],
+        expected_activities_keys: %w[point_id travel_distance travel_time travel_value begin_time end_time service_id original_service_id detail current_distance type departure_time],
         expected_unassigned_keys: %w[point_id service_id original_service_id detail type reason]
       },
       ortools: {
         problem: VRP.lat_lon_capacitated,
```
fonsecadeline, line 272: We need to change the problem to make sure we call _ortools_ because _vroom_ output is already tested in previous case.

test/api/v01/output_test.rb
```diff
@@ -265,21 +265,21 @@ def test_returned_keys_json_result
...
+        expected_route_keys: %w[vehicle_id original_vehicle_id activities total_travel_time total_distance total_time start_time end_time],
+        expected_activities_keys: %w[point_id travel_distance travel_time travel_value begin_time end_time service_id original_service_id pickup_shipment_id delivery_shipment_id original_shipment_id detail current_distance type departure_time],
+        expected_unassigned_keys: %w[point_id service_id original_service_id detail type reason]
       },
       periodic_ortools: {
         problem: VRP.lat_lon_two_vehicles,
-        solver_name: 'ortools',
-        expected_route_keys: %w[vehicle_id original_vehicle_id activities total_travel_time total_distance total_time total_waiting_time start_time end_time costs initial_loads],
-        expected_activities_keys: %w[point_id travel_distance travel_time travel_value waiting_time begin_time end_time departure_time service_id original_service_id detail current_distance alternative type],
+        solver_name: 'vroom',
```
fonsecadeline, line 280: Is VROOM really able to solve scheduling problems when we do not call periodic heuristic? (i.e. problems with more than one visit per service)

test/api/v01/vrp_test.rb
```diff
@@ -77,6 +77,15 @@ def test_ignore_unknown_parameters
     submit_vrp api_key: 'demo', vrp: vrp
   end
 
+  def test_refute_solver_and_solver_parameter
+    vrp = VRP.toy
+    vrp[:configuration][:resolution][:solver] = true
+    vrp[:configuration][:resolution][:solver_parameter] = 0
```
fonsecadeline, line 83: Do we really want to reject this case ? These are not uncompatible.

wrappers/vroom.rb
```diff
@@ -29,31 +29,42 @@ def initialize(hash = {})
...
+        # Mission constraints
+        :assert_no_direct_shipments,
+        :assert_no_exclusion_cost,
+        :assert_no_setup_duration,
+        :assert_services_no_late_multiplier,
+
+        # Solver
         :assert_no_first_solution_strategy,
+        :assert_no_free_approach_or_return,
+        :assert_no_planning_heuristic,
```
fonsecadeline, line 65: :assert_no_several_visits too ?

wrappers/vroom.rb
```diff
@@ -62,7 +73,7 @@ def solve_synchronous?(_vrp)
     end
 
     def solve(vrp, job = nil, _thread_proc = nil)
-      if vrp.points.empty? || vrp.services.empty?
+      if vrp.points.empty? || vrp.services.empty? && vrp.shipments.empty?
```
fonsecadeline, line 76: Isn't this in fact already managed by OptimiezWrapper::solve ?

fonsecadeline: Use Wrappers::Wrapper.new.empty_result() ?

fonsecadeline (changes requested):
Some suggestions for Grape validation part + one question : 

Should we not rather update vroom here and only in a second time decide for which cases we now call vroom instead of ortools ? 

I am afraid we do not know how vroom will behave for those new cases. Do we have any data about it ? This should at least require to run massive tests. Also, I am surprised : is vroom able to deal with services/shipments with more than one visit ?

api/v01/entities/vrp_input.rb
braktar: I centralize in VrpShared the objects which are used by multiple others + it helps to comply with rubocop rules on module sizes

api/v01/entities/vrp_input.rb
braktar: These objects are related to other objects of the model and will be interpreted by the model create

test/api/v01/vrp_test.rb
braktar: These parameters were never expected to be send together, `solver_parameter` is now deprecated for a long time and this way we have no ambiguity between these two fields ? 
What should we do with solver = true and solver_parameter = -1 ?

At this purpose we add a mutually_exclusive condition.

wrappers/vroom.rb
braktar: the solve verification is done before the call of detect_unfeasible_services and check_distances for the particular sub problem

braktar: And yes, the empty_result method should be used here

api/v01/entities/vrp_input.rb
braktar: The two words are similar, Shared is very specific and Common is a bit more general

api/v01/entities/vrp_result.rb
```diff
@@ -100,7 +100,7 @@ class VrpResultSolution < Grape::Entity
       expose :heuristic_synthesis, documentation: { type: Hash, desc: 'When first_solution_strategies are provided, sum up of tryied heuristics and their performance.' }
       expose :solvers, documentation: { is_array: true, type: String, desc: 'Solvers used to perform the optimization' }
       expose :cost, documentation: { type: Float, desc: 'The actual cost of the solution considering all costs' }
-      expose :costs, using: VRPResultDetailedCosts, documentation: { desc: 'The detail of the different costs which impact the solution' }
+      expose :cost_details, using: VRPResultDetailedCosts, documentation: { desc: 'The detail of the different costs which impact the solution' }
       expose :iterations, documentation: { type: Integer, desc: 'Total number of iteration performed to obtain the current result' }
       expose :total_distance, documentation: { type: Integer, desc: 'cumulated distance of every route' }
       expose :total_time, documentation: { type: Integer, desc: 'Cumulated time of every route' }
```
fonsecadeline, line 106: we should remove _type_ fields for unassigned and activities.

test/api/v01/output_test.rb
```diff
@@ -302,11 +302,12 @@ def test_returned_keys_json_result
     }]
 
     methods[:periodic_ortools][:problem][:configuration][:first_solution_strategy] = nil
+    dimensions = [:time, :distance]
```
fonsecadeline, line 305: this adds an extra line but is only used once ?

test/api/v01/output_test.rb
```diff
@@ -365,11 +366,12 @@ def test_returned_keys_csv
     expected_unassigned_keys = %w[point_id id type unassigned_reason]
 
     [:ortools, :periodic_ortools].each{ |method| methods[method][:problem][:vehicles].first[:timewindow] = { start: 28800, end: 61200 } }
+    dimensions = [:time, :distance]
```
fonsecadeline, line 369: this adds an extra line but is only used once ?

test/models/vrp_test.rb
```diff
@@ -155,61 +155,6 @@ def test_deduce_sticky_vehicles_if_route_and_clustering
       refute_empty generated_vrp.services[2].sticky_vehicles
     end
 
-    def test_solver_parameter_retrocompatibility
```
fonsecadeline, line 158: Since we keep _solver_parameter_ I think we should also keep this one test.

wrappers/vroom.rb
```diff
@@ -29,31 +29,42 @@ def initialize(hash = {})
...
+        :assert_vehicles_no_force_start,
+        :assert_vehicles_no_late_multiplier_or_single_vehicle,
+        :assert_vehicles_no_overload_multiplier,
+        :assert_vehicles_start_or_end,
+
+        # Mission constraints
+        :assert_no_direct_shipments,
+        :assert_no_exclusion_cost,
+        :assert_no_setup_duration,
+        :assert_services_no_late_multiplier,
```
fonsecadeline, line 60: needs _assert_only_one_visit_ or equivalent

fonsecadeline (changes requested):

test/api/v01/output_test.rb
braktar: This variable is used within every step of the each loop. + It complies with rubocop warns

api/v01/entities/vrp_result.rb
braktar: Why ?

wrappers/wrapper.rb
```diff
@@ -446,10 +464,74 @@ def assert_not_too_many_visits_in_route(vrp)
...
+      vrp.vehicles.empty? || (assert_only_time_dimension(vrp) ^ assert_only_distance_dimension(vrp) ^ assert_only_value_dimension(vrp))
+    end
+
     def assert_no_route_if_schedule_without_periodic_heuristic(vrp)
       vrp.routes.empty? || !vrp.schedule_range_indices || vrp.preprocessing_first_solution_strategy.to_a.include?('periodic')
     end
 
+    # TODO: Need a better way to represent solver preference
+    def assert_small_minimum_duration(vrp)
+      vrp.resolution_minimum_duration.nil? || vrp.vehicles.empty? || vrp.resolution_minimum_duration / vrp.vehicles.size < 5000
```
fonsecadeline, line 520: || vrp.services.empty? && vrp.shipments.empty?  ?

wrappers/wrapper.rb
```diff
@@ -372,6 +386,10 @@ def assert_no_vehicle_free_approach_or_return_if_heuristic(vrp)
       vrp.vehicles.none?{ |vehicle| vehicle[:free_approach] || vehicle[:free_return] } || vrp.preprocessing_first_solution_strategy.to_a.first != 'periodic'
     end
 
+    def assert_no_free_approach_or_return(vrp)
+      vrp.vehicles.none?{ |vehicle| vehicle[:free_approach] || vehicle[:free_return] }
```
fonsecadeline, line 390:
```suggestion
      vrp.vehicles.none?{ |vehicle| vehicle.free_approach || vehicle.free_return }
```

wrappers/vroom.rb
fonsecadeline, line 60:
```suggestion
        :assert_missions_no_late_multiplier,
```
Should we check shipments too?

fonsecadeline (changes requested):

test/wrappers/ortools_test.rb
```diff
@@ -5684,11 +5684,12 @@ def test_force_start_when_there_is_a_service_with_late_tw_begin
     size = 3
 
     # A time matrix that leads to incorrect order in the old force_start implementation
+    array_base = [1]
```
senhalil, line 5687: This is actually the first column (except the first row). So maybe `array_first_column` ?

test/wrappers/vroom_test.rb
```diff
@@ -425,7 +424,6 @@ def test_with_rest_at_the_end
           end: 20000
         },
         rest_ids: ['rest_a'],
-        cost_late_multiplier: 1
```
senhalil, line 428: Let me guess, we weren't calling Vroom because of these costs, is that it?

test/wrappers/vroom_test.rb
```diff
@@ -425,7 +424,6 @@ def test_with_rest_at_the_end
           end: 20000
         },
         rest_ids: ['rest_a'],
```
senhalil, line 426:
```suggestion
        rest_ids: ['rest_a']
```

test/wrappers/vroom_test.rb
```diff
@@ -580,4 +578,182 @@ def test_ensure_total_time_and_travel_info_with_vroom
     assert result[:routes].all?{ |route| route[:activities].empty? || route[:total_travel_time] }, 'At least one route total_travel_time was not provided'
     assert result[:routes].all?{ |route| route[:activities].empty? || route[:total_distance] }, 'At least one route total_travel_distance was not provided'
   end
+
+  def test_shipments
+    vroom = OptimizerWrapper.config[:services][:ortools]
```
senhalil, line 583:
```suggestion
    vroom = OptimizerWrapper.config[:services][:vroom]
```

test/wrappers/vroom_test.rb
```diff
@@ -580,4 +578,182 @@ def test_ensure_total_time_and_travel_info_with_vroom
...
+    result = vroom.solve(vrp, 'test')
+    assert result
+    assert_equal(result[:routes][0][:activities].index{ |activity| activity[:pickup_shipment_id] == 'shipment_0' } + 1, result[:routes][0][:activities].index{ |activity| activity[:delivery_shipment_id] == 'shipment_0' })
+    assert_equal(result[:routes][0][:activities].index{ |activity| activity[:pickup_shipment_id] == 'shipment_1' } + 1, result[:routes][0][:activities].index{ |activity| activity[:delivery_shipment_id] == 'shipment_1' })
+    assert_equal 0, result[:unassigned].size
+    assert_equal 6, result[:routes][0][:activities].size
+  end
+
+  def test_mixed_shipments_and_services
+    vroom = OptimizerWrapper.config[:services][:ortools]
```
senhalil, line 683:
```suggestion
    vroom = OptimizerWrapper.config[:services][:vroom]
```

test/wrappers/vroom_test.rb
```diff
@@ -580,4 +578,182 @@ def test_ensure_total_time_and_travel_info_with_vroom
...
+        }
+      }
+    }
+    vrp = TestHelper.create(problem)
+    result = vroom.solve(vrp, 'test')
+    assert result
+    assert result[:routes][0][:activities].index{ |activity| activity[:pickup_shipment_id] == 'shipment_1' } < result[:routes][0][:activities].index{ |activity| activity[:delivery_shipment_id] == 'shipment_1' }
+    assert_equal 0, result[:unassigned].size
+    assert_equal 5, result[:routes][0][:activities].size
+  end
```
senhalil, line 758: It looks like we copy these 3 tests from `ortools_test.rb` to here. 

I wonder if we cannot move them to a more centralized location (for example `wrapper_solver_test`?) and call both ortools and vroom one after another inside the test? If not, we can maybe extract the problems to TestHelper (if they are completely the same) so that they will be in one location.

wrappers/jsprit.rb
```diff
@@ -389,7 +389,7 @@ def parse_output(path, iterations, fleet, vrp)
       if solution
         {
           cost: Float(solution.at_xpath('cost').content),
-          costs: Models::Costs.new({}), # TODO: fulfill with solution costs
+          cost_details: Models::CostDetails.new({}), # TODO: fulfill with solution costs
```
senhalil, line 392: Why do we still hold on to this file. I am ~100% sure it doesn't work :) If we delete it git history will take good care of it for us. :D

wrappers/wrapper.rb
braktar: `vrp.vehicles.empty?` verification is only present to avoid the division by zero during the solver selection

wrappers/ortools.rb
```diff
@@ -645,21 +660,20 @@ def parse_output(vrp, _services, points, _matrix_indices, _cost, _iterations, ou
...
             original_shipment_id: shipment.original_id,
-            shipment_id: shipment_id.to_s,
+            pickup_shipment_id: shipment_id.to_s,
             type: 'pickup',
             point_id: shipment.pickup.point_id,
             detail: build_detail(shipment, shipment.pickup, shipment.pickup.point, nil, nil, nil)
           }, {
             original_shipment_id: shipment.original_id,
-            shipment_id: shipment_id.to_s,
+            delivery_shipment_id: shipment_id.to_s,
```
senhalil, line 676: I asked a question earlier about this but it is buried under the review I am afraid.

wrappers/ortools.rb
```diff
@@ -774,8 +788,11 @@ def run_ortools(problem, vrp, services, points, matrix_indices, thread_proc = ni
         log 'Job killed', level: :fatal # Keep trace in worker
         raise OptimizerWrapper::JobKilledError
       else # Fatal Error
-        message = if @thread.value == 127
+        message = case @thread.value
+                  when 127
                     'Executable does not exist'
+                  when 137 # Segmentation Fault
+                    'Solver has met an unexpected error'
```
senhalil, line 795: I believe 137 is related to out-of-memory 

Exit Code 137: Indicates failure as container received SIGKILL (Manual intervention or ‘oom-killer’ [OUT-OF-MEMORY])
```suggestion
                  when 137 # Segmentation Fault
                    "Executable killed by 'oom-killer' [OUT-OF-MEMORY]"
```

wrappers/ortools_result_pb.rb
```diff
@@ -15,7 +15,7 @@
     optional :lateness, :int64, 8
   end
 
-  add_message 'ortools_result.Costs' do
+  add_message 'ortools_result.CostDetails' do
```
senhalil, line 18: Doesn't this mean that we need to change the proto on the optimizer-ortools side as well?

wrappers/wrapper.rb
senhalil: @braktar  there are others just above as well. While you are at it, you can do them all maybe?

wrappers/wrapper.rb
senhalil: this is only for "duration" The reason why we check vehicles is that we want to calculate "per vehicle" duration (sent by mt-web for example) and if it is not too high we apply Vroom but if the user asked specifically asked +5 seconds per vehicle then we launch or-tools even if it can be solved by Vroom.

senhalil: Adeline your question made me think of another issue we talked this morning during the daily point. 

@braktar can you look at vroom's performance benchmarks and check if it ever passes 30 seconds. If so at what size and for what type of problem?

We might need to add another assert for vroom to prevent it "accepting" too big VRP's

senhalil (changes requested):
We should have separate at least the Vroom update on another PR :nerd_face:  Oh well there is always the next time. Let's not have a PR of this size ever again!

wrappers/wrapper.rb
braktar: Vroom Handke large dose problems. The limitation would probably come from Matrix computation https://github.com/VROOM-Project/vroom/wiki/Benchmarks

api/v01/entities/vrp_result.rb
```diff
@@ -83,7 +83,7 @@ class VrpResultSolutionRoute < Grape::Entity
       expose :end_time, documentation: { type: Integer, desc: 'Give the actual end time of the current route if provided by the solver' }
       expose :geometry, documentation: { type: String, desc: 'Contains the geometry of the route, if asked in first place' }
       expose :initial_loads, using: VrpResultDetailQuantities, documentation: { is_array: true, desc: 'Give the actual initial loads of the route' }
-      expose :costs, using: VRPResultDetailedCosts, documentation: { desc: 'The impact of the current route within the solution cost' }
+      expose :cost_details, using: VRPResultDetailedCosts, documentation: { desc: 'The impact of the current route within the solution cost' }
```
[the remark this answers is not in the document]
braktar: It is not an input field, if we don't let this version extend over months it's ok

api/v01/vrp.rb
```diff
@@ -36,573 +37,114 @@ def self.call(object, _env)
...
           r.each_key{ |key|
             next unless key.include?('.')
 
             part = key.split('.', 2)
-            r.deep_merge!(part[0] => { part[1] => r[key] })
-            r.delete(key)
+            new_r.deep_merge!(part[0] => { part[1] => r[key] })
+            new_r.delete(key)
           }
+          r = new_r
```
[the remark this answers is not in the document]
braktar: From what I remember, due to recursiveness some keys were not updated by assignation otherwise

api/v01/vrp.rb
```diff
@@ -36,573 +37,114 @@ def self.call(object, _env)
...
 
           json = r['json']
           if json # Open the secret short cut
             r.delete('json')
             r.deep_merge!(JSON.parse(json))
           end
 
-          r
+          r.with_indifferent_access
```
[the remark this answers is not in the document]
braktar: Json comes as symbol, csv as strings and i think tests depends if they come directly or from a submit_vrp. It says to interpreter that both strings or symbols are equivalent

test/real_cases_test.rb
```diff
@@ -185,7 +185,7 @@ def test_vrp_ten_routes_with_rest
       assert result[:routes].sum{ |r| r[:total_travel_time] } < 40500, "Too long travel time: #{result[:routes].sum{ |r| r[:total_travel_time] }}"
 
       # Check elapsed time
-      assert result[:elapsed] < 600, "Too long elapsed time: #{result[:elapsed]}"
+      assert result[:elapsed] < 4500, "Too long elapsed time: #{result[:elapsed]}"
```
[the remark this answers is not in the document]
braktar: OR-tools find the previous solution quickly but find improvments that were not reachable. That explains the time extension

docker/README.md
```diff
@@ -11,18 +11,14 @@ Optimizer requires the two following images that must be manually built.
...
+see https://github.com/VROOM-Project/vroom-docker
 
 ## Build API
 
 ```
 export CI_COMMIT_REF_NAME=latest
 export OPTIMIZER_ORTOOLS_VERSION=v1.1.0
-export VROOM_VERSION=v1.5.0
+export VROOM_VERSION=v1.8.0
 export BRANCH=${BRANCH:-beta}
```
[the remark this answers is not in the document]
braktar: I would suggest to uniformize and remove useless values in another PR, maybe #106

models/vrp.rb
```diff
@@ -190,13 +190,6 @@ def self.check_consistency(hash)
...
       end
 
-      if configuration[:resolution]
-        if configuration[:resolution][:solver] && configuration[:resolution][:solver_parameter].to_i == -1 ||
-           !configuration[:resolution][:solver].nil? && configuration[:resolution][:solver] == false && configuration[:resolution][:solver_parameter] && configuration[:resolution][:solver_parameter] != -1
-          raise OptimizerWrapper::DiscordantProblemError, 'Deprecated and new solver parameter used at the same time with uncompatible values'
-        end
-      end
-
       # periodic consistency
```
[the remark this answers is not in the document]
braktar: Yes, both parameters are mutually exclusive now

optimizer_wrapper.rb
```diff
@@ -568,7 +568,7 @@ def self.build_csv(solutions)
   def self.check_result_consistency(expected_value, results)
     [results].flatten(1).each{ |result|
       nb_assigned = result[:routes].collect{ |route| route[:activities].select{ |a| a[:service_id] || a[:pickup_shipment_id] || a[:delivery_shipment_id] }.size }.sum
-      nb_unassigned = result[:unassigned].count{ |unassigned| unassigned[:service_id] || unassigned[:shipment_id] }
+      nb_unassigned = result[:unassigned].count{ |unassigned| unassigned[:service_id] || unassigned[:pickup_shipment_id] || unassigned[:delivery_shipment_id] }
```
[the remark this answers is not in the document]
braktar: The output documentation mentions `pickup_shipment_id` and `delivery_shipment_id`. return only `shipment_id` doesn't allow to represent both the pickup and the delivery in the output.

test/wrappers/vroom_test.rb
braktar: Vroom is still called even with cost_late_multiplier, but, as we remove the timewindow.end from the vehicle, the rest is placed in first position of the route

wrappers/ortools.rb
braktar: It could be a manual intervention OR the 'oom-killer'.

Nevertheless, I will correct it accordingly

wrappers/ortools_result_pb.rb
braktar: Costs are already renamed for a long time on Optimizer-ortools. As Costs and CostDetails have the same index and structure in the proto there it generates no conflict

test/wrappers/vroom_test.rb
braktar: Does it seems ok to you to centralize the tests in another PR ? And rework the tests in a more general way ?

api/v01/vrp.rb
```diff
@@ -36,573 +37,114 @@ def self.call(object, _env)
...
-                  else
-                    present({ solutions: [ret], job: { status: :completed }}, with: Grape::Presenters::Presenter)
-                  end
+          post do # rubocop:disable Metrics/BlockLength
+            # Api key is not declared as part of the VRP and must be handled carefully and separatly from other parameters
+            api_key = params[:api_key]
+            checksum = Digest::MD5.hexdigest Marshal.dump(params)
+            d_params = declared(params, include_missing: false)
+            vrp_params = d_params[:points] ? d_params : d_params[:vrp]
+            APIBase.dump_vrp_dir.write([api_key, vrp_params[:name], checksum].compact.join('_'), { vrp: vrp_params }.to_json) if OptimizerWrapper.config[:dump][:vrp]
```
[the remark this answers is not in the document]
braktar: As the PR is under dev for a long time, i probably miss this change

models/base.rb
```diff
@@ -130,7 +130,7 @@ def self.belongs_to(name, options = {})
 
       define_method("#{name}_id=") do |val_id|
         c = class_from_string(options[:class_name])
-        self[name] = c.find(val_id)
+        self[name] = val_id && c.find(val_id)
```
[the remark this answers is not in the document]
braktar: Objects like start_point_id which can have a value set to nil are not found and return a ActiveHash::RecordNotFound: Couldn't find Models::Point without an ID

the val_id avoid those cases

docker/Dockerfile
```diff
@@ -17,6 +16,7 @@ RUN apt-get install -y git libgeos-dev ${libgeos} libicu-dev nano > /dev/null
 
 COPY . /srv/app
 WORKDIR /srv/app
+RUN gem install bundler --version 2.1.4
```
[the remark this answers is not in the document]
braktar: Or more probably, this update of bundler was not possible previously due to very old gems which were not compatible

wrappers/wrapper.rb
senhalil: I agree the matrix is definitely a problem. Still there are instance for which Vroom takes long time as well. For example, for `CVRPLIB` instances the longest time is 4+ minutes (I assume this is the instance with 1000 jobs).

api/v01/entities/vrp_result.rb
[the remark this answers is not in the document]
senhalil: It is not input but it is output. As long as it is the public API input and output are the same importance. We will merge this as soon as possible to dev but the problem is dev->master... if it takes as long as the last time (10+ months) then your client will be long using the `costs` field.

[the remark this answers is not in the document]
senhalil: It is output but still public API. this will be merged soon but dev->master took 10 months last time. So if it takes that long then your client will be long using the "costs"

test/real_cases_test.rb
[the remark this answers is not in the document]
senhalil: This is not or-tools but Vroom.

api/v01/vrp.rb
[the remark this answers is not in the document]
senhalil: What about the posts coming from users?

docker/README.md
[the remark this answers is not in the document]
senhalil: I prefer it is us who update the readme.

wrappers/jsprit.rb
senhalil: Maybe another ticket.

docker/README.md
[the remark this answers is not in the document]
braktar: Let's integrate #106 then ! :+1:

api/v01/vrp.rb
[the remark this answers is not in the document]
braktar: Posts from users follows the two first mentions: `Json comes as symbol, csv as strings` :wink:

test/real_cases_test.rb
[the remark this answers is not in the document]
braktar: The test is a bit odd now. We tolerate lateness on services and we transmit timewindows for them. Neverthess, VROOM didn't not tolerate lateness and timewindows are now stricts. + The rest was previously placed in post-process and is now integrated to request.

So the resolution is longer but it has few in common with the previous state (It takes 1.88s locally, but I've set it to 4.5s for travis-ci)

CHANGELOG.md
```diff
@@ -2,6 +2,20 @@
...
+- Bump grape to v1.5.0 - It speeds up the processing of POST requests for nested problems (up to 20 times faster)
+- Bump grape-swagger to v1.3.0 - The documentation is now correctly generated. It allows SDK generation.
+- The fields `costs` introduced in v0.1.5 is renamed `cost_details` to avoid confusion with field `cost`
+- Bump VROOM to v1.8.0 and integrates the features integrated since v1.3.0
+- Bump OR-Tools v7.8
+- VROOM were previously always called synchronously, it is now reserved to a set of effective `router_mode` (:car, :truck_medium) within in a limit of points (<200).
+
+### Fixed
+
+- `unassigneds` output were in some cases returning the key `shipment_id` instead of `pickup_shipment_id` and `delivery_shipment_id`
```
fonsecadeline, line 16:
```suggestion
- `unassigned` output were in some cases returning the key `shipment_id` instead of `pickup_shipment_id` and `delivery_shipment_id`
```

fonsecadeline (changes requested):

CHANGELOG.md
```diff
@@ -2,6 +2,20 @@
...
 ## [Unreleased]
 
+### Changed
+
+- Bump grape to v1.5.0 - It speeds up the processing of POST requests for nested problems (up to 20 times faster)
+- Bump grape-swagger to v1.3.0 - The documentation is now correctly generated. It allows SDK generation.
+- The fields `costs` introduced in v0.1.5 is renamed `cost_details` to avoid confusion with field `cost`
+- Bump VROOM to v1.8.0 and integrates the features integrated since v1.3.0
+- Bump OR-Tools v7.8
+- VROOM were previously always called synchronously, it is now reserved to a set of effective `router_mode` (:car, :truck_medium) within in a limit of points (<200).
```
braktar, line 12:
```suggestion
- VROOM were previously always called synchronously, it is now reserved to a set of effective `router_mode` (:car, :truck_medium) within a limit of points (<200).
```

wrappers/vroom.rb
```diff
@@ -29,175 +29,329 @@ def initialize(hash = {})
...
+      when 'job'
+        read_job(vrp, vehicle, step)
+      when 'break'
+        read_break(step)
+      when 'pickup'
+        read_shipment(vrp, vehicle, step)
+      when 'delivery'
+        read_shipment(vrp, vehicle, step)
+      else
+        raise
```
senhalil, line 169: We can merge the same items and order the checks so that it is more efficient (it is better if we check start and end last so that we that the condition is not executed for all services for example)

Also maybe an explanation for the raise

```suggestion
      case step['type']
      when 'job'
        read_job(vrp, vehicle, step)
      when 'break'
        read_break(step)
      when 'pickup', 'delivery'
        read_shipment(vrp, vehicle, step)
      when 'start', 'end'
        read_depot(vehicle, step)
      else
        raise 'Unimplemented stop type in wrappers/vroom.rb' 
```

wrappers/vroom.rb
```diff
@@ -29,175 +29,329 @@ def initialize(hash = {})
...
       problem = { vehicles: [], jobs: [], matrix: [] }
-      vehicle = vehicles.first
-      problem[:vehicles] << {
-        id: 0,
-        start_index: vehicle.start_point_id ? points[vehicle.start_point_id].matrix_index : nil,
-        end_index: vehicle.end_point_id ? points[vehicle.end_point_id].matrix_index : nil
-      }.delete_if{ |_k, v| v.nil? }
-      problem[:jobs] = services.collect.with_index{ |service, index|
-        [{
+      vrp_skills = vrp.vehicles.flat_map{ |vehicle| vehicle.skills.first }.uniq + vrp.services.flat_map{ |service| service.sticky_vehicles.map(&:id) }.uniq
```
senhalil, line 276: If we have `sticky_vehicles`, we should have `sticky_vehicle_ids` as well right?

A note "WARNING: only first alternative set of skills is used" might be useful.

Don't we need uniq one more time in case vehicles have their ID's as skills?

fonsecadeline (approved):

wrappers/vroom.rb
```diff
@@ -29,175 +29,329 @@ def initialize(hash = {})
...
+      vrp_units = vrp.units.select{ |unit| vrp.vehicles.map{ |vehicle| vehicle.capacities.find{ |capacity| capacity.unit.id == unit.id }&.limit }&.max&.positive? }
```
senhalil, line 277: Do we have to eliminate the units? Or-tools doesn't mind if there are units that are not used for example.

wrappers/vroom.rb
```diff
@@ -29,175 +29,329 @@ def initialize(hash = {})
...
+          delivery: {
+            id: vrp.services.size + index * 2 + 1,
+            service: shipment.delivery.duration,
+            location_index: points[shipment.delivery.point_id].matrix_index,
+            time_windows: shipment.delivery.timewindows.map{ |timewindow| [timewindow.start || 0, timewindow.end || 2**30] }
+          }.delete_if{ |_k, v| v.nil? || v.is_a?(Array) && v.empty? }
+        }.delete_if{ |_k, v|
+          v.nil? || v.is_a?(Array) && v.empty?
+        }
+      }
```
senhalil, line 341: I think if we extract the logic here into functions `collect_X(vrp)` for X in `vehicles`, `jobs` and `shipments` `run_vroom` would stay more legible. Maybe matrix logic as well if you think it is easy to extract.

```ruby
problem[:vehicles] = collect_vehicles(vrp)
problem[:jobs] = collect_jobs(vrp)
problem[:shipments] = collect_shipments(vrp)
```

api/v01/vrp.rb
```diff
@@ -36,573 +37,114 @@ def self.call(object, _env)
...
-                  else
-                    present({ solutions: [ret], job: { status: :completed }}, with: Grape::Presenters::Presenter)
-                  end
+          post do # rubocop:disable Metrics/BlockLength
+            # Api key is not declared as part of the VRP and must be handled carefully and separatly from other parameters
+            api_key = params[:api_key]
+            checksum = Digest::MD5.hexdigest Marshal.dump(params)
+            d_params = declared(params, include_missing: false)
+            vrp_params = d_params[:points] ? d_params : d_params[:vrp]
+            APIBase.dump_vrp_dir.write([api_key, vrp_params[:name], checksum].compact.join('_'), vrp_params.to_json) if OptimizerWrapper.config[:dump][:vrp]
```
senhalil, line 93: There is still a modification on this line. Current code is as follows.
```suggestion
            APIBase.dump_vrp_dir.write([api_key, vrp_params[:name], checksum].compact.join('_'), d_params.to_json) if OptimizerWrapper.config[:dump][:vrp]
```
 If we want to be able to re-post a vrp launched and dumped via front "the same way" then we need to dump d_params, right ?

senhalil (changes requested):
I believe there is still an unwanted modification in the post function

test/real_cases_test.rb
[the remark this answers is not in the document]
senhalil: I don't think the perf diff on Travis.com is not that much but we are gonna abandon Travis soon anyways. 

So to not forget the limits, maybe you can add a note something like  `# 4500 is due to Travis, 2000 is enough on local.`

wrappers/vroom.rb
braktar: Vroom documenation advise to put the most important units first, order seems to have some importance. So, removing useless units seems to be a good compromise, when Mweb could send 20+ empty/useless capacities

https://github.com/VROOM-Project/vroom/blob/master/docs/API.md#capacity-restrictions

senhalil: Cheers! Effectively this is an important distinction, I wasn't aware of it. Definitely removing useless units makes lots of sense in this case.

wrappers/vroom.rb
braktar: `vehicle['sticky_vehicle_ids']` is defined not `vehicle.sticky_vehicle_ids` 

Are we sure that sticky_vehicle_ids doesn't contains duplicated keys ?

wrappers/vroom.rb
braktar: Does the performances of case depends on the order of the clauses ?

ok to merge the same items :+1:

senhalil: On C++ it does, on Ruby I am not 100% sure it depends on the underlying implementation. It cannot hurt I imagine.

wrappers/vroom.rb
senhalil: I know that's what I meant but I have no idea about your question. In that case, we would have the same vehicle in the `sticky_vehicles` list as well, no?

senhalil: It was just an idea by the way, if you think it is risky, no reason to hold the PR, you can mark this resolved.

senhalil (approved):
It looks great. Finger's crossed... 

@braktar did you launch all the test on local (including non-travis ones)?

PR #109, opened 2021-01-15
commpress/decompress dumped files

senhalil:
To prevent dumped files hitting the size limit.

I verified that everything works as expected by launching some optims on a `mapotempo/dev` server and then I "get" the previously dumped unzipped solutions w/o a problem on a  `feat/zip_dumps` server.

1 - In case needed, `zlib-flate` can be used to inflate a compressed file via terminal.
For a quick look for small enough files:
`zlib-flate -uncompress < /tmp/optimizer-api/dump/file | less` 
and with json syntax
`zlib-flate -uncompress < /tmp/optimizer-api/dump/file | json_pp | less`

To write the content to a file 
`zlib-flate -uncompress < /tmp/optimizer-api/dump/file > content.json`
and with syntax 
`zlib-flate -uncompress < /tmp/optimizer-api/dump/file | json_pp > content.json`

2 - In Ruby
```ruby
require 'zlib'

content = Zlib::Inflate.inflate(File.read('/tmp/optimizer-api/dump/file'))
```

Closes https://gitlab.com/mapotempo/optimizer-api/-/issues/652

### DIFF
--- a/lib/cache_manager.rb
+++ b/lib/cache_manager.rb
@@ -1,4 +1,5 @@
 require 'fileutils'
+require 'zlib'
 
 class CacheError < StandardError; end
 
@@ -14,7 +15,7 @@ class CacheManager
   def read(name, _options = nil)
     filtered_name = name.to_s.parameterize(separator: '')
     if File.exist?(File.join(@cache, filtered_name))
-      File.read(File.join(@cache, filtered_name), mode: 'r')
+      Zlib::Inflate.inflate(File.read(File.join(@cache, filtered_name), mode: 'r'))
     end
   rescue StandardError => e
     raise CacheError, "Got error #{e} attempting to read cache #{name}." if !cache.is_a? ActiveSupport::Cache::NullStore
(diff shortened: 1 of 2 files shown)


fonsecadeline (approved):

braktar (approved):

lib/cache_manager.rb
```diff
@@ -14,7 +15,12 @@ def initialize(cache)
   def read(name, _options = nil)
     filtered_name = name.to_s.parameterize(separator: '')
     if File.exist?(File.join(@cache, filtered_name))
-      File.read(File.join(@cache, filtered_name), mode: 'r')
+      content = File.read(File.join(@cache, filtered_name), mode: 'r')
+      begin
+        Zlib::Inflate.inflate(content)
+      rescue Zlib::DataError # Previously dumped files were not compressed, protect them
```
fab-girard, line 21: is it possible to check instead file type ? 
- txt --> no need inflate
- other --> inflate

senhalil: `Zlib::Inflate.inflate` has an efficient internal check in fact.

It checks the header flags (https://stackoverflow.com/a/30794147/1200528) and returns immediately without doing much if it doesn't see the magic headers. Since we already have the content, I think this is less error prone compared to maintaining a flag or checking file type manually. What do you say Fabien?

PR #111, opened 2021-01-20
Fix: eliminate negatif infeasible quantities

senhalil:
First step of https://gitlab.com/mapotempo/optimizer-api/-/issues/663


braktar (commented):
Could you add this fix to the Changelog ? To notify users they will now receive an unassigned reason in this case ?

braktar (approved):

PR #113, opened 2021-01-22
return explicit error when using front with file not in JSON format

fonsecadeline:
Fix for : https://gitlab.com/mapotempo/optimizer-api/-/issues/668

### DIFF
--- a/public/js/index.js
+++ b/public/js/index.js
@@ -26,6 +26,7 @@ postForm.on('submit', function (e) {
       vrp = JSON.parse(evt.target.result);
     } catch(e) {
       alert("Le fichier fourni n'est pas dans un format JSON valide :\n" + e);
+      return false;
     }
 
     if (!vrp.vrp) {
--- a/public/js/index.js
+++ b/public/js/index.js
@@ -21,7 +21,12 @@ postForm.on('submit', function (e) {
 
   var reader = new FileReader();
   reader.onload = function (evt) {
-    var vrp = JSON.parse(evt.target.result);
+    var vrp;
+    try {
+      vrp = JSON.parse(evt.target.result);
+    } catch(e) {
+      alert("Le fichier fourni n'est pas dans un format JSON valide :\n" + e);
+    }
 
     if (!vrp.vrp) {
       alert("Fichier invalide");


public/js/index.js
```diff
@@ -21,7 +21,12 @@ postForm.on('submit', function (e) {
 
   var reader = new FileReader();
   reader.onload = function (evt) {
-    var vrp = JSON.parse(evt.target.result);
+    var vrp;
+    try {
+      vrp = JSON.parse(evt.target.result);
+    } catch(e) {
+      alert("Le fichier fourni n'est pas au format JSON");
```
senhalil, line 28: Do we call this function for the file submitted by the user or for the solution returned by the API? 

If it is for the files submitted by the user, is there any useful information in `e`, does it tell why parsing failed -- like, extra comma in line 99 for example? If that is the case, it might help the user to debug their JSON file.

I also suggest adding "valide" to the message "Le fichier fourni n'est pas au format JSON valide"  but this might be just my personal pref..

fonsecadeline: We call this function for file submitted by the user.

For my case, error returned was `JSON.parse: unexpected character at line 8002 column 35 of the JSON data`, while problem was one field was blank. I am not sure JSON parse error are very explicit. 
I run some other tests and you are right, in some cases it is explicit. I will integrate JSON parse error to a clear error message.

I agree with 'valide'. I wrote it in my English version and then realize I needed to write in French :)

senhalil: The language of the front is another thing we need to fix at some point :/

fonsecadeline: I aggree : https://gitlab.com/mapotempo/optimizer-api/-/issues/670

I think this is among most prioritary things. At least having an English version only instead of French version only?

public/js/index.js
```diff
@@ -21,7 +21,12 @@ postForm.on('submit', function (e) {
 
   var reader = new FileReader();
   reader.onload = function (evt) {
-    var vrp = JSON.parse(evt.target.result);
+    var vrp;
+    try {
+      vrp = JSON.parse(evt.target.result);
+    } catch(e) {
+      alert("Le fichier fourni n'est pas dans un format JSON valide :\n" + e);
+    }
```
senhalil, line 29: Do you know if we need `return false` here like the if condition couple of lines down below or not? 

I don't know much JS. If it is not needed here because of a generic reason (for example `alert` already cuts the execution and returns false automatically) then we can fix the if condition below. 

If it is needed then we can return here because otherwise vrp becomes an uninitialized variable (again with my zero JS knowledge, I can't see what would happen :D).

Maybe these questions are easier to address for @AntoineCAMILLE 

```suggestion
      alert("Le fichier fourni n'est pas dans un format JSON valide :\n" + e);
      return false;
    }

fonsecadeline (commented):
@fab-girard doit-on rajouter un 'return false' ? (messages précédents)

public/js/index.js
AntoineCAMILLE: I think that it could be useful to add a `return false` after the alert to avoid (maybe?) a second alert message with `"Fichier invalide"` because vrp is not initialized

senhalil (approved):

braktar (approved):

PR #114, opened 2021-01-22
Fix fakeredis count system

alain-andre:
Fix error due to fakeredis and use a real one like other tests. It fixes also some errors like about `status` in returned json (not `job[status]`) and quotas mistake. 

@fab-girard, @braktar we were not agree where to put the `count_increase` for an optimization, this still can be moved. Do we want to increase the count every time a VRP is valid or only when a full result is given ? 

The only errors and failures are now due to known `build_detail` method, rubocop (@senhalil), a random time test and array that should be empty.


config/environments/test.rb
```diff
@@ -46,7 +46,7 @@ module OptimizerWrapper
 
   PARAMS_LIMIT = { points: 150, vehicles: 10 }.freeze
   QUOTAS = [{ daily: 100000, monthly: 1000000, yearly: 10000000 }] # Only taken into account if REDIS_COUNT
-  REDIS_COUNT = Redis.new # Fake redis
+  REDIS_COUNT = ENV['REDIS_COUNT_HOST'] && Redis.new(host: ENV['REDIS_COUNT_HOST']) || Redis.new
```
alain-andre, line 49: Here is the trick. When launching inside docker, the api service does not have a local redis because it uses a redis service.

test/api/v01/vrp_test.rb
```diff
@@ -181,7 +186,6 @@ def test_cannot_get_vrp
 
     get "/0.1/vrp/jobs/#{@job_id}", api_key: 'vroom'
     assert_equal 404, last_response.status, last_response.body
-    assert_equal JSON.parse(last_response.body)['status'], 'Not Found'
```
alain-andre, line 184: Status is not used anymore.

senhalil:
@alain-andre I don't know why you tagged me for rubocop?

Rubocop rules are changed in PR #68 and it is merged recently without fixing the new errors due changed rules. I am not surprised that it fails.

alain-andre:
> @alain-andre I don't know why you tagged me for rubocop?
> 
> Rubocop rules are changed in PR #68 and it is merged recently without fixing the new errors due changed rules. I am not surprised that it fails.

Ok, I did not notice this, I still had in mind that fixing rubocop to v1.81 was enough not to break tests. I'll check.

senhalil:
Don't know about the rubocop version, I don't remember when I touched it last time. However, I know that two days ago when I created PR #111 and pushed commit 3a696e9a2b0d56d5469759da82752732b784068f rubocop test was okay (because I have a git-hook which verifies this) and it still is okay at that moment in time.

```
(system-3.6.9) halil@:~/dev/optimizer-api((HEAD detached at 3a696e9a))$ git checkout 3a696e9a2b0d56d5469759da82752732b784068f
HEAD is now at 3a696e9a Fix: eliminate negatif infeasible quantities
post-checkout hook is called
(system-3.6.9) halil@:~/dev/optimizer-api((HEAD detached at 3a696e9a))$ optim_test_rubocop 
Started with run options --verbose --seed 58700

  /0: [=---=---=---=---=---=---=---=---=---=---=---=---=---=---=---=---=---=---=] 0% Time: 00:00:00,  ETA: ??:??:??

RubocopTest#test_no_offenses_in_tests
  1/1: [========================================================================] 100% Time: 00:00:01, Time: 00:00:01

Finished in 1.05855s
1 tests, 1 assertions, 0 failures, 0 errors, 0 skips
```

Rakefile
```diff
@@ -38,6 +38,7 @@ Rake::TestTask.new do |t|
   $stderr.sync = true
   ENV['APP_ENV'] ||= 'test'
   t.pattern = 'test/**/*_test.rb'
+  t.warning = false
```
fab-girard, line 41: I think it's dangerous to remove warning for future problems, no?

test/api/v01/vrp_test.rb
```diff
@@ -20,7 +20,12 @@
 
 class Api::V01::VrpTest < Api::V01::RequestHelper
   include Rack::Test::Methods
-  include FakeRedis
+
+  def clear_optim_redis_count
```
fab-girard, line 24: it creates a new test called `clear_optim_redis_count` is'n it? Is it what we want if it is the case?

test/api/v01/vrp_test.rb
fab-girard: status has been replaced, you should not remove this line, but replace it

fab-girard (approved):

test/api/v01/vrp_test.rb
```diff
@@ -440,7 +444,7 @@ def test_multitrips
 
   def test_count_optimizations
      [
-      { method: 'post', uri: 'submit', operation: :submit_vrp, options: { vrp: VRP.toy }},
```
braktar, line 443: the nickname of the API operation is :submit_vrp

test/api/v01/vrp_test.rb
```diff
@@ -20,9 +20,14 @@
 
 class Api::V01::VrpTest < Minitest::Test
   include Rack::Test::Methods
-  include FakeRedis
   include TestHelper
 
+  def clear_optim_redis_count
+    OptimizerWrapper.config[:redis_count].keys.select{ |key| key =~ /optimizer:/ }.each do |to_remove|
```
braktar, line 26: Where does the `optimizer` comes from ?

braktar (changes requested):

test/api/v01/vrp_test.rb
fab-girard: I think, even if it is not always natural and we should avoid this, that operation prefix `optimize` is enough explicit to use it. But thread from braktar raises a question for me @alain-andre : why test endpoint without count system ? And in addition you are not testing all endpoints (`: get_job_list` is not here). Maybe you should remove all no counted endpoints in tests...

test/api/v01/vrp_test.rb
fab-girard: `optimizer` is the prefix of all redis count keys (see other opened thread)

test/api/v01/vrp_test.rb
alain-andre: @fab-girard My point of view is that test should cover all possibilities (positives and negatives). I prefer adding `get_job_list` I forgot that remove all other because it makes sure we will never count inside those endpoints. Tell me if I remove uncounted endpoints, or add `get_job_list`.

alain-andre: I just noticed I already test `get_job_list`, it's the third in the array.

alain-andre: ...my bad you were talking in the `when` condition.

senhalil:
If I do `bundle install` in this branch Gemfile.lock changes.. @alain-andre is this the case if you do `bundle install` on your side? If that is the case, can you push that version of the Gemfile.lock?

alain-andre:
> If I do `bundle install` in this branch Gemfile.lock changes.. @alain-andre is this the case if you do `bundle install` on your side? If that is the case, can you push that version of the Gemfile.lock?

Yes you are right, I forgot to remove the gem limitations in lock file, I'm updating the commit.

test/api/v01/vrp_test.rb
alain-andre: Ok, I have changed the test, it assert to get count only during optimize, using the `else` condition to validate other endpoints.

braktar (approved):

fonsecadeline (approved):

test/api/v01/vrp_test.rb
```diff
@@ -439,8 +443,9 @@ def test_multitrips
   end
 
   def test_count_optimizations
+    clear_optim_redis_count
      [
```
senhalil, line 447:
```suggestion
    [
```

test/api/v01/vrp_test.rb
```diff
@@ -439,8 +443,9 @@ def test_multitrips
...
-      { method: 'post', uri: 'submit', operation: :submit_vrp, options: { vrp: VRP.toy }},
+      { method: 'post', uri: 'submit', operation: :optimize, options: { vrp: VRP.toy }},
       { method: 'get', uri: "jobs/#{@job_id}.json", operation: :get_job, options: {}},
       { method: 'get', uri: 'jobs', operation: :get_job_list, options: {}},
       { method: 'delete', uri: "jobs/#{@job_id}.json", operation: :delete_job, options: {}} # delete must be the last one !
```
senhalil, line 451:
```suggestion
      { method: 'delete', uri: "jobs/#{@job_id}.json", operation: :delete_job, options: {}} # delete must be the last one!
```

senhalil (changes requested):
white-space typos otherwise :ok_hand:

PR #115, opened 2021-01-22
dicho rests shouldn't be nil inside corresponding_mission_ids

senhalil:
Closes https://mapotempo.atlassian.net/jira/servicedesk/projects/TICKETS/queues/custom/84/TICKETS-570

This PR basically fixes the following two issues and adds unit tests for each: 

1. In the end phase of dichotomous approach, (due to historical reasons) we include the rest-ids in the initial routes. However, rests are interval variables now and should not be in the initial solution.

2. In the creation of protobuf in `ortools.rb`, we re-collect the IDs and if an ID cannot be found it becomes `nil` which leads to the error `Google::Protobuf::TypeError Invalid argument for string field '' (given NilClass)`.

### DIFF
--- a/wrappers/ortools.rb
+++ b/wrappers/ortools.rb
@@ -418,9 +418,12 @@ module Wrappers
       routes = vrp.routes.collect{ |route|
         next if route.vehicle.nil? || route.mission_ids.empty?
 
+        service_ids = corresponding_mission_ids(services.collect(&:id), route.mission_ids)
+        next if service_ids.empty?
+
         OrtoolsVrp::Route.new(
           vehicle_id: route.vehicle.id,
-          service_ids: corresponding_mission_ids(services.collect(&:id), route.mission_ids)
+          service_ids: service_ids
         )
       }
 
@@ -822,7 +825,7 @@ module Wrappers
 
         available_ids.delete(correct_id)
         correct_id
-      }
+      }.compact
     end
   end
 end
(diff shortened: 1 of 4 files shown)


braktar (approved):

fonsecadeline (approved):

wrappers/ortools.rb
```diff
@@ -418,9 +418,12 @@ def solve(vrp, job, thread_proc = nil, &block)
       routes = vrp.routes.collect{ |route|
         next if route.vehicle.nil? || route.mission_ids.empty?
 
+        service_ids = corresponding_mission_ids(services.collect(&:id), route.mission_ids)
+        next if service_ids.empty?
+
         OrtoolsVrp::Route.new(
           vehicle_id: route.vehicle.id,
-          service_ids: corresponding_mission_ids(services.collect(&:id), route.mission_ids)
```
fonsecadeline, line 423: What is expected behaviour when there is a shipment in route.mission_ids ?
For now, users can not provide _pickup_ neither _delivery_ extension is _mission_ids_ because it would raise `Each mission_ids should refer to an existant service or shipment` DiscordantProblemError (from vrp create function)

senhalil: If a user uses `shipment` type, then they will have to give the shipment id in the route because that is the only id that they have. The ids with delivery/pickup suffixes are dangerous to expose since they are internal and might change any time.

I guess, if they are using the `shipment` type, and they want to give an initial route then they can supply the shipment_id twice and since it is shipment type, then we are sure that the first is pickup and the second is delivery.

PR #117, opened 2021-01-25
Start worker in tests environment only if necessary

fab-girard:


senhalil (changes requested):

braktar (approved):

test/api/v01/request_helper.rb
```diff
@@ -118,7 +120,7 @@ def self.solve_asynchronously
   ensure
     Resque.inline = old_resque_inline
     OptimizerWrapper.config[:solve][:synchronously] = old_config_solve_synchronously
-    if pgid_worker
+    if options[:start_worker]
```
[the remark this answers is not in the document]
fab-girard: we are in test mode, if we cannot spawn worker, do you prefer there is crash to get info for why process has not been spawned or do you prefer to have something like "silent death" (hard to analyse)?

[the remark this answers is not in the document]
senhalil: It is easier with `&& pid_worker` I think. 

Because without it, instead of the test breaking in the expected way, we get a raise from `worker_pids = 'ps -o pgid,pid | grep #{pgid_worker}'.split(/\n/)` which "hides" the broken test. 

With it, if we couldn't launch the worker, this means there is a raise in line `Process.spawn` and if we do not "create" a new raise inside this `if` condition, then the test will correctly point to `Process.spawn` line.

Rakefile
```diff
@@ -29,7 +28,9 @@ namespace :resque do
     }
   end
 
-  task :workers => :clean_working_job_ids # add job cleaning as a pre-requisite
+  # Add job cleaning as a pre-requisite
+  task work: :clean_working_job_ids
```
senhalil, line 32: I thought about doing it like this at first but decided against it, I don't remember why :facepalm: if you can't think of a reason, we can do it like this.

Rakefile
```diff
@@ -29,7 +28,9 @@ namespace :resque do
...
+  task workers: :clean_working_job_ids
```
senhalil, line 33: If we do `clean_working_job_ids` for `:work` do we still need it for `workers`? 

Because `workers` calls `work` internally as far as I know and that should trigger the `clean_working_job_ids` for all workers automatically. I haven't tested this by the way, my memory is hazy, it has been sometime since I worked on this rake task.

senhalil (approved):
I have a question regarding the task on `work` and `workers`otherwise it looks good to me

Rakefile
fab-girard: `workers` seems to have as prerequisite `work` so I am going to remove it as requested at your risk

PR #122, opened 2021-01-28
Remove obsolete tests

fonsecadeline:
I apparently forgot to remove some tests :/ (or probably missed that they were removed when rebasing).


senhalil (approved):

test/wrapper_test.rb
```diff
@@ -3204,4 +3188,72 @@ def test_empty_result_when_no_mission
...
+  end
+
+  def test_assert_inapplicable_vroom_with_periodic_heuristic
+    problem = VRP.scheduling
+    problem[:services].first[:visits_number] = 2
+
+    assert_includes OptimizerWrapper.config[:services][:vroom].inapplicable_solve?(TestHelper.create(problem)), :assert_only_one_visit
+  end
+
+  def test_assert_applicable_for_vroom_if_initial_routes
```
braktar, line 3208: Vroom should not be inapplicable due to initial routes 
https://github.com/Mapotempo/optimizer-api/pull/59

test/wrapper_test.rb
```diff
@@ -3204,4 +3188,72 @@ def test_empty_result_when_no_mission
 
     assert_equal 2, result[:routes].size
   end
+
+  def test_assert_inapplicable_for_vroom_if_vehicle_distance
```
braktar, line 3192: we could actually define the timewindow to the maximum distance as we are sure we have no timewindow currently.

But for now, this test is usefull ;)

test/wrapper_test.rb
```diff
@@ -3204,4 +3188,72 @@ def test_empty_result_when_no_mission
...
+    problem[:relations] = [{
+      type: 'vehicle_group_duration',
+      linked_ids: [],
+      linked_vehicle_ids: [],
+      lapse: nil
+    }]
+
+    vrp = TestHelper.create(problem)
+    refute_includes OptimizerWrapper.config[:services][:vroom].inapplicable_solve?(vrp), :assert_no_relations
+    refute_includes OptimizerWrapper.config[:services][:ortools].inapplicable_solve?(vrp), :assert_no_relations
```
braktar, line 3228: the test name only mention vroom, but we are testing ortools too

braktar (changes requested):

test/wrapper_test.rb
senhalil: test name says "applicable" :confused: but yeah we can use vroom if everything else is "right" even if there is initial solutuion

test/wrapper_test.rb
senhalil: Does that mean that if there is no time dimension we can use TW's to implement max distance?

senhalil (approved):

test/wrapper_test.rb
braktar: Yes !

test/wrapper_test.rb
braktar: Mea culpa !

braktar (approved):

senhalil (approved):

PR #123, opened 2021-01-29
Multi tour

fonsecadeline:
Prerequisites : https://github.com/Mapotempo/optimizer-api/pull/122 (to validate current tests on mapotempo/dev)

Closes : https://gitlab.com/mapotempo/optimizer-api/-/issues/540


fonsecadeline:
We need https://github.com/Mapotempo/optimizer-ortools/pull/20 to be merged at the same time.

models/vrp.rb
```diff
@@ -138,93 +138,8 @@ def self.create(hash, delete = true)
...
-
-      raise OptimizerWrapper::DiscordantProblemError, 'Shipments are not available with periodic heuristic.' unless hash[:shipments].to_a.empty?
-
-      raise OptimizerWrapper::DiscordantProblemError, 'Rests are not available with periodic heuristic.' unless hash[:vehicles].all?{ |vehicle| vehicle[:rests].to_a.empty? }
-
-      if hash[:configuration][:resolution][:same_point_day]
-        raise OptimizerWrapper.UnsupportedProblemError, 'Same_point_day is not supported if a set has one service with several activities' if hash[:services].any?{ |service| service[:activities].to_a.size.positive? }
-      end
+      validator = HashValidator.new(hash)
+      validator.check_hash
```
braktar, line 142: Did you only move the validations or add some in the same time ?

test/wrappers/ortools_test.rb
```diff
@@ -2527,6 +2527,7 @@ def test_route_force_start
         start_point_id: 'point_0',
         end_point_id: 'point_0',
         matrix_id: 'matrix_0',
+        timewindow: { start: 0 },
```
braktar, line 2530: It should not be necessary anymore since we are using grape 1.5.1

wrappers/ortools.rb
```diff
@@ -407,6 +407,8 @@ def solve(vrp, job, thread_proc = nil, &block)
         }.uniq
         next if current_linked_ids.empty? && current_linked_vehicles.empty?
 
+        relation.lapse ||= 0 if relation.type == 'vehicle_trips'
```
braktar, line 410: Protobuf already set nil values to zero for numerical fields

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@
 ### Added
 
 - Corresponding vehicle_id is returned within each service's skills if problem is partitioned with vehicle entity [#110](https://github.com/Mapotempo/optimizer-api/pull/110)
+- Implementation of vehicle_trips relation : these vehicles' routes are successive [#123] (https://github.com/Mapotempo/optimizer-api/pull/123)
```
braktar, line 8:
```suggestion
- Implementation of vehicle_trips relation : these vehicles' routes are successive [#123] (https://github.com/Mapotempo/optimizer-api/pull/123)
```
```suggestion
- Implementation of `vehicle_trips` relation : these vehicles' routes are successive [#123] (https://github.com/Mapotempo/optimizer-api/pull/123)
```

braktar (changes requested):

models/vrp.rb
fonsecadeline: I did both, but in separated commits

wrappers/ortools.rb
fonsecadeline: In this case optimizer-ortools received -1. I will have to check what happens then.

fonsecadeline: I think our code (ortools.rb) sends -1 by default.

test/wrappers/ortools_test.rb
fonsecadeline: I added it because I add a checker to say there can not be force_start option activated if vehicle has no timewindow start. Does it make sense for you ?

wrappers/ortools.rb
braktar: Could you think of a relation where having -1 or 0 would have different behavior ?

fonsecadeline: I checked and no. The only difference would be when we check if lapse == -1 from tsp_simple side. I replaced it by checking lapse == 0 and in fact it even makes more sense because we don't want to generate a vehicle group duration relation where duration is 0.

https://github.com/Mapotempo/optimizer-ortools/pull/20

senhalil: Unless someone would like to "disable" a group of vehicles easily by setting their group duration 0. That would be the modification with least change and a lot easier then removing the vehicles or setting all their duration to 0. 

I wonder if that was the original reasoning of `-1`

CHANGELOG.md
```diff
@@ -5,6 +5,8 @@
 ### Added
 
 - Corresponding vehicle_id is returned within each service's skills if problem is partitioned with vehicle entity [#110](https://github.com/Mapotempo/optimizer-api/pull/110)
+- Implementation of `vehicle_trips` relation : these vehicles' routes are successive [#123] (https://github.com/Mapotempo/optimizer-api/pull/123)
+- Implementation of lapses within vehicle_trips relations : minimum duration between two consecutive trips [#123] (https://github.com/Mapotempo/optimizer-api/pull/123)
```
senhalil, line 9:
```suggestion
- Implementation of `vehicle_trips` relation: the routes can be successive or with a minimum duration `lapse` in between [#123] (https://github.com/Mapotempo/optimizer-api/pull/123)
```

api/v01/entities/vrp_input.rb
```diff
@@ -239,12 +239,16 @@ module VrpMisc
...
                                                                  vehicle_group_duration_on_months vehicle_group_number],
                     desc: 'Relations allow to define constraints explicitly between activities and/or vehicles.
                            It could be the following types: same_route, sequence, order, minimum_day_lapse, maximum_day_lapse,
-                           shipment, meetup, minimum_duration_lapse, maximum_duration_lapse')
-    optional(:lapse, type: Integer, desc: 'Only used for relations implying a duration constraint : minimum/maximum day lapse, vehicle group durations...')
+                           shipment, meetup, minimum_duration_lapse, maximum_duration_lapse, vehicle_trips')
+    optional(:lapse, type: Integer, desc:
+      'Only used for relations implying a duration constraint.
+      Lapse expressed in days for minimum/maximum day lapse,
+      in seconds for vehicle_group_durations or vehicle_trips.')
```
senhalil, line 251: Do we want this style? How does it look in the auto generated documentation? 

I think it is better if we leave the description parsing to grape-swagger and disable line width cop for this file because thanks to all descriptions being a single line, all parameter definitions come one after another and it is very easy to follow/skim through the parameters; otherwise, it will look scattered I think. If this is what we want we should do it for every definition; if not, it is like reading a book with different font for every word -- i.e., hard on the eyes :eyes: 

What do you think @fonsecadeline, @braktar?

lib/interpreters/multi_trips.rb
```diff
@@ -21,14 +21,14 @@
...
+      vrp.vehicles = vrp.vehicles.flat_map{ |vehicle|
         if vehicle.trips > 1
           new_ids = Array.new(vehicle.trips) { |index| vehicle.id + '_trip_' + index.to_s }
-          new_vehicles = new_ids.collect{ |id|
+          new_vehicles = new_ids.collect.with_index{ |id, index|
             new_vehicle = Marshal.load(Marshal.dump(vehicle))
-            new_vehicle.original_id = new_vehicle.id
             new_vehicle.id = id
             new_vehicle.trips = 1
+            new_vehicle.force_start = false unless index.zero?
```
senhalil, line 31: Do we want this? What if I have 5 trips but I want them all to be as early as possible, can't I do that with setting `force_start = true`? I don't know the answer by the way, just brain storming

test/models/vehicle_test.rb
```diff
@@ -21,6 +21,34 @@ module Models
   class VehicleTest < Minitest::Test
     include Rack::Test::Methods
```
senhalil, line 22: not related to this PR but I just saw this for the first time `Rack::Test::Methods`, what do we do with it?

test/models/vrp_consistency_test.rb
```diff
@@ -41,14 +41,14 @@ def test_reject_if_periodic_with_any_relation
...
-       'vehicle_group_duration', 'vehicle_group_duration_on_weeks', 'vehicle_group_duration_on_months'].each{ |relation_type|
+       'vehicle_group_duration', 'vehicle_group_duration_on_weeks', 'vehicle_group_duration_on_months', 'vehicle_trips'].each{ |relation_type|
         vrp[:relations] = [{
             type: relation_type,
             linked_ids: ['service_1', 'service_2']
         }]
 
         assert_raises OptimizerWrapper::DiscordantProblemError do
-            TestHelper.create(vrp)
+          TestHelper.create(vrp)
```
senhalil, line 51: Do we need to call the `create(vrp)` in these tests? Can't we call `check_consistency` (or the other sub-functions of vrp.create) so that we don't actually call `active_hash` and create the VRP object? Because `DiscordantProblemError` would be only raise by these sub-functions no? 

It looks like there are other tests like this -- i.e., they call create just for the raise but then don't use the generated VRP object.

senhalil: I don't repeat the same comment for the other tests down below but  I think they can all just call the apprriate "validation" function.

test/wrappers/ortools_test.rb
senhalil: start is not 0 by default? I can't imagine another default value for start. If we had the proper default value for it, then that checker and this modification would be unnecessary, no?

senhalil (changes requested):

test/wrappers/ortools_test.rb
fonsecadeline: We can use 0 as start default value. I guess that would make some parts of the code cleaner.
@braktar do you see any case where that would be uncompatible with current logic ? I do not see how we could be using negative start values anyway.

wrappers/ortools.rb
fonsecadeline: I am not sure because I think `-1` value is much older than `VehicleGroupDuration`. Also, those `VehicleGroupDuration` relations are not documented anymore. We are not sure about their performance. I am not even sure that providing group duration of 0 would allow to assign services on other vehicles...

Question is : do we want to allow customers to disable vehicles through this relation ? I would suggest to keep 0 as default relation.lapse value and add new relation (if needed one day) to disable a group of vehicles.

lib/interpreters/multi_trips.rb
fonsecadeline: You can do it but then you would have to provide 5 vehicles with independent timewindows and link those vehicles with `vehicle_trips` relation. 

If you provide `trips` field to one vehicle then every trip will have same timewindow and try to start at the same time if `force_start` is activated for all of them.

test/models/vehicle_test.rb
fonsecadeline: I don't know :/

api/v01/entities/vrp_input.rb
senhalil: @fonsecadeline raised a valid issue "the `coerce_with` logic comes after the `description` which hides them off screen", I think we can make sure that the description is always last. What do you guys think, other suggestions?

lib/interpreters/multi_trips.rb
senhalil: Yes the timing will be the same but this doesn't prevent us, no? The new force_start implementation should be flexible enough to accommodate multiple vehicles with the same time window start and force_start=true with order constraints (trips) between them. Did we try?

In my opinion, the `trips` shortcut functionality is there to help the user to copy/paste their vehicles easily and in this copy/paste logic if do such modifications, then a user who really have 5 identical vehicles for trips need to duplicate their vehicles manually instead of being able to use `trips`.

test/models/vehicle_test.rb
senhalil: @braktar ?

wrappers/ortools.rb
senhalil: This is not about allowing them or not using a valid option to do something valid. It is about expectations of the behavior of parameters.

If someone sets a duration of a vehicle (for example) to 1 second, it means 1 second of work. If they set it to 0, then it should mean 0 seconds of wotk. I don't think we should remove this possibility for no good reason. `-1` is clearly an "impossible" setting that we can use internally for the "turn off" logic. If we use 0 for this "turn-off" logic, we are using a valid setting which does the opposite of what it would normally do.

[principle of least surprise](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) leads me to think that if a user sets the duration limit to `x`, then they would expect the vehicle to work no more than `x`. And it would be unexpected if x=0 would do the opposite. If the user didn't want a duration limit in the first place, they could just not supply the `duration` parameter or do `duration:  nil`

senhalil (changes requested):

api/v01/entities/vrp_input.rb
senhalil: I apparently started my review before you sent you comment Adeline :D

senhalil: I don't think it is that hard to follow `description comes always last`..

Maybe we should compare it visually? I tested and my vote is on keeping the description always at the last position and dis-activating the line with cop on this file.

test/wrappers/ortools_test.rb
braktar: Ok for me to set 0 as default value for the start of a timewindow

api/v01/entities/vrp_input.rb
senhalil: so it turns out the generated docs are not correct :facepalm: so we decided to get the best of two worlds... 

description is always at last, always on a separate line of its own but even if it violates the line with cop we don't cut it.

test/models/vehicle_test.rb
braktar: Probably a copy/paste from another test file

lib/interpreters/multi_trips.rb
fonsecadeline: Do you mean new implementation of force start will be able to adjust one vehicle start according to previous trip end? If this is the case then I agree, we can use force_start on every vehicle!

senhalil: No it doesn't but it doesn't need to.

All `force_start` does (this is from memory, you should check it) is to `minimize (actual_vehicle_start - vehicle_timwindow_start)`. So  second vehicle having the same time window start as the first vehicle doesn't change anything, they both get minimized by the solver and both vehicles leave the depot as early as possible (again need to test it with a realistic instance but I see no reason for it not to work). 

Actually thanks to this we might have less gap in between vehicles but not sure :thinking:

test/models/vehicle_test.rb
senhalil: :facepalm: ffs

wrappers/ortools.rb
fonsecadeline: I agree. In fact main problem is : why does ortools consider group_durations with no lapse (which used to correspond to lapse = -1). In preprocessing I will just filter relations that need a lapse and have no lapse (that would mean lapse is infinite, so relation is not needed). Like this if tsp_simple gets lapse = 0 it will do what we expect.

fonsecadeline (commented):
See https://github.com/Mapotempo/optimizer-api/pull/145#discussion_r598480921

api/v01/entities/vrp_input.rb
```diff
@@ -479,7 +480,6 @@ module VrpVehicles
     optional(:force_start, type: Boolean, documentation: { hidden: true }, desc: '[ DEPRECATED ]')
     optional(:shift_preference, type: String, values: ['force_start', 'force_end', 'minimize_span'], desc: 'Force the vehicle to start as soon as the vehicle timewindow is open,
       as late as possible or let vehicle start at any time. Not available with periodic heuristic, it will always leave as soon as possible.')
-    optional(:trips, type: Integer, default: 1, desc: 'The number of times a vehicle is allowed to return to the depot within its route. Not available with periodic heuristic.')
```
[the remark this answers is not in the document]
senhalil: We can remove it directly since it wasn't even in the released docs (that's the case right?) so it didn't exist from the point of view of the end user. If it was released, I still want to say, we can just get rid of it if there is no one using it.

lib/validators.rb
```diff
@@ -137,9 +137,107 @@ def check_shipments_specificities
...
+
+      [vehicles_days, vehicles_unavailable_indices]
+    end
+
+    def check_vehicle_trips_stores_consistency(relation_vehicles)
+      relation_vehicles.each_with_index{ |vehicle_trip, v_i|
+        case v_i
+        when 0
+          unless vehicle_trip[:end_point_id]
+            raise OptimizerWrapper::DiscordantProblemError.new('First trip should at least have end point id')
```
senhalil, line 162: First trip should at least have **an** end point id

Cannot do suggestions :thinking:  I guess it is beacuse I am reviewing commit by commit :(

lib/validators.rb
```diff
@@ -137,9 +137,107 @@ def check_shipments_specificities
...
+          end
+        when relation_vehicles.size - 1
+          unless vehicle_trip[:start_point_id]
+            raise OptimizerWrapper::DiscordantProblemError.new('Last trip should at least have start point id')
```
senhalil, line 166: **a**  starting point id

lib/validators.rb
```diff
@@ -137,9 +137,107 @@ def check_shipments_specificities
...
+          end
+        else
+          unless vehicle_trip[:start_point_id] && vehicle_trip[:end_point_id]
+            raise OptimizerWrapper::DiscordantProblemError.new('Trip should have start and end point id')
```
senhalil, line 170: **Intermediary**(?) trips   

 start and  end point id**s**

lib/validators.rb
```diff
@@ -137,9 +137,107 @@ def check_shipments_specificities
...
+          end
+        end
+
+        next if v_i.zero? ||
+                vehicle_trip[:start_point_id] == relation_vehicles[v_i - 1][:end_point_id] ||
+                (@hash[:points].find{ |pt| pt[:id] == vehicle_trip[:start_point_id] }[:location] ==
+                  @hash[:points].find{ |pt| pt[:id] == vehicle_trip[:end_point_id] }[:location])
```
senhalil, line 177: this looks like a bug :thinking:  shouldn't it be `relation_vehicles[v_i - 1][:end_point_id]` instead of `vehicle_trip[:end_point_id]` ?

lib/validators.rb
```diff
@@ -137,9 +137,107 @@ def check_shipments_specificities
...
+
+        raise OptimizerWrapper::DiscordantProblemError.new('One trip should start where previous trip ended')
```
senhalil, line 179: **the** previous

lib/validators.rb
```diff
@@ -137,9 +137,107 @@ def check_shipments_specificities
...
+      }
+    end
+
+    def check_vehicle_trips_timewindows_consistency(relation_vehicles)
+      vehicles_timewindows =
```
senhalil, line 184: General comment: I know why you want to do this but the double plural makes it hard to read it in English. 

At the end of the day, this is an array of `vehicle timewindow` objects so the 's' would come to the end -- like `vehicle timewindow`s -- or to be more explicit `all_vehicle_timewindows` or (we don't have the timewindows of services but still) just `all_timewindows` or `timewindows`

lib/validators.rb
```diff
@@ -137,9 +137,107 @@ def check_shipments_specificities
       }
     end
 
+    def vehicle_days_availability(vehicles_set, timewindows_set)
```
senhalil, line 140: I think we can be more explicit here, for example `calculate_day_availabilities(vehicles, timewindows)` is more explanatory and gives the raison d'etre

We don't need the `set`s beucase they are not sets but also just `vehicles` and `timewindows` are enough.

lib/validators.rb
```diff
@@ -137,9 +137,107 @@ def check_shipments_specificities
...
+      vehicles_timewindows =
+        relation_vehicles.collect{ |v| v[:timewindow] ? [v[:timewindow]] : v[:sequence_timewindows].to_a }
+
+      return if vehicles_timewindows.all?(&:empty?)
+
+      week_days, unavailable_indices = vehicle_days_availability(relation_vehicles, vehicles_timewindows)
+      if week_days.uniq.size > 1 ||
+         unavailable_indices.uniq.size > 1
+        raise OptimizerWrapper::DiscordantProblemError.new(
+          'Vehicles in vehicle_trips relation have different available days'
```
senhalil, line 193: **cannot** have diff...

or

**should** have **the same** available days

lib/validators.rb
```diff
@@ -137,9 +137,107 @@ def check_shipments_specificities
...
+      vehicles_timewindows.each_with_index{ |v_tw, v_i|
+        next if v_i.zero?
+
+        v_tw.each{ |tw|
+          day = tw[:day_index]
+          previous_tw =
+            vehicles_timewindows[v_i - 1].select{ |ptw|
+              ptw[:day_index].nil? ||
+                ptw[:day_index] == day
+            }.min_by{ |ptw| ptw[:start] }
```
senhalil, line 206: Just to verify: this ``select.min_by` can't return `nil` because previous_vehicle should have the same timewindows with the "current" vehicle (`v_tw`), is that it? otherwise, if the select returns empty, min_by returns `nil`.

lib/validators.rb
```diff
@@ -137,9 +137,107 @@ def check_shipments_specificities
...
+      @hash[:relations].group_by{ |relation| relation[:type] }.each{ |type, relations|
+        case type
+        when 'vehicle_trips'
+          relations.each{ |relation|
+            relation_vehicles =
+              relation[:linked_vehicle_ids].to_a.collect{ |v_id| @hash[:vehicles].find{ |v| v[:id] == v_id } }
+
+            if relation_vehicles.empty?
+              raise OptimizerWrapper::DiscordantProblemError.new(
+                'Non empty list of vehicles should be provided for every vehicle_trips relations'
```
senhalil, line 227: **A** non empty list of vehicle~~s~~ **ids** should be provided for ~~every~~ vehicle_trips relations

test/models/vrp_consistency_test.rb
```diff
@@ -368,5 +368,95 @@ def test_ensure_no_skill_matches_with_internal_skills_format
       end
       assert_equal "There are vehicles or services with 'vehicle_partition_*', 'work_day_partition_*' skills. These skill patterns are reserved for internal use and they would lead to unexpected behaviour.", error.message
     end
+
+    def test_vehicle_trips_relation_validity
```
senhalil, line 372: test_vehicle_trips_relation_must_have_linked_vehicle_ids

senhalil (commented):
Expose vehicle_trips relation commit

lib/validators.rb
```diff
@@ -280,9 +280,17 @@ def check_configuration(configuration, periodic_heuristic)
     end
 
     def check_clustering_parameters(configuration)
+      if @hash[:relations].to_a.any?{ |relation| relation[:type] == 'vehicle_trips' }
+        if configuration[:preprocessing][:partitions].to_a.size.positive?
+          raise OptimizerWrapper::DiscordantProblemError.new(
+            'Clustering is not able to deal with vehicle_trips relation for now'
+          )
```
senhalil, line 287: [:partitions]&.any? instead of [:partitions].to_a.size.positive? 

'Partitioning (or :partitions parameter) is not currently available with vehicle_trips relation'

senhalil (commented):
Partitions are not allowed with vehicle_trips commit

I think this commit can be merged with the precious one.

api/v01/entities/vrp_input.rb
```diff
@@ -259,7 +259,7 @@ module VrpMisc
                            shipment, meetup, minimum_duration_lapse, maximum_duration_lapse, vehicle_trips')
     optional(:lapse,
              type: Integer, values: ->(v) { v.positive? },
-             desc: 'Only used for relations implying a duration constraint : minimum/maximum day lapse, vehicle group durations...')
+             desc: 'Only used for relations implying a duration constraint. Lapse expressed in days for minimum/maximum day lapse, in seconds for vehicle_group_durations or vehicle_trips.')
```
senhalil, line 262: vehicle_group_durations ~~or~~**and** vehicle_trips (Is there any other relations with lapse in seconds?)

senhalil (commented):
[OA] multi-trips with lapses commit

I think this commit should also be merged with the main one

lib/validators.rb
fonsecadeline: Haha no problem. Yes it is, because it assumes that this line can change from this commit to the end of PR.

senhalil (commented):
Remove trips field in vehicle model, add original info and transform :heavy_check_mark:

lib/validators.rb
fonsecadeline: Right ! Thank you !

fonsecadeline: Do you think _find_ is costly enough to keep this line ? 
```
                  vehicle_trip[:start_point_id] == relation_vehicles[v_i - 1][:end_point_id] ||
```
because in fact this does exactly the same.

senhalil: I thought the exact same thing :D haa 

I think your way is good since it does the least costly check first and the other one only if it is necessary.

Since it is a find in all points, it might be depending on the size of the problem (and since this is all inside the POST process, anything faster (as long as it is correct) is better)

lib/validators.rb
senhalil: I think you can resolve typo/grammar "type" conversations while you are making the corrections if you don't mind. There is no point for me to go over them again.

lib/validators.rb
fonsecadeline: Here, min_by can return _nil_. So you are right, my next condition needs to be fixed : we need to ensure there is one previous_tw.

api/v01/entities/vrp_input.rb
fonsecadeline: No other relations with lapse in seconds

fonsecadeline:
This PR needs :
- [ ] merge of https://github.com/Mapotempo/optimizer-api/pull/181

models/concerns/validate_data.rb
```diff
@@ -140,9 +140,108 @@ def check_shipments_specificities
...
+      raise OptimizerWrapper::DiscordantProblemError.new(
+        'Vehicles in vehicle_trips relation should have the same available days'
+      )
+    end
+
+    vehicles_timewindows.each_with_index{ |v_tw, v_i|
+      next if v_i.zero?
+
+      v_tw.each{ |tw|
+        day = tw[:day_index]
```
braktar, line 205: what if a vehicle has a timewindow {start: 0:00, end:8:00, day_index: 1} and the previous one in the trip has {start: 16:00, end: 23:45, day_index:0} with a lapse of one hour ?

models/concerns/validate_data.rb
```diff
@@ -140,9 +140,108 @@ def check_shipments_specificities
...
+        relations.each{ |relation|
+          relation_vehicles =
+            relation[:linked_vehicle_ids].to_a.collect{ |v_id| @hash[:vehicles].find{ |v| v[:id] == v_id } }
+
+          if relation_vehicles.empty?
+            raise OptimizerWrapper::DiscordantProblemError.new(
+              'A non empty list of vehicles IDs should be provided for vehicle_trips relations'
+            )
+          elsif relation_vehicles.any?(&:nil?)
+            raise OptimizerWrapper::DiscordantProblemError.new(
```
braktar, line 234: Could you add this comment ? `FIXME: The linked_vehicle_ids should be directly related to vehicle objects by the model`

braktar (changes requested):

models/concerns/validate_data.rb
fonsecadeline: For now, vehicles trip should take place at the exact same day, so this case is not supported yet.

braktar: The above check verify that all vehicles have the same days, not only one and exact same day ?

fonsecadeline:
- [ ] merge https://github.com/Mapotempo/optimizer-ortools/pull/20
- [ ] add tag on optimizer-ortools : 1.5.0
- [ ] edit README & Dockerfile before merging this PR

lib/interpreters/periodic_visits.rb
```diff
@@ -456,67 +450,69 @@ def compute_possible_days(vrp)
...
+        )
+      end
+
+      additional_relations
+    end
+
+    def collect_weeks_in_schedule
+      all_days = (@schedule_start..@schedule_end).to_a
+      current_index = 0
+      next_monday = all_days.find_index{ |d| d % 7 == 0 && d > current_index }
```
braktar, line 483: Its seems to me that the find loop is unnecessary.

We only have to care about the first and the last week. The weeks between are all complete

braktar (changes requested):

test/test_helper.rb
```diff
@@ -106,9 +111,17 @@ def self.coerce(vrp)
...
     vrp.preprocessing_partitions&.each{ |partition| partition[:entity] = partition[:entity].to_sym } if vrp.is_a?(Models::Vrp)
 
-    vrp.provide_original_info unless vrp.is_a?(Hash) # TODO: re-dump with this modification
+    vrp[:relations]&.each{ |r| r[:type] = r[:type]&.to_sym }
 
-    vrp[:relations]&.each{ |r| r[:type] = r[:type]&.to_sym } # TODO: re-dump with this modification
+    vrp[:vehicles]&.each{ |v|
+      next if v[:skills].to_a.empty?
+
+      if v[:skills].first&.is_a?(Array)
```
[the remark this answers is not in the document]
fonsecadeline: Yes, that is what I am trying to do. However I do not understand what is your suggestion here. 

> is_a?(Array[Array[String]]) 

is not possible. We are not calling grape here, that is why we call test_helper::coerce function so transformation into nil does not exist in this case.

test/test_helper.rb
```diff
@@ -68,16 +68,21 @@ def self.coerce(vrp)
...
 
         [s[:activity] || s[:activities] || s[:pickup] || s[:delivery]].flatten.each{ |activity|
           next unless activity[:position]
 
           activity[:position] = activity[:position].to_sym
         }
+
+        next if vrp.is_a?(Hash) && !s.key?(:visits_number)
+
+        unless s[:visits_number].is_a?(Integer) && s[:visits_number].positive?
```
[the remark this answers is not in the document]
fonsecadeline: Grape is not called in tests, that is why we need to do this in coerce.

lib/interpreters/periodic_visits.rb
fonsecadeline: yes, maybe I can simplify this function

test/test_helper.rb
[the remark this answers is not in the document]
braktar: The else case should return an error then. if skills is a string, then the vrp is incorrect.

test/test_helper.rb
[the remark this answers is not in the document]
braktar: I was convicted that  it was called right after grape. Mea culpa then ;)

test/wrappers/ortools_multi_trips_test.rb
```diff
@@ -0,0 +1,120 @@
+# Copyright © Mapotempo, 2019
```
braktar, line 1:
```suggestion
# Copyright © Mapotempo, 2021
```

fonsecadeline: oh, we are in 2021 already ? xD thanks !

braktar (approved):

senhalil (approved):

PR #125, opened 2021-02-03
remove deprecated assert

fonsecadeline:


wrappers/vroom.rb
```diff
@@ -42,6 +42,7 @@ def solver_constraints
         :assert_no_subtours,
         :assert_points_same_definition,
         :assert_single_dimension,
+        :assert_no_planning_heuristic,
```
braktar, line 45: This assert is relative to the resolution method, it should belong to the Solver catgory

braktar (changes requested):

wrappers/vroom.rb
```diff
@@ -65,6 +65,7 @@ def solver_constraints
         :assert_no_free_approach_or_return,
         :assert_small_minimum_duration,
         :assert_solver,
+        :assert_no_planning_heuristic,
```
braktar, line 68: I would prefer to respect the alphabetic order :)

braktar (changes requested):

braktar (approved):

senhalil (approved):

PR #126, opened 2021-02-04
Define several sub test tasks

fab-girard:


test/test_helper.rb
```diff
@@ -29,7 +28,7 @@
 
 require 'minitest/reporters'
 Minitest::Reporters.use! [
-  Minitest::Reporters::ProgressReporter.new,
+  !ENV['TIME'] ? Minitest::Reporters::ProgressReporter.new : nil,
```
fonsecadeline, line 31:
```suggestion
  ENV['TIME'] ? Minitest::Reporters::ProgressReporter.new : nil,
```
I am not sure but it seems to me that the two cases have been reversed

fab-girard: nop, it is not correct

fonsecadeline (approved):

Rakefile
```diff
@@ -33,13 +33,58 @@ namespace :resque do
...
+    ENV['TEST'] ||= 'test/**/*clustering*_test.rb'
+    Rake::Task['test'].invoke
+  end
+
+  task :periodic do
+    ENV['COV'] = 'false'
+    ENV['TEST'] ||= 'test/**/*scheduling*_test.rb'
+    Rake::Task['test'].invoke
+  end
+end
```
fonsecadeline, line 86: we might need a task for wrappers (and dichotomious) tests

test/test_helper.rb
braktar: Which one is not correct ?

Rakefile
braktar: It would be nice to have a equivalent for each scenario described in ci-utils/tests.sh

test/test_helper.rb
```diff
@@ -16,8 +16,11 @@
 # <http://www.gnu.org/licenses/agpl.html>
 #
 require 'minitest'
-require 'simplecov'
-SimpleCov.start if (!ENV.has_key?('COV') && !ENV.has_key?('COVERAGE')) || (ENV['COV'] != 'false' && ENV['COVERAGE'] != 'false')
+
+if Rake.application.top_level_tasks.include?('test') && (!ENV.key?('COV') && !ENV.key?('COVERAGE')) || (ENV['COV'] != 'false' && ENV['COVERAGE'] != 'false')
```
braktar, line 20: The formulation of the condition seems odd, I suggest the following
```suggestion
if Rake.application.top_level_tasks.include?('test') && !(ENV.key?('COV') || ENV.key?('COVERAGE')) || (ENV['COV'] == 'true' && ENV['COVERAGE'] == 'true')
```

test/test_helper.rb
fab-girard: I let you test the Adeline's suggestion, but it's not the targeted behavior: `ProgressReporter` and `SpecReporter` are exclusive

test/test_helper.rb
fab-girard: we do not support `COV=true` or `COVERAGE=true`: see our README

Rakefile
fab-girard: yes I agree, I've created some tasks to explain you all the abilities, you are free to create any task you would need in the future (maybe in a new PR after this one)

test/test_helper.rb
senhalil: I guess what Gwen is suggesting, the condition don't read/sound natural. Otherwise, we don't have `COV=false` or `COVERAGE=false` in the README either.

Even after Gwen's modification I think the condition is still too complicated. 

Why don't we do as follows (if our intention is to keep the coverage on by default)

```
if Rake.application.top_level_tasks.include?('test') && [ENV['COVERAGE'], ENV['COV']].exclude? 'false'
```

or if our intention is to let it be disabled by default then we can do 

```
if Rake.application.top_level_tasks.include?('test') && [ENV['COVERAGE'], ENV['COV']].include? 'true'
```

senhalil: What do you think @fab-girard ?

test/test_helper.rb
```diff
@@ -29,7 +32,7 @@
 
 require 'minitest/reporters'
 Minitest::Reporters.use! [
-  Minitest::Reporters::ProgressReporter.new,
+  !ENV['TIME'] ? Minitest::Reporters::ProgressReporter.new : nil,
   ENV['HTML'] && Minitest::Reporters::HtmlReporter.new, # Create an HTML report with more information
   ENV['TIME'] && Minitest::Reporters::SpecReporter.new, # Generate a report to find slowest tests
```
braktar, line 37: If they are mutually exclusive, it could be better to show it explicitly 
```suggestion
  ENV['TIME'] ? Minitest::Reporters::SpecReporter.new : Minitest::Reporters::ProgressReporter.new , # Generate a report to find slowest tests
```

braktar (changes requested):

README.md
```diff
@@ -89,6 +89,26 @@ TIME=true HTML=true APP_ENV=test bundle exec rake test
...
+APP_ENV=test bundle exec rake test TESTOPTS="--name=test_instance_clustered"
+```
+Tests are splitted into several sets and you can replay only one or more set(s):
+```
+APP_ENV=test bundle exec rake test:api
+APP_ENV=test bundle exec rake test:api test:models
+```
+You can list all available tasks with:
+```
+bundle exec rake -T
```
senhalil, line 108: `rake -T` only shows the tasks which has a description so currently this doesn't show the sub test tasks because they are defined in an unexpected format. So currently it shows 

```
rake resque:failures:sort  # Sort the 'failed' queue for the redis_multi_queue failure backend
rake resque:work           # Start a Resque worker
rake resque:workers        # Start multiple Resque workers
rake test                  # Run tests
rake test_dump_vrp         # Dump all necessary fixture files
```

With the modification I suggested below, it shows all the test options

```
rake resque:failures:sort  # Sort the 'failed' queue for the redis_multi_queue failure backend
rake resque:work           # Start a Resque worker
rake resque:workers        # Start multiple Resque workers
rake test                  # Run tests
rake test:api              # Run tests for api
rake test:clustering       # Run tests for clustering
rake test:lib              # Run tests for lib
rake test:model            # Run tests for model
rake test:periodic         # Run tests for periodic
rake test:structure        # Run tests for structure
rake test_dump_vrp         # Dump all necessary fixture files
```

Rakefile
```diff
@@ -33,13 +33,58 @@ namespace :resque do
...
+
```
senhalil, line 87:
```suggestion
namespace :test do
  Rake::TestTask.new(:api){ |t| t.pattern = 'test/api/**/*_test.rb' }

  Rake::TestTask.new(:model){ |t| t.pattern = 'test/models/**/*_test.rb' }

  Rake::TestTask.new(:structure){ |t| t.test_files = ['test/api/**/*_test.rb', 'test/models/**/*_test.rb'] }

  Rake::TestTask.new(:lib){ |t| t.pattern = 'test/lib/**/*_test.rb' }

  Rake::TestTask.new(:clustering){ |t| t.pattern = 'test/**/*clustering*_test.rb' }

  Rake::TestTask.new(:periodic){ |t| t.pattern = 'test/**/*scheduling*_test.rb' }
end

```

This way `rake -T` shows the sub tasks correctly and we don't have to touch the `TEST` variable which stays free to be used injunction if necessary.

If we set the default value of `COV` correctly down below then we don't need to fiddle with it here.

README.md
```diff
@@ -89,6 +89,26 @@ TIME=true HTML=true APP_ENV=test bundle exec rake test
...
+APP_ENV=test bundle exec rake test TEST=test/real_cases_scheduling_test.rb
+```
+If you want to run only one specific test (let's say test_instance_clustered test only) you can use `focus` or call:
+```
+APP_ENV=test bundle exec rake test TESTOPTS="--name=test_instance_clustered"
+```
+Tests are splitted into several sets and you can replay only one or more set(s):
+```
+APP_ENV=test bundle exec rake test:api
+APP_ENV=test bundle exec rake test:api test:models
```
senhalil, line 104: I think, after your corrections we don't need the `APP_ENV=test` anymore.

Also, as you suggested, if the correct version of `rake` is the default then `bundle exec` is not necessary either but I guess this way is more secure..

senhalil (changes requested):

test/test_helper.rb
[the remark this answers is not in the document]
alain-andre: I disagree with this as when putting coverage to false, no-one will launch it with true.

Rakefile
alain-andre: About the **TEST** variable, these tasks are meant to launch a folder, I don't see the point of leaving the TEST variable free. However I prefer the `t.pattern` as it adds the description in `rake -T` command. 

Removing the `ENV['COV']` is not a good idea for me, I explain it below.

test/test_helper.rb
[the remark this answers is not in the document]
senhalil: If all tests set it to `ENV['COV'] = 'false'` in the Rakefile then what's the point ;) no one could launch it with `true` in that case!

Plus the default behavior of this part of the code should be in accord with the default behavior of the other parts. The procedures we define should let the CI and the developers know when to launch the tests with coverage or not. 

For example, with my suggestions, the CI can launch the tests always with `COV=true` and we can have check that fails whenever there is a regression in the coverage. Forcing the coverage in this manner would only mean that the devs will launch the tests with `COV=false rake test` on local.

Rakefile
senhalil: I think your argument "these tasks are meant to launch a folder" is incorrect because these tasks are meant to launch all the tests related to a "concept" and they are not necessarily in a single folder. 

For example, `:periodic` launches the following 4 tests
```
optimizer-api/test/real_cases_scheduling_solver_test.rb
optimizer-api/test/lib/heuristics/scheduling_functions_test.rb
optimizer-api/test/lib/heuristics/scheduling_instance_validity_test.rb
optimizer-api/test/lib/heuristics/scheduling_test.rb
```
but it doesn't launch the other tests in these folders -- `optimizer-api/test/lib/heuristics/`, `optimizer-api/test/`. 

I pointed out **TEST** because it can be used injunction to further **filter** the tests as follows: `rake test:periodic TEST="test/lib/heuristics/*"` would launch all the scheduling heuristic `library` tests but not the `real_cases` ones. For me leaving this variable free for the advanced use cases like this is a nice **cherry on the top**.

senhalil: I don't understand why you suggest not touching `ENV['COV']` here, it is the opposite of your suggestion below. There you suggest that we shouldn't make the default behavior without coverage but here you suggest leaving `ENV['COV']='false'` in place... As I explained below, I think the code should do its job and the procedures should do their job, we shouldn't make the code less elegant and more complex to make it handle the duty of CI procedures.

test/test_helper.rb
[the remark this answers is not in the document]
alain-andre: All test are set to COV=true, only the specifics one set the coverage to false. The coverage is actually set to false in the CI so if it's the only place coverage is calculated, there is indeed some changes to do.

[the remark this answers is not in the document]
senhalil: As far as I know, the local tests are launched with an alias which has `COVERAGE=false`, that is, without the modification above the coverage is never launched. With the modification above, the `COV=true` which we define in the main test will be enforced correctly because we check "if there exists true" in the environment vars --  `[ENV['COVERAGE'], ENV['COV']].include? 'true'`.

I still think this sort of task is something that we can handle very efficiently with CI and it shouldn't be left as a manual verification on local. What do you think @alain-andre?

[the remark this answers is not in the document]
fab-girard: for the moment, I prefer to keep coverage by default when replaying all tests (same thing for all projects), we will see other strategies when we will have an advanced CI

README.md
fab-girard: yes we don't need, as `bundle exec` in some cases, but I prefer to keep them in README to reveal the different ways to run tasks

README.md
fab-girard: what is your suggestion instead of `bundle exec rake -T`

Rakefile
fab-girard: did you test all tasks Hali with your suggestions?

test/test_helper.rb
```diff
@@ -29,9 +32,9 @@
 
 require 'minitest/reporters'
 Minitest::Reporters.use! [
-  Minitest::Reporters::ProgressReporter.new,
+  !ENV['TIME'] ? Minitest::Reporters::ProgressReporter.new : nil,
```
braktar, line 35: This line seems a deprecated duplicate of line 37

braktar (changes requested):

README.md
senhalil: with my suggestion below, it works as expected  -- i.e., it lists all the tests.

Rakefile
senhalil: :+1:  yeap they are all good

test/test_helper.rb
senhalil, line 20: If we want to keep it ON by default then I suggest the following which is more natural/easy to read 
```suggestion
if Rake.application.top_level_tasks.include?('test') && [ENV['COVERAGE'], ENV['COV']].exclude? 'false'
```

alain-andre (approved):

senhalil (approved):

senhalil:
I just resolved the conflict and @braktar's comment. We can merge this. 

I will create a ticket to use a combination of these subtask to launch the CI instead of using global env variable switches to control the tests.

PR #128, opened 2021-02-04
use first_solution_strategy instead of solver from pickup_delivery front

fonsecadeline:


public/js/pickup_delivery.js
```diff
@@ -116,14 +116,15 @@ $(document).ready(function() {
...
     if (data.customers.length > 0 && data.vehicles.length > 0) {
       if (debug) console.log('Build json from csv: ', data);
       var vrp = {points: [], units: [], shipments: [], services: [], vehicles: [], configuration: {
         preprocessing: {
-          cluster_threshold: 0
+          cluster_threshold: 0,
+          first_solution_strategy: correspondant[parseInt($('#optim-solver-parameter').val())]
         },
         resolution: {
-          solver_parameter: parseInt($('#optim-solver-parameter').val()),
```
braktar, line 126: Can't we directly display the heuristic names ?

fonsecadeline: You mean, for customers ? That is an option but do they really know which heuristic corresponds to which number ?

braktar: `solver_parameter` is deprecated. Every user should now use first_solution_strategy. We should not keep this logic even in facade.

It seems, to me, easier to associate resolution behavior to a name than to a number.

fonsecadeline: As discussed today in daily point : we agreed that we need both number and name not to get our users confused because they don't know the correspondance yet.

senhalil (approved):

public/js/pickup_delivery.js
senhalil: is this resolved or there are more things to do?

fonsecadeline: It is okay by me because modification we discussed would take too long : changing JS file to output each number and heuristic name. I am not sure much customers use this front anyway ?

senhalil: In that case, @braktar can you take another look so that we can move this PR forward

braktar (approved):

PR #131, opened 2021-02-05
First last possible day

fonsecadeline:
@senhalil  @braktar do we want to be able to provide a set of available periods ?


api/v01/entities/vrp_input.rb
```diff
@@ -329,6 +329,16 @@ module VrpMissions
       use :vrp_request_quantity
     end
     mutually_exclusive :quantity_ids, :quantities
+
+    optional(:available_indices_interval,
```
braktar, line 333:
```suggestion
    optional(:range_indices,
```

api/v01/entities/vrp_input.rb
```diff
@@ -329,6 +329,16 @@ module VrpMissions
...
+             type: Hash, desc: '(Scheduling only) Day indices where visits can take place') do
+      use :vrp_request_indice_range
+    end
+    optional(:available_date_interval,
```
braktar, line 337:
```suggestion
    optional(:range_date,
```

api/v01/entities/vrp_input.rb
braktar: Otherwise, it is not uniform both with the configuration definition and the mutually_exclusive constraint

api/v01/entities/vrp_input.rb
```diff
@@ -362,6 +372,16 @@ module VrpMissions
       use :vrp_request_quantity
     end
     mutually_exclusive :quantity_ids, :quantities
+
+    optional(:available_indices_interval,
```
braktar, line 376:
```suggestion
    optional(:range_indices,
```

api/v01/entities/vrp_input.rb
```diff
@@ -362,6 +372,16 @@ module VrpMissions
...
+             type: Hash, desc: '(Scheduling only) Day indices where visits can take place') do
+      use :vrp_request_indice_range
+    end
+    optional(:available_date_interval,
```
braktar, line 380:
```suggestion
    optional(:range_date,
```

braktar (changes requested):

api/v01/entities/vrp_input.rb
senhalil: I am not sure if these have the same meaning. 

scheduling configuration `range_indices` defines the horizon of the whole planning and these availability intervals are parts of the horizon where the service is "available" to be served. Is that the case @fonsecadeline ?

If that is the case, I would suggest the following naming `:available_index_range` and `:available_date_range`.

senhalil (approved):
We can improve the name of the new parameters, they don't sound "correct" English at the moment.

braktar (approved):

test/lib/heuristics/scheduling_test.rb
```diff
@@ -867,5 +867,37 @@ def test_empty_scheduling_result_when_no_mission
       end
       assert_equal 1 * 4, result[:routes].size
     end
+
+    def test_scheduling_wiht_unavailable_interval
```
braktar, line 871:
```suggestion
    def test_scheduling_with_unavailable_interval
```

models/vrp.rb
```diff
@@ -337,44 +337,91 @@ def self.remove_unecessary_units(hash)
       hash[:shipments]&.map{ |s| (s[:quantities] || []).delete_if{ |quantity| needed_units.exclude? quantity[:unit_id] } }
     end
 
-    def self.generate_schedule_indices_from_date(hash)
-      return hash if !hash[:configuration] || !hash[:configuration][:schedule] ||
-                     hash[:configuration][:schedule][:range_indices]
+    def self.convert_availability_dates_into_indices(element, hash, start_indice, type)
```
braktar, line 340:
```suggestion
    def self.convert_availability_dates_into_indices(element, hash, start_index, type)
```

models/vrp.rb
```diff
@@ -337,44 +337,91 @@ def self.remove_unecessary_units(hash)
...
+        when :schedule
+          [:unavailable_indices, :unavailable_date]
+        end
 
-      start_indice = hash[:configuration][:schedule][:range_date][:start].to_date.cwday - 1
-      end_indice = (hash[:configuration][:schedule][:range_date][:end].to_date - hash[:configuration][:schedule][:range_date][:start].to_date).to_i + start_indice
+      element[unavailable_indices] ||= []
+      element[unavailable_dates].to_a.each{ |unavailable_date|
+        element[unavailable_indices] |=
+          [(unavailable_date.to_date - hash[:configuration][:schedule][:range_date][:start]).to_i + start_indice]
```
braktar, line 354:
```suggestion
          [(unavailable_date.to_date - hash[:configuration][:schedule][:range_date][:start]).to_i + start_index]
```

models/vrp.rb
```diff
@@ -337,44 +337,91 @@ def self.remove_unecessary_units(hash)
...
+          end: (range[:end].to_date - hash[:configuration][:schedule][:range_date][:start]).to_i + start_indice,
+        }
       }
+      element.delete(:unavailable_date_ranges)
+    end
 
-      # remove vehicle unavailable_work_date
-      hash[:vehicles].each{ |vehicle|
-        next if vehicle[:unavailable_work_date].to_a.empty?
+    def self.collect_unvaible_day_indices(element, type, start_indice)
```
braktar, line 370:
```suggestion
    def self.collect_unavailable_day_indices(element, type, start_index)
```

models/vrp.rb
```diff
@@ -337,44 +337,91 @@ def self.remove_unecessary_units(hash)
...
+      return hash if !hash[:configuration] || !hash[:configuration][:schedule]
+
+      if hash[:configuration][:schedule][:range_indices] &&
+         hash[:services]&.any?{ |s| s[:unavailable_date_ranges] }
+        raise OptimizerWrapper::DiscordantProblemError.new(
+          'Date intervals are not compatible with schedule range indices'
+        )
+      end
+
+      start_indice, end_indice =
```
braktar, line 404:
```suggestion
      start_index, end_index =
```

braktar (changes requested):

models/vrp.rb
```diff
@@ -337,49 +337,96 @@ def self.remove_unecessary_units(hash)
...
+          end: (range[:end].to_date - hash[:configuration][:schedule][:range_date][:start]).to_i + start_index,
+        }
       }
+      element.delete(:unavailable_date_ranges)
+    end
 
-      # remove vehicle unavailable_work_date
-      hash[:vehicles].each{ |vehicle|
-        next if vehicle[:unavailable_work_date].to_a.empty?
+    def self.collect_unvaible_day_indices(element, type, start_index)
```
senhalil, line 370: unavailable

models/vrp.rb
```diff
@@ -337,49 +337,96 @@ def self.remove_unecessary_units(hash)
       hash[:shipments]&.map{ |s| (s[:quantities] || []).delete_if{ |quantity| needed_units.exclude? quantity[:unit_id] } }
     end
 
-    def self.generate_schedule_indices_from_date(hash)
-      return hash if !hash[:configuration] || !hash[:configuration][:schedule] ||
-                     hash[:configuration][:schedule][:range_indices]
+    def self.convert_availability_dates_into_indices(element, hash, start_index, type)
+      unavailable_indices, unavailable_dates =
```
braktar, line 341:
```suggestion
      index_keyword, date_keyword =
```

models/vrp.rb
```diff
@@ -337,49 +337,96 @@ def self.remove_unecessary_units(hash)
...
+        when :schedule
+          [:unavailable_indices, :unavailable_date]
+        end
 
-      start_indice = hash[:configuration][:schedule][:range_date][:start].to_date.cwday - 1
-      end_indice = (hash[:configuration][:schedule][:range_date][:end].to_date - hash[:configuration][:schedule][:range_date][:start].to_date).to_i + start_indice
+      element[unavailable_indices] ||= []
+      element[unavailable_dates].to_a.each{ |unavailable_date|
+        element[unavailable_indices] |=
+          [(unavailable_date.to_date - hash[:configuration][:schedule][:range_date][:start]).to_i + start_index]
```
braktar, line 354: `hash[:configuration][:schedule][:range_date][:start].to_i + start_index` could be put within a variable as it is used 3 times.

models/vrp.rb
```diff
@@ -337,49 +337,96 @@ def self.remove_unecessary_units(hash)
...
     end
 
-    def self.generate_schedule_indices_from_date(hash)
-      return hash if !hash[:configuration] || !hash[:configuration][:schedule] ||
-                     hash[:configuration][:schedule][:range_indices]
+    def self.convert_availability_dates_into_indices(element, hash, start_index, type)
+      unavailable_indices, unavailable_dates =
+        case type
+        when :visit
+          [:unavailable_visit_day_indices, :unavailable_visit_day_date]
```
senhalil, line 344: Why don't we make the rest of the code uniform like the way we did for the API.

That is, we have `unavailable_index_ranges` at the API level for services/shipments/vehicle/config and we can do `unavailable_indices` for all inside the code base instead of `unavailable_visit_day_indices`/`unavailable_work_day_indices`/`unavailable_indices` and `unavailable_visit_day_date`/`unavailable_visit_day_date`/`unavailable_date`. 

This separation (of the same exact logic "unavailability") makes the code more complex for no gain. For example here we needed to jump through hoops to make the parsing work for all objects.

models/vrp.rb
```diff
@@ -337,49 +337,96 @@ def self.remove_unecessary_units(hash)
...
 
-        vehicle[:unavailable_work_day_indices] = [] unless vehicle[:unavailable_work_day_indices]
-        vehicle[:unavailable_work_day_indices] += vehicle[:unavailable_work_date].collect{ |unavailable_date|
-          (unavailable_date.to_date - hash[:configuration][:schedule][:range_date][:start]).to_i + start_indice
-        }.compact
-        vehicle.delete(:unavailable_work_date)
-      }
+      element[unavailable_indices] += element[:unavailable_index_ranges].collect{ |range|
+        (range[:start]..range[:end]).to_a
+      }.flatten.uniq.select{ |day| day >= start_index }
```
braktar, line 385: selecting within a range is heavier than working with [range[:start], start_index].max
and then expanding the range

senhalil (commented):
What do you think @fonsecadeline, is it doable making the `unavailable_indices` uniform for services/shipments/vehicles/config throughout the code base?

models/vrp.rb
senhalil: I didn't mark the other occasions by the way

senhalil: I meant "I assume you will use find-all to correct the typo everywhere and I didn't comment on the other `collect_unvaible_day_indices` typos"

models/vrp.rb
fonsecadeline: Do you mean making names uniform within the code ? (but still having _unavailable_indices_ for each vehicle/service/shipment/schedule ?). We could do that in our internal model, you are right.

models/vrp.rb
```diff
@@ -337,49 +337,96 @@ def self.remove_unecessary_units(hash)
...
+      if hash[:configuration][:schedule][:range_indices] &&
+         hash[:services]&.any?{ |s| s[:unavailable_date_ranges] }
+        raise OptimizerWrapper::DiscordantProblemError.new(
+          'Date intervals are not compatible with schedule range indices'
+        )
+      end
+
+      start_index, end_index =
+        if hash[:configuration][:schedule][:range_indices]
+          hash[:configuration][:schedule][:range_indices][:start]
```
braktar, line 406: in this case, isn't `end_index` set to nil ?

braktar (changes requested):

models/vrp.rb
fonsecadeline: haha yes, I understood, the face was right but the smiley name was not. :/

models/vrp.rb
senhalil: bien vu ! if this didn't raise an error during testing this means we have no test passing through this function, @fonsecadeline this might be a good occasion to add a unit test on `generate_schedule_indices_from_date(hash)` or `filter(hash)`

models/vrp.rb
fonsecadeline: I am not sure if this is possible : only difference between two dates can be converted to_i, it seems to me that this value needs to be computed from each corresponding date.

models/vrp.rb
fonsecadeline: yes it is, but that is expected because it won't be used in this case.

models/vrp.rb
braktar: Indeed, `.to_i` has no meaning, as it requires a zero. There is multiple way to obtain an integer from a date.
The difference between two dates returns a rational. But you can obtain this rational with `.ajd` or you can get the integer using `.jd` (if you want it to) (julian calendar)

`.ld` in Lilian calendar

models/vrp.rb
fonsecadeline: we do test this case

fonsecadeline: should I return nil explicitly ?

braktar: But, you put `end_index` in `hash[:configuration][:schedule][:range_indices]`

Isn't this value used ?

braktar: ! But you do `return hash if hash[:configuration][:schedule][:range_indices]` if you have a range indices

fonsecadeline: Right, so when we have range_indices we do not need end_index. 
Anyway, I am changing this to ensure unavailable days are between schedule start and schedule end so I will return end_index in this case too.

models/shipment.rb
```diff
@@ -37,7 +37,7 @@ class Shipment < Base
     # validates_numericality_of :visits_number
 
     field :unavailable_visit_indices, default: []
-    field :unavailable_visit_day_indices, default: [] # extends unavailable_visit_day_date
+    field :unavailable_days, default: [] # extends unavailable_visit_day_date and unavailable_visit_day_indices
```
braktar, line 40: It may be interesting to have a Set instead of an array. The order doesn't really matter, we only want to know if a value is included ? `include?` method of a Set is in O(1) against O(n) for an Array

models/vrp.rb
```diff
@@ -81,7 +81,7 @@ class Vrp < Base
     field :restitution_allow_empty_result, default: false
 
     field :schedule_range_indices, default: nil # extends schedule_range_date
-    field :schedule_unavailable_indices, default: [] # extends unavailable_date
+    field :schedule_unavailable_days, default: [] # extends unavailable_date and schedule_unavailable_indices
```
senhalil, line 84: This

braktar (approved):

PR #132, opened 2021-02-08
Wrapper object within scheduling class

fonsecadeline:
Needs : https://github.com/Mapotempo/optimizer-api/pull/130 to be merged

### DIFF
--- a/lib/heuristics/scheduling_heuristic.rb
+++ b/lib/heuristics/scheduling_heuristic.rb
@@ -22,8 +22,8 @@ require './lib/output_helper.rb'
 require './lib/heuristics/concerns/scheduling_data_initialisation'
 require './lib/heuristics/concerns/scheduling_end_phase'
 
-module Heuristics
-  class Scheduling
+module Wrappers
+  class SchedulingHeuristic < Wrapper
     include SchedulingDataInitialization
     include SchedulingEndPhase
 
@@ -82,7 +82,7 @@ module Heuristics
     def compute_initial_solution(vrp, &block)
       if vrp.services.empty?
         # TODO : create and use result structure instead of using wrapper function
-        vrp[:preprocessing_heuristic_result] = Wrappers::Wrapper.new.empty_result('heuristic', vrp)
+        vrp[:preprocessing_heuristic_result] = empty_result('heuristic', vrp)
         return []
       end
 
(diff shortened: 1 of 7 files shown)


lib/heuristics/scheduling_heuristic.rb
```diff
@@ -63,6 +63,7 @@ def initialize(vrp, job = nil)
       @uninserted = {}
       @missing_visits = {}
 
+      @wrapper_tool = Wrappers::Wrapper.new
```
senhalil, line 66: I am not sure if we need to generate a new Wrapper object in fact any  `OptimizerWrapper.config[:services][service]` object would do.

We can name the variable simply `@wrapper`, with `tool` it sounds as if it is something different.

senhalil (commented):
it is a lot better! not creating a new wrapper object for each service  :ok_hand:

braktar (approved):

senhalil (approved):

lib/heuristics/scheduling_heuristic.rb
```diff
@@ -28,6 +28,8 @@ class Scheduling
     include SchedulingEndPhase
 
     def initialize(vrp, job = nil)
+      @wrapper = OptimizerWrapper.config[:services][:ortools]
```
fab-girard, line 31: Why ortools is used instead of another solver?

senhalil: It was me who insisted on this. In fact it doesn't matter which solver is used because the `@wrapper` variable is used to call functions from the parent class of the solvers -- i.e., `Wrapper` -- since heuristic doesn't have a copy of the `Wrapper` object the options were to create a new `Wrapper` object here or use the existing objects.

fab-girard: I think we can use/instantiate a base class with necessary methods to work

senhalil: Yeah that would be more appropriate. I don't think I know how to that in Ruby. Cheers Fabien!

fonsecadeline:
Waiting for : https://gitlab.com/mapotempo/optimizer-api/-/issues/695

fonsecadeline:
This will also fix a bug encountered by one of our users.

PR #134, opened 2021-02-11
Use passenger in docker.

alain-andre:
This PR makes use of passenger instead of puma.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ services:
 before_install:
   - export REGISTRY=${REGISTRY:-registry.mapotempo.com/}
   - source ci-utils/version.sh
-  - docker build --build-arg VROOM_VERSION=${VROOM_VERSION} --build-arg OPTIMIZER_ORTOOLS_VERSION=${OPTIMIZER_ORTOOLS_VERSION} -f docker/Dockerfile -t registry.test.com/mapotempo/optimizer-api:latest .
+  - docker build --build-arg BUNDLE_WITHOUT=development --build-arg VROOM_VERSION=${VROOM_VERSION} --build-arg OPTIMIZER_ORTOOLS_VERSION=${OPTIMIZER_ORTOOLS_VERSION} -f docker/Dockerfile -t registry.test.com/mapotempo/optimizer-api:latest .
   - docker swarm init
   - mkdir -p ./redis
   - mkdir -p ./redis-count
--- a/ci-utils/tests.sh
+++ b/ci-utils/tests.sh
@@ -31,7 +31,7 @@ case "$1" in
 esac
 
 max_time=60 # Time in secondes
-total_dockercompose_services=4
+total_dockercompose_services=5
 for ((cpt=1;cpt<=$max_time;cpt++));
 do
   if [ $cpt == ${max_time} ];
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -2,14 +2,19 @@ ARG BRANCH=${BRANCH:-ce}
 ARG OPTIMIZER_ORTOOLS_VERSION=${OPTIMIZER_ORTOOLS_VERSION}
 ARG REGISTRY=${REGISTRY:-registry.mapotempo.com/}
 ARG VROOM_VERSION=${VROOM_VERSION}
+ARG BUNDLE_WITHOUT="development test"
 
 # Install Vroom
 FROM vroomvrp/vroom-docker:${VROOM_VERSION:-v1.8.0} as vroom
 
 # Rake
 FROM ${REGISTRY}mapotempo-${BRANCH}/optimizer-ortools:${OPTIMIZER_ORTOOLS_VERSION:-v1.4.0}
+ARG BUNDLE_WITHOUT
 
 ENV LANG C.UTF-8
+
+# Set correct environment variables.
+ENV HOME /root
 
 RUN apt-get update > /dev/null
 RUN libgeos=$(apt-cache search 'libgeos-' | grep -P 'libgeos-\d.*' | awk '{print $1}')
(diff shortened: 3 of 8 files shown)


docker/Dockerfile
```diff
@@ -19,7 +22,7 @@ COPY . /srv/app
 WORKDIR /srv/app
 RUN gem install bundler --version 2.1.4
 RUN bundle --version
-RUN bundle check || bundle install --full-index --without development -j $(nproc)
+RUN bundle install --path vendor/bundle --full-index --without development -j $(nproc)
```
fab-girard, line 25: is it possible to add `--without development test` or not? my question is in case travis CI uses this image

docker/snippets/app.conf
```diff
@@ -0,0 +1,13 @@
+
+location / {
+  send_timeout 10m;
+  include snippets/cors.conf;
+
+  client_max_body_size 100M;
```
senhalil, line 6: Does this limit apply to GET as well? If so we need to increase this limit because we already have solution files bigger than 100M.

.travis.yml
```diff
@@ -1,5 +1,8 @@
 services:
   - docker
+cache:
+  directories:
+    - vendor/bundle
```
senhalil, line 5: If the cache is not working as expected, can we remove the cache logic from this PR to another WIP PR?

docker/Dockerfile
alain-andre: Yes this image is used by Travis to test. But I think we should use a variable allowing use to remove test gems for production image.

Like:
```dockerfile
RUN bundle install --path vendor/bundle --full-index --without ${WITHOUT} -j $(nproc)
```

where in Travis `WITHOUT=development` and for production image `WITHOUT="development test"`

What do you thing about this option ?

docker/snippets/app.conf
```diff
@@ -0,0 +1,13 @@
...
+
+  passenger_app_env production;
+  passenger_enabled on;
+  passenger_app_root /srv/app;
+
+  passenger_env_var APP_ENV production;
```
senhalil, line 12: This `APP_ENV=production` is gonna be okay on Travis, right?

ci-utils/tests.sh
```diff
@@ -61,6 +61,9 @@ done
 while sleep 60; do echo "=====[ $SECONDS seconds still running ]====="; done &
 
 CONTAINER=${DOCKER_SERVICE_NAME}.1.$(docker service ps -f "name=${DOCKER_SERVICE_NAME}.1" ${DOCKER_SERVICE_NAME} -q --no-trunc | head -n1)
+
+# Copy vendor/bundle for cache
+docker cp ${CONTAINER}:/srv/app/vendor/bundle vendor/bundle
```
senhalil, line 66: if the copying is just for Travis cache then we don't need this either.

senhalil (commented):
I don't know anything about passenger, I commented on the stuff that I could understand.

.travis.yml
alain-andre: I do change the directory to `/srv/app/vendor/bundle` and test it. If it does not fix the problem I will remove the caching tasks.

docker/snippets/app.conf
alain-andre: I used the configuration actually in production, what the max size you think we should allow ?

docker/snippets/app.conf
alain-andre: It worked fine on my Travis tests but your question is interesting. Thing is Nginx overrides all environment variable and Passenger set it to production but I can pass the variable in the docker compose.

docker/snippets/app.conf
senhalil: If the limit applies to GETs then 200M would be better because we already have solution of 134M

.travis.yml
senhalil: :crossed_fingers:

docker/snippets/app.conf
senhalil: Yeah it is better if if have one central location for setting env variables and the rest of the code uses it as needed.

docker/Dockerfile
senhalil: if `APP_ENV` is already defined before we arrive here, we can condition on `APP_ENV=='test'` like `$([[ $APP_ENV = 'test' ]] && echo 'development' || echo 'development test')` without an extra env variable. this is bash syntax, I don't know the equivalent in Dockerfile syntax.

.travis.yml
alain-andre: It doe not work, I remove it from the PR.

docker/Dockerfile
alain-andre: In Dockerfile we use argument for this kind of value as it is only use for build. The commit contains now the default value with is *"development test"* allowing the production build not to change anything and in travis build, I added a `--build-arg WITHOUT=development`

ci-utils/tests.sh
alain-andre: I removed it.

docker/snippets/app.conf
alain-andre: Ok, nice of @senhalil to put the "GET" back in your answer. This only concerns the size of the incoming requests so I won't change it.

senhalil: no worries happy to help :ok_hand:

docker/Dockerfile
```diff
@@ -2,15 +2,20 @@ ARG BRANCH=${BRANCH:-ce}
...
 ARG REGISTRY=${REGISTRY:-registry.mapotempo.com/}
 ARG VROOM_VERSION=${VROOM_VERSION}
+ARG WITHOUT="development test"
 
 # Install Vroom
 FROM vroomvrp/vroom-docker:${VROOM_VERSION:-v1.8.0} as vroom
 
 # Rake
 FROM ${REGISTRY}mapotempo-${BRANCH}/optimizer-ortools:${OPTIMIZER_ORTOOLS_VERSION:-v1.4.0}
+ARG WITHOUT
```
senhalil, line 12: This might be a Dockerfile specific thing but in case it is a typo, is this double definition normal? `WITHOUT` is already defined as argument on line 5 above.

alain-andre: Yes it's definitely a Dockerfile specific trick. `ARG` must be present in second `FROM` block to be used in.

senhalil: Cheers! Then I think this is ready.

senhalil (approved):
It looks good to me with the disclaimer that I don't know much about passenger/docker/etc :) but if all tests are passing then it should be alright :crossed_fingers:

fonsecadeline (approved):
Same as @senhalil :)

PR #138, opened 2021-02-15
remove schedule_expanded_vehicles from configuration

fonsecadeline:


wrappers/wrapper.rb
```diff
@@ -931,23 +931,21 @@ def unassigned_rests(vrp)
...
-      else
-        vrp.vehicles
-      end
+      periodic = Interpreters::PeriodicVisits.new(vrp)
+      periodic.generate_vehicles(vrp)
     end
 
-    def empty_result(solver, vrp, unassigned_reason = nil)
-      vrp.vehicles = expand_vehicles_for_consistent_empty_result(vrp)
+    def empty_result(solver, vrp, unassigned_reason = nil, already_expanded = true)
```
braktar, line 938: we could progressively migrate from multiple optional parameters, to a single hash `options` which contains those keys and values ?

fonsecadeline: I tried this, but we sometimes need to overload only one of those. 

Wouldn't this force us to provide them all ? For instance if we provide unassigned_reason then we need to specify already_expanded, right ?

senhalil: I think its better if we can keep this PR as simple as possible. I need it to rebase my PR which is a prerequisite to another PR which I am trying to create...

Lets create a ticket for this "options as hash" idea and discuss the details on the ticket but I think I am already sold :) 

@fonsecadeline it doesn't force us https://codereview.stackexchange.com/a/73503. Instead of variables, they become hash keys, and this forces the caller to write the name of the key.

Like this for example, `empty_result(solver, vrp, already_expanded: true)` here there is a "hidden" hash in fact `empty_result(solver, vrp, { already_expanded: true })` of course we need to decide for a way to separate "real" arguments from "options"

optimizer_wrapper.rb
```diff
@@ -200,7 +200,8 @@ def self.solve(service_vrp, job = nil, block = nil)
       multi_modal = Interpreters::MultiModal.new(vrp, service)
       optim_result = multi_modal.multimodal_routes
     elsif vrp.vehicles.empty? || (vrp.services.empty? && vrp.shipments.empty?)
-      optim_result = Wrappers::Wrapper.new.empty_result(service.to_s, vrp, 'No vehicle available for this service')
+      optim_result = config[:services][service].empty_result(
+        service.to_s, vrp, 'No vehicle available for this service', false)
```
senhalil, line 204: in fact we do not need the `solver` argument -- i.e., `service.to_s` each solver needs to surcharge the empty_result function so that, wrapper gets the necessary information from the child class.

For example, we can have the following function in `ortools.rb`, `vroom.rb` and `demo.rb`
```
def empty_result(*args)
  super("ortools"/"vroom"/"demo", *args)
end
```

We can leave this for another time as well. But we should keep in mind that solves (ortools,vroom,demo) are based on `Wrapper` class so, they can overload its functions for this kind of purposes and call the original funciton

senhalil: I didn't test the code above by the way but the implementation is along the lines...

senhalil (approved):

braktar (approved):

PR #139, opened 2021-02-15
fix duplicate_safe function

fonsecadeline:


lib/interpreters/periodic_visits.rb
```diff
@@ -510,7 +510,8 @@ def generate_relations_on_periodic_vehicles(vrp, list)
     private
 
     def get_original_values(original, options)
-      [original.attributes.keys + options.keys - [:unavailable_work_day_indices] + [:unavailable_days]].flatten.each_with_object({}) { |key, data|
+      old_keys = [:unavailable_work_day_indices, :unavailable_visit_day_indices]
```
senhalil, line 513: why do we need this? since these fields don't exists `original.attributes.keys` shouldn't return them unless it is a "dump" issue again?

senhalil (changes requested):

PR #140, opened 2021-02-15
max_split_size supports skills and initial routes

senhalil:
Complete overhaul of split_solve (max_split_size) algorithm
- lays the ground work for supporting P&D in split_solve
- skills and initial routes are now supported in split_solve
- more balanced split
- decreased optim-api overhead
- preliminary results show better vehicles utilization and less unplanned services

  |   | Percent Unassigned |   | # Vehicles Used |   | Solution Time |  
-- | -- | -- | -- | -- | -- | -- | --
Max_split_value | Total Vehicles | old | new | old | new | old | new
300 | 15 | 50.6% | 48.5% | 15.0 | 15.0 | 388 | 267
  | 30 | 26.7% | 25.2% | 30.0 | 30.0 | 396 | 282
  | 50 | 5.5% | 4.9% | 50.0 | 49.5 | 288 | 294
  | 60 | 2.2% | 2.1% | 54.0 | 52.8 | 322 | 279
500 | 35 | 17.0% | 0.0% | 32.0 | 32.0 | 94 | 93

PS: @fonsecadeline and @braktar I didn't remove the old split_solve algorithm in case we need to take a quick look at something in the old code during the reviews but once it is validated, I shall remove the old split_solve logic and rename the "new_split_xxx" functions accordingly

### DIFF
--- a/lib/interpreters/split_clustering.rb
+++ b/lib/interpreters/split_clustering.rb
@@ -376,7 +376,7 @@ module Interpreters
         raise OptimizerWrapper::ClusteringError, 'Usage of options[:entity] requires that number of clusters (nb_clusters) is equal to number of vehicles in the vrp.'
       end
 
-      default_options = { max_iterations: 300, restarts: 10, cut_symbol: :duration }
+      default_options = { max_iterations: 300, restarts: 10, cut_symbol: :duration, build_sub_vrps: true }
       options = default_options.merge(options)
       vrp = service_vrp[:vrp]
       # Split using balanced kmeans
@@ -415,13 +415,18 @@ module Interpreters
         log "Balanced K-Means (#{toc - tic}sec): split #{result_items.sum(&:size)} activities into #{result_items.map(&:size).join(' & ')}"
         log "Balanced K-Means (#{toc - tic}sec): split #{data_items.size} data_items into #{clusters.map{ |c| "#{c.data_items.size}(#{c.data_items.map{ |i| i[3][options[:cut_symbol]] || 0 }.inject(0, :+)})" }.join(' & ')}"
 
-        result_items.collect.with_index{ |result_item, result_index|
-          next if result_item.empty?
+        if options[:build_sub_vrps]
+          result_items.collect.with_index{ |result_item, result_index|
+            next if result_item.empty?
 
-          vehicles_indices = [result_index] if options[:entity] == :work_day || options[:entity] == :vehicle
-          # TODO: build_partial_service_vrp can work directly with the list of services instead of ids.
-          build_partial_service_vrp(service_vrp, result_item.collect(&:id), vehicles_indices, options[:entity])
-        }.compact
+            vehicles_indices = [result_index] if options[:entity] == :work_day || options[:entity] == :vehicle
+            # TODO: build_partial_service_vrp can work directly with the list of services instead of ids.
+            build_partial_service_vrp(service_vrp, result_item.collect(&:id), vehicles_indices, options[:entity])
+          }.compact
+        else
+          # list of services by vehicle
+          result_items
+        end
       else
         log 'Split is not available if there are services with no activity, no location or if the cluster size is less than 2', level: :error
 
--- a/lib/interpreters/split_clustering.rb
+++ b/lib/interpreters/split_clustering.rb
@@ -122,72 +122,303 @@ module Interpreters
 
     def self.split_solve_candidate?(service_vrp)
       vrp = service_vrp[:vrp]
-      empties_or_fills = vrp.services.select{ |s| s.quantities.any?(&:fill) || s.quantities.any?(&:empty) }
-      depot_ids = vrp.vehicles.flat_map{ |vehicle| [vehicle.start_point_id, vehicle.end_point_id].compact }.uniq
+      if service_vrp[:split_level].nil?
+        empties_or_fills = vrp.services.select{ |s| s.quantities.any?(&:fill) || s.quantities.any?(&:empty) }
+        depot_ids = vrp.vehicles.flat_map{ |vehicle| [vehicle.start_point_id, vehicle.end_point_id].compact }.uniq
 
-      !vrp.scheduling? &&
-        vrp.preprocessing_max_split_size &&
-        vrp.vehicles.size > 1 &&
-        (vrp.resolution_vehicle_limit.nil? || vrp.resolution_vehicle_limit > 1) &&
-        (vrp.shipments.size + vrp.services.size - empties_or_fills.size) > vrp.preprocessing_max_split_size &&
-        vrp.shipments.all?{ |s| depot_ids.include?(s.pickup.point_id) || depot_ids.include?(s.delivery.point_id) }
+        !vrp.scheduling? &&
+          vrp.preprocessing_max_split_size &&
+          vrp.vehicles.size > 1 &&
+          (vrp.resolution_vehicle_limit.nil? || vrp.resolution_vehicle_limit > 1) &&
+          (vrp.shipments.size + vrp.services.size - empties_or_fills.size) > vrp.preprocessing_max_split_size &&
+          vrp.shipments.all?{ |s| depot_ids.include?(s.pickup.point_id) || depot_ids.include?(s.delivery.point_id) }
+      else
+        ss_data = service_vrp[:split_solve_data]
+        current_vehicles = ss_data[:current_vehicles]
+        service_vehicle_assignments = ss_data[:service_vehicle_assignments] # no empties_or_fills in here
+
+        current_vehicles.size > 1 &&
+          (ss_data[:current_vehicle_limit].nil? || ss_data[:current_vehicle_limit] > 1) &&
+          current_vehicles.sum{ |v| service_vehicle_assignments[v.id].size } > vrp.preprocessing_max_split_size
+      end
     end
 
+    # TODO: 0- see below, there are multiple
+    # TODO: 1- implement a notion of "same_vehicle" relation inside balanced_vrp_clustering gem
+    # TODO: 2- decrease iteration-complexity by "relaxing" the convergence limits (movement, balance-violation) of k-means for max_split?
+    # TODO: 3- decrease point-complexity by "grouping" by lat/lon more aggressively for this split
+    # TODO: 4- decrease vehicle-complexity by improving balanced_vrp_clustering
+    #          - by not re-calculating the item-vehicle compatibility every iteration and
+    #          - by considering only a subset of compatible vehicles
+    # TODO: 5- improve best cluster metric so that it considers depot distance and cluster size;
+    #          otherwise, metric can "penalize" better splits
     def self.split_solve(service_vrp, job = nil, &block)
       log '--> split_solve (clustering by max_split)'
       vrp = service_vrp[:vrp]
-      available_vehicle_ids = vrp.vehicles.collect(&:id)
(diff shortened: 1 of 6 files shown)


senhalil:
In fact it's better if you wait for PR #124 to be merged; otherwise, since the dev is based on configuration dev, the commits of that PR appear here too and it would be hard to do the review right now.

senhalil:
`create_fake_split_vrp`, `create_fake_sub_split_vrp` need renaming...

Basically `create_fake_split_vrp` creates the "fake" VRP we generate after splitting the original problem by vehicles. In the fake VRP, the points/services are created by merging all the points of each vehicle zone. That is, there is one service for each non-empty vehicle zone. We use this fake VRP to do the 2-splits (left/right) until max_split_size condition is satisfied. That way instead of calling kmeans for hundreds of points, we call it for `# of vehicles` of points.

I thought about `vehicle_vrp` but that wasn't clear enough either... @braktar @fonsecadeline any suggestions?

braktar:
The points are merged, so I would call it merged_vrp ? (I use it on my method)

I could also suggest : representative_vrp (as one point is 'elected' to represent the others)
Or simply high_level_vrp, as the point of view is macro

senhalil:
I like both `merged_vrp` and `representative_vrp` they give the clear indication that it is not the original VRP and it is a version that gone through a process. But I would keep "merge" for another case because it can be used for multiple VRPs getting "merged".

If Adeline doesn't come up with something better, I might go with the `representative_vrp` I think. It is a bit mouthful though, what about `reduced_vrp`? It clearly implies that this VPR is a projection of the original VRP  and it fits in the Ruby "reduce" convention :) -- reduction for points is averaging and reduction for services summation of duration and visit.

lib/interpreters/split_clustering.rb
```diff
@@ -663,64 +960,59 @@ def collect_data_items_metrics(vrp, cumulated_metrics, basic_split)
...
-            s_characteristics = {
-              v_id: [s[:sticky_vehicle_ids]].flatten.compact,
-              skills: related_skills,
-              day_skills: day_skills
-            }
+          location =
+            if s.activity
+              s.activity.point.location
+            elsif s.activities.size.positive?
+              raise UnsupportedProblemError, 'Clustering is not supported yet if one service has serveral activies.'
```
braktar, line 971:
```suggestion
              raise UnsupportedProblemError, 'Clustering is not supported yet if one service has several activities.'
```

lib/interpreters/split_clustering.rb
```diff
@@ -129,6 +120,302 @@ def self.generate_split_vrps(service_vrp, _job = nil, block = nil)
...
+
+        # RECURSIVELY CALL new_split_solve_core AND MERGE RESULTS
+        results = sides.collect.with_index{ |side, index|
+          service_vrp[:split_level] = split_level + 1
+
+          ss_data[:current_vehicles] = side
+
+          vehicle_limit_ratio = current_vehicle_limit.to_f * side.size / enum_current_vehicles.size
+          # Warning: round does not work if there is an even "half" split
+          ss_data[:current_vehicle_limit] == current_vehicle_limit &&
```
braktar, line 230: What is the purpose of this line as there is no affectation ?

lib/interpreters/split_clustering.rb
```diff
@@ -129,6 +120,302 @@ def self.generate_split_vrps(service_vrp, _job = nil, block = nil)
...
+      sub_vrp.matrices = o_vrp.matrices
+      sub_vrp.units = o_vrp.units
+      sub_vrp.rests = o_vrp.rests
+      sub_vrp.zones = o_vrp.zones
+      sub_vrp.subtours = o_vrp.subtours
+
+      sub_vrp.configuration = Oj.load(Oj.dump(o_vrp.config)) # time and other limits are correct below
+      # split the limits
+      sub_vrp.resolution_vehicle_limit = ss_data[:current_vehicle_limit] + ss_data[:transferred_vehicle_limit] if ss_data[:current_vehicle_limit]
+      ratio = (sub_vrp.services.size + sub_vrp.shipments.size).to_f / (o_vrp.services.size + o_vrp.shipments.size)
```
braktar, line 302:
```suggestion
      ratio = (sub_vrp.services.size + 2 * sub_vrp.shipments.size).to_f / (o_vrp.services.size + 2 * o_vrp.shipments.size)
```

lib/interpreters/split_clustering.rb
```diff
@@ -129,6 +120,302 @@ def self.generate_split_vrps(service_vrp, _job = nil, block = nil)
...
+      split_level = service_vrp[:split_level]
+      ss_data = service_vrp[:split_solve_data]
+
+      if split_solve_candidate?(service_vrp) # if it still needs splitting
+        enum_current_vehicles = ss_data[:current_vehicles].select # Enumerator to keep a local copy of current_vehicles
+        current_vehicle_limit = ss_data[:current_vehicle_limit]
+
+        # SPLIT current_vehicles list (by-vehicle-centroids) to create two "sides"
+        sides = split_balanced_kmeans(
+          { vrp: create_fake_sub_split_vrp(ss_data) }, 2,
```
braktar, line 210: Using the work 'fake' seems like (to me) we are creating some stub object and not really representative objects

braktar (changes requested):
Mostly comments. I will to look at it multiple times to really get the logic

lib/interpreters/split_clustering.rb
```diff
@@ -122,15 +122,298 @@ def self.generate_split_vrps(service_vrp, _job = nil, block = nil)
...
+        relations: relations
+      }, false)
+    end
+
+    def self.create_fake_sub_split_vrp(split_solve_data)
+      split_vrp = split_solve_data[:split_vrp]
+      sub_split_vrp = ::Models::Vrp.create({ name: 'SUB_SPLIT_VRP' }, false)
+      sub_split_vrp.vehicles = split_vrp.vehicles # 2-fake-vehicles
+      sub_split_vrp.services = split_vrp.services.select{ |service|
+        split_solve_data[:current_vehicles].any?{ |v| v.id == service.id }
```
fonsecadeline, line 361: This line is confusing to me. I just want to make sure this is not a bug, but I still need to investigate more to make sure split_solve_data[:current_vehicles] contain service.ids.

fonsecadeline: I feel like split_solve_data[:current_vehicles] is define by  ```current_vehicles: vrp.vehicles.map{ |v| v }```

lib/interpreters/split_clustering.rb
fonsecadeline: I quite agree with this, it seems it is a temporary object which does not matter.
What about ```create_current_sub_split_vrp```?

fonsecadeline (commented):
This seems right for me, I will probably check it again tomorrow to review with a better understanding.

lib/interpreters/split_clustering.rb
```diff
@@ -129,6 +120,302 @@ def self.generate_split_vrps(service_vrp, _job = nil, block = nil)
...
+      vrp = service_vrp[:vrp]
+      if service_vrp[:split_level].nil?
+        empties_or_fills = vrp.services.select{ |s| s.quantities.any?(&:fill) || s.quantities.any?(&:empty) }
+        depot_ids = vrp.vehicles.flat_map{ |vehicle| [vehicle.start_point_id, vehicle.end_point_id].compact }.uniq
+
+        !vrp.scheduling? &&
+          vrp.preprocessing_max_split_size &&
+          vrp.vehicles.size > 1 &&
+          (vrp.resolution_vehicle_limit.nil? || vrp.resolution_vehicle_limit > 1) &&
+          (vrp.shipments.size + vrp.services.size - empties_or_fills.size) > vrp.preprocessing_max_split_size &&
```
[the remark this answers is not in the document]
senhalil: in fact each shipment acts as a "single" point inside clustering (since we cannot separate the pickup and delivery of a shipment) so I find this condition more clear which handles the extreme case of "max_split_size=1" naturally as well.

lib/interpreters/split_clustering.rb
```diff
@@ -129,6 +120,302 @@ def self.generate_split_vrps(service_vrp, _job = nil, block = nil)
...
+      vrp.services -= empties_or_fills
+      # TODO: 0- relations needs to be taken into account inside clustering during this split
+      #          - the services need to be in the same route (and by extension, in the same sub-vrp) for the following relations:
+      #            => same_route, order, sequence, shipment
+      split_by_vehicle = split_balanced_kmeans(service_vrp, vrp.vehicles.size, cut_symbol: :duration, restarts: 2, build_sub_vrps: false)
+
+      # ss_data
+      service_vrp[:split_level] = 0
+      service_vrp[:split_solve_data] = {
+        current_vehicles: vrp.vehicles.map{ |v| v }, # new array but original objects
```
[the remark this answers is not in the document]
senhalil: I didn't know this, cheers! it is "longer" :p but definitely clearer!

fonsecadeline (approved):

lib/interpreters/split_clustering.rb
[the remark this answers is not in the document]
senhalil: I resolve this convo, feel free to unresolve it if you want to discuss further

lib/interpreters/split_clustering.rb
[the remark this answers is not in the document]
senhalil: done

lib/interpreters/split_clustering.rb
senhalil: I renamed it to `representative_sub_vrp` as we decided

lib/interpreters/split_clustering.rb
braktar: `==` performs a comparison and returns a boolean
`&& (index.zero? ? vehicle_limit_ratio.ceil : vehicle_limit_ratio.floor)` is only performed if the previous assertion is true

`vehicle_limit_ratio.ceil` or `vehicle_limit_ratio.floor` are never assigned ?

senhalil: I got your question now! `==` is a bug, it should have been `=` 
(I did not see the mistake, even after your question and writing an answer :facepalm:) 

So it is as follows normally
```ruby
          ss_data[:current_vehicle_limit] = current_vehicle_limit &&
            (index.zero? ? vehicle_limit_ratio.ceil : vehicle_limit_ratio.floor)
```

lib/interpreters/split_clustering.rb
```diff
@@ -129,61 +120,306 @@ def self.generate_split_vrps(service_vrp, _job = nil, block = nil)
...
+
+    def self.create_sub_vrp(split_solve_data)
+      ss_data = split_solve_data
+      o_vrp = ss_data[:original_vrp]
+
+      sub_vrp = ::Models::Vrp.create({}, false)
+
+      # Select the vehicles and services belonging to this sub-problem from the service_vehicle_assignments
+      sub_vrp.vehicles = ss_data[:current_vehicles] + ss_data[:transferred_vehicles]
+      sub_vrp.vehicles.each{ |v| v.cost_fixed = 1 if v.cost_fixed == 0 }
```
braktar, line 270: Do you set cost_fixed to 1 to avoid dicho ? Can't we call it after split_solve, to split really big problems and avoid requesting large matrices ?

braktar (approved):

PR #143, opened 2021-02-18
timewindow start is zero by default

fonsecadeline:


wrappers/ortools.rb
```diff
@@ -531,7 +531,7 @@ def parse_output(vrp, _services, points, _matrix_indices, _cost, _iterations, ou
               current_load: 0
             }
           }
-          route_start = (vehicle.timewindow && vehicle.timewindow[:start]) ? vehicle.timewindow[:start] : 0
+          route_start = vehicle.timewindow ? vehicle.timewindow[:start] : 0
```
braktar, line 534:
```suggestion
          route_start = vehicle.timewindow&.start || 0
```

optimizer_wrapper.rb
```diff
@@ -927,15 +927,15 @@ def self.zip_cluster(vrp, cluster_threshold, force_cluster)
             # find intersection with tw of service_tw
             compatible_tws = service_tw.select{ |tw|
               tw[:day_index].nil? || new_tw[:day_index].nil? || tw[:day_index] == new_tw[:day_index] &&
-                (tw[:start].nil? || new_tw[:end].nil? || tw[:start] <= new_tw[:end]) &&
-                (tw[:end].nil? || new_tw[:start].nil? || tw[:end] >= new_tw[:start])
+                (new_tw[:end].nil? || tw[:start] <= new_tw[:end]) &&
```
braktar, line 930:
```suggestion
                (new_tw.end.nil? || tw.start <= new_tw.end) &&
                (tw.end.nil? || tw.end >= new_tw.start)
```
Could you also edit the rest of the function ?

lib/heuristics/scheduling_heuristic.rb
```diff
@@ -602,7 +602,7 @@ def compute_tw_for_next(inserted_final_time, route_next, dist_from_inserted, cur
       setup_duration = dist_from_inserted.zero? ? 0 : @services_data[route_next[:id]][:setup_durations][route_next[:activity]]
       if !route_next[:tw].empty?
         tw = find_corresponding_timewindow(current_day, route_next[:arrival], @services_data[route_next[:id]][:tws_sets][route_next[:activity]], route_next[:end] - route_next[:arrival])
-        sooner_start = tw[:start] - dist_from_inserted - setup_duration if tw && tw[:start]
+        sooner_start = tw[:start] - dist_from_inserted - setup_duration if tw
```
braktar, line 605:
```suggestion
        sooner_start = tw.start - dist_from_inserted - setup_duration if tw
```

lib/heuristics/concerns/scheduling_data_initialisation.rb
```diff
@@ -248,8 +248,8 @@ def best_common_tw(set)
         group_tw.delete_if{ |tw1|
           data[:tws_sets].first.none?{ |tw2|
             (tw1[:day_index].nil? || tw2[:day_index].nil? || tw1[:day_index] == tw2[:day_index]) &&
-              (tw1[:start].nil? || tw2[:end].nil? || tw1[:start] <= tw2[:end]) &&
-              (tw1[:end].nil? || tw2[:start].nil? || tw1[:end] >= tw2[:start])
+              (tw2[:end].nil? || tw1[:start] <= tw2[:end]) &&
```
braktar, line 251:
```suggestion
              (tw2.end.nil? || tw1.start <= tw2.end) &&
```

braktar (changes requested):
As long as we are manupulating the active hash model, we should always prefer to use 
`tw.start` over `tw[:start]`

senhalil:
While we are setting `timewindow.start` to zero by default, why not make it so that `timewindow` has the default value `{ start: 0 }` as well. Then we don't need to use '&'s  (or worse `||` and ternary operators) everywhere. @braktar @fonsecadeline is it possible to do this without too much headache?

braktar:
As vehicle.timewindow is mutually exclusive with sequence_timewindows it will probably add extra checks

lib/heuristics/concerns/scheduling_data_initialisation.rb
fonsecadeline: these timewindows are not from model timewindow (this should be improved in another PR : https://gitlab.com/mapotempo/optimizer-api/-/issues/689)

braktar (approved):

optimizer_wrapper.rb
```diff
@@ -926,22 +929,24 @@ def self.zip_cluster(vrp, cluster_threshold, force_cluster)
...
+              compatible_end = compatible_tws.collect(&:end).compact.min
+              new_tw.start = [new_tw.start, compatible_start].max
+              new_tw.end = [new_tw.end, compatible_end].min if compatible_end
             end
           }
         end
       }
-      raise OptimizerWrapper::DiscordantProblemError, 'Zip cluster : no intersecting tw could be found' if !new_tws.empty? && (new_tws - to_remove_tws).empty?
+      if !new_tws.empty? && (new_tws - to_remove_tws).empty?
+        raise OptimizerWrapper::DiscordantProblemError.new('Zip cluster : no intersecting tw could be found')
```
senhalil, line 948: no need for space before colons `:`

senhalil (approved):

senhalil (changes requested):
we decided to have the default of `timewindow.start` only on the model side and leave the API to be "empty". we need to make sure that `nil` values (coming from the API) do not over-write the default `0` on the model side.

fonsecadeline:
> We said its better if we have the default of `timewindow.start` only on the model side and leave the API to be "empty".
> 
>     * [ ]  need to make sure that `nil` values (coming from the API) do not over-write the default `0` on the model side.

I missed part of the discussion. Why do we not want to assign default value within grape model (if I understood correctly) ?

senhalil:
My message was not clear sorry. While you are away, in one of the daily meetings, Fabien pointed out that for the API users the meaning of "no TW.start" should stay the same (i.e., no start value) and they don't need to know `TW.start=nil` means `TW.start=0`.

fonsecadeline:
Ok thanks for the update :) I will make corresponding changes.

fonsecadeline:
@senhalil I added a function regarding your point but I could not find a way to test `null` value within JSON file using mini-tests. All `null` values are converted to `nil` when parsing into JSON.

senhalil:
@fonsecadeline I feel like you are mixing JSON (text file) and its representation inside the code (Hash). So nulls becoming nil is normal.

```
irb(main):002:0> require 'json'
=> true
irb(main):003:0> {a: nil}.to_json
=> "{\"a\":null}"
irb(main):004:0> JSON.parse("{\"a\":null}")
=> {"a"=>nil}
irb(main):005:0> 
```

models/vrp.rb
```diff
@@ -477,6 +478,20 @@ def self.generate_schedule_indices_from_date(hash)
...
+        }
+      }
+
+      hash[:vehicles]&.each{ |v|
+        [v[:timewindow], v[:sequence_timewindows]].compact.flatten.each{ |tw|
+          tw[:start] ||= 0
+        }
+      }
+    end
+
```
senhalil, line 494: So the default start value of `model::timewindow` is not enough, is that it? :thinking: I am confused. When I asked "we need to make sure..." I wasn't expecting that we need such a function. Is the source of the problem, we let the users to send {"start": null} ?

test/models/vrp_test.rb
```diff
@@ -277,5 +277,25 @@ def test_available_interval
...
+
+    def test_timewindow_start_can_not_be_nil
+      vrp = VRP.basic
+      vrp[:services].first[:activity][:timewindows] = [{ end: 10 }]
+      created_vrp = TestHelper.create(vrp)
+      assert_equal 0, created_vrp.services.first.activity.timewindows.first.start
+
+      vrp = VRP.basic
+      vrp[:services].first[:activities] = [{ timewindows: [{ start: nil, end: 10 }] },
+                                           { timewindows: [{ start: nil, end: 10 }] }]
```
senhalil, line 289: maybe like this, to cover all possibilities
```suggestion
      vrp[:services].first[:activities] = [{ timewindows: [{ start: nil, end: 10 }] },
                                           { timewindows: [{ end: 10 }] }]
```

models/vrp.rb
senhalil: If that is the case I wonder if "allow_blank: false"  would  be a better solution :nerd_face:  lets talk about it in the daily tomorrow

fonsecadeline: Right, this might be better ! I did not think about it.

fonsecadeline: However, this might return an error for VRP that were previously accepted. As you said, let's talk about this during daily :)

fonsecadeline:
@senhalil more or less. I just don't know how to send a JSON file "{\"a\":null}" within tests structure.

fonsecadeline:
@senhalil I apparently did something wrong yesterday, it works now. (In fact I was expecting a raise, when error is just returned by POST). Thank you !

models/vrp.rb
braktar: We can't add allow_blank: false. The timewindows transmitted through a CSV without a start will always be in error

senhalil:
> @senhalil I apparently did something wrong yesterday, it works now. (In fact I was expecting a raise, when error is just returned by POST). Thank you !

Couldn't quite understand how I did but I am glad to help!

> @senhalil more or less. I just don't know how to send a JSON file "{"a":null}" within tests structure.

I was trying to explain `{a: nil}` in the code becomes `{"a": null}` when it is converter to JSON to make the curl (for example `submit_vrp` of request_helper.rb does that `post '/0.1/vrp/submit', params.to_json, 'CONTENT_TYPE' => 'application/json'`).

models/vrp.rb
senhalil: Before I forget all, we said it would be better to do all of this by modifying the `coerce_with`. It  will have an input argument `default=nil` to be used when the parameter is `nil`. It will be used for the parameter for which we don't want to use the default value logic of the grape entity directly and we cannot do `allow_blank: false`. Feel free to add if I missed any details.

fonsecadeline: In fact it does not seem possible to use coerce_with to define default values (https://github.com/ruby-grape/grape/pull/2040#issuecomment-615483919).

Either we accept that users are aware default start value is 0 and we can easily define `default: 0` within grape, or we do need this function.

fonsecadeline:
I need to test this.

wrappers/ortools.rb
```diff
@@ -318,7 +318,7 @@ def solve(vrp, job, thread_proc = nil, &block)
             [
               rest.timewindows.collect{ |tw|
                 [
-                  tw.start || -2**56,
+                  tw.start,
                   end: tw.end || 2**56,
```
braktar, line 322: Seems like the `end` key is unexpected

braktar (approved):

wrappers/ortools.rb
fonsecadeline: or ```start:``` is missing, I will check this!

models/types/schedule_type.rb
```diff
@@ -18,24 +18,25 @@
...
+        else
+          log 'error', level: :error
+          raise ArgumentError.new('Invalid Time value')
+        end
+      end
 
     if return_value&.negative?
       log 'error', level: :error
-      raise ArgumentError, 'Invalid Time value'
+      raise ArgumentError.new('Invalid Time value')
```
senhalil, line 39: We can be more precise here since we know the reason
```suggestion
      raise ArgumentError.new('Time value cannot be negative')
```

wrappers/wrapper.rb
```diff
@@ -223,8 +223,7 @@ def assert_no_distance_limitation(vrp)
     def assert_vehicle_tw_if_schedule(vrp)
       vrp.preprocessing_first_solution_strategy.to_a.first != 'periodic' ||
         vrp.vehicles.all?{ |vehicle|
-          vehicle.timewindow && (vehicle.timewindow.start || vehicle.timewindow.end) ||
-            vehicle.sequence_timewindows&.any?{ |tw| (tw.start || tw.end) }
+          vehicle.timewindow || vehicle.sequence_timewindows&.size&.positive?
```
senhalil, line 226:
```suggestion
          vehicle.timewindow || vehicle.sequence_timewindows&.any?
```

lib/filters.rb
```diff
@@ -175,7 +175,7 @@ def self.merge_timewindows(vrp)
 
       service.activity.timewindows.each{ |timewindow|
         unified_timewindows[timewindow.id] = {
-          start: timewindow.start && ((timewindow.day_index || 0) * 86400 + timewindow.start) || (0 + (timewindow.day_index || 0) * 86400),
+          start: (timewindow.day_index || 0) * 86400 + timewindow.start,
```
senhalil, line 178: I ask myself if we can do a similar thing for `(timewindow.day_index || 0)` as well or not?

senhalil (changes requested):

lib/filters.rb
fonsecadeline: I don't think we can because timewindow.day_index == nil can mean that timewindow is valid every day. 

Let's have : 
- vehicle1.timewindow = { start: 0, end: 1000, day_index: 0 } and vehicle2.timewindow = { start: 0, end: 1000, day_index: 2 },
- service.activity.timewindows = [{ start: 5, end: 10 }]

If we set day_index = 0 (by default) to the service then problem changes, it is like service can only take place on mondays which does not correspond to initial constraint.
 
If we do want to always have day_index for timewindows then in create we should duplicate timewindows with no day_index in order to have one timewindow per day_index. I do not think it is worth to do that.

lib/interpreters/periodic_visits.rb
```diff
@@ -74,7 +74,7 @@ def generate_timewindows(timewindows_set)
         if @have_day_index
           first_day = timewindow.day_index ? (@schedule_start..@schedule_end).find{ |day| day % 7 == timewindow.day_index } : @schedule_start
           (first_day..@schedule_end).step(timewindow.day_index ? 7 : 1).collect{ |day_index|
-            Models::Timewindow.new(start: (timewindow.start || 0) + day_index * 86400,
+            Models::Timewindow.new(start: timewindow.start + day_index * 86400,
                                    end: (timewindow.end || 86400) + day_index * 86400)
```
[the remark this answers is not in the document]
fonsecadeline: You are right. I do not see what could break if timewindow has default end.
And in fact I think even in general context we could set default timewindow to 86400. But you are right, in some specific cases this could be a problem.

If you are okay with that I will create a ticket for that so that we can merge this PR and think better about consequences of assigning default timewindow end.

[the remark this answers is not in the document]
senhalil: Sure ticket is alright, thanks Adeline ! You can give the the link to [this conversation](https://github.com/Mapotempo/optimizer-api/pull/143#discussion_r587367128) to have the context.

> And in fact I think even in general context we could set default timewindow to 86400. But you are right, in some specific cases this could be a problem.

Multi-day VRP is still general for me but in any case the "problem" is big enough (i.e., the user has to define TW.ends, even if they don't have any time-windows) that we cannot do it. Plus in the other simpler general cases, we might have some logic inside the wrappers that depend on `tw.end.nil?` at the moment (we have to check) but for the periodic case the initialiser will be :ok_hand: hopefully.

senhalil (approved):

PR #144, opened 2021-02-19
Relations and services point to each other

senhalil:
- Prevent uninitialized `has_many` objects (e.g. `vrp.services`, `vrp.relations`, `vehicle.rests`, etc.) returning a different empty array object with each call
- Instead of ids, use actual references inside the relations and vice versa

### DIFF
--- a/models/base.rb
+++ b/models/base.rb
@@ -53,7 +53,7 @@ module Models
       super
 
       define_method(name) do
-        self[name] || []
+        self[name] ||= []
       end
 
       define_method("#{name}=") do |vals|
--- a/models/concerns/expand_data.rb
+++ b/models/concerns/expand_data.rb
@@ -45,6 +45,14 @@ module ExpandData
     }
   end
 
+  def add_relation_references
+    self.relations.each{ |relation|
+      relation.linked_services.each{ |service|
+        service.relations << relation
+      }
+    }
+  end
+
   def add_sticky_vehicle_if_routes_and_partitions
     return if self.preprocessing_partitions.empty?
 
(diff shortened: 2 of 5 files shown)


braktar (approved):

senhalil:
Converted to `draft` because some tests are broken on Travis after I added the `vehicle` part as we discussed this morning. 

It looks like there are vehicle ids in the `relation[:linked_vehicle_ids]` list which does not exist. I am not sure if it is broken test logic or a bug due to the location of "expansion" of vehicle trips.. 

Gonna check next week.

senhalil:
Moved vehicle-relation references to #146 since it needs a change in the multi-tour expansion location but the relation-service references are good to go.

fonsecadeline:
@senhalil are you talking about multitour expansion from multitrips.expand ? I think we talked about moving this expand within vrp.create but I can not find related comments.

fonsecadeline (approved):

senhalil:
> @senhalil are you talking about multitour expansion from multitrips.expand ? I think we talked about moving this expand within vrp.create but I can not find related comments.

Yes, I was referring to  `MultiTrips::expand`. Maybe we talked it at the daily point, I don't remember. 

Once that function is moved to the beginning of vrp.create -- before the "vrp.send"s -- we can go ahead with the other PR (#146) at the moment creation of relations poses problem because the vehicles "don't exist" from the point of view of base.rb.

fonsecadeline:
I see, I will work on this expand move as soon as https://github.com/Mapotempo/optimizer-api/pull/142 and https://github.com/Mapotempo/optimizer-api/pull/143 are merged (prerequisits for multi tour dev)

PR #145, opened 2021-02-19
Support relations in split & partitions

senhalil:
Adds relations support to split_solve & partitions

LINKING_RELATIONS := Relations that link multiple services to be on the same route 
        (order, same_route, sequence, shipment) are supported in 
	split_solve algorithm (`max_split_size`) and
	partitions (`[:configuration][:preprocessing][:partitions]`)

FORCING_RELATIONS := Relations that force multiple services/vehicle to stay in the same VRP 
	(vehicle_trips, meetup, minimum_duration_lapse, maximum_duration_lapse, minimum_day_lapse, maximum_day_lapse)
	are respected by split_solve (`max_split_size`) algorithm

- [ ] Needs https://github.com/Mapotempo/balanced_vrp_clustering/pull/16 to be merged and Gemfile/Gemfile.lock corrected/updated

Closes 291, 544, 564, 634

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 - Corresponding vehicle_id is returned within each service's skills if problem is partitioned with vehicle entity [#110](https://github.com/Mapotempo/optimizer-api/pull/110)
 - Support initial routes and skills in split_solve (`max_split_size`) algorithm [#140](https://github.com/Mapotempo/optimizer-api/pull/140)
+- Support relations (`order`, `same_route`, `sequence`, `shipment`) in split_solve algorithm (`max_split_size`) and partitions (`[:configuration][:preprocessing][:partitions]`) [](https://github.com/Mapotempo/optimizer-api/pull/)
+- split_solve algorithm (`max_split_size`) respects relations (`vehicle_trips`, `meetup`, `minimum_duration_lapse`, `maximum_duration_lapse`, `minimum_day_lapse`, `maximum_day_lapse`) [](https://github.com/Mapotempo/optimizer-api/pull/)
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,8 +6,8 @@
 
 - Corresponding vehicle_id is returned within each service's skills if problem is partitioned with vehicle entity [#110](https://github.com/Mapotempo/optimizer-api/pull/110)
 - Support initial routes and skills in split_solve (`max_split_size`) algorithm [#140](https://github.com/Mapotempo/optimizer-api/pull/140)
-- Support relations (`order`, `same_route`, `sequence`, `shipment`) in split_solve algorithm (`max_split_size`) and partitions (`[:configuration][:preprocessing][:partitions]`) [](https://github.com/Mapotempo/optimizer-api/pull/)
-- split_solve algorithm (`max_split_size`) respects relations (`vehicle_trips`, `meetup`, `minimum_duration_lapse`, `maximum_duration_lapse`, `minimum_day_lapse`, `maximum_day_lapse`) [](https://github.com/Mapotempo/optimizer-api/pull/)
+- Support relations (`order`, `same_route`, `sequence`, `shipment`) in split_solve algorithm (`max_split_size`) and partitions (`[:configuration][:preprocessing][:partitions]`) [PR145](https://github.com/Mapotempo/optimizer-api/pull/145)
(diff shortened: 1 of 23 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
 
 - Corresponding vehicle_id is returned within each service's skills if problem is partitioned with vehicle entity [#110](https://github.com/Mapotempo/optimizer-api/pull/110)
 - Support initial routes and skills in split_solve (`max_split_size`) algorithm [#140](https://github.com/Mapotempo/optimizer-api/pull/140)
+- Support relations (`order`, `same_route`, `sequence`, `shipment`) in split_solve algorithm (`max_split_size`) and partitions (`[:configuration][:preprocessing][:partitions]`) [](https://github.com/Mapotempo/optimizer-api/pull/)
+- split_solve algorithm (`max_split_size`) respects relations (`vehicle_trips`, `meetup`, `minimum_duration_lapse`, `maximum_duration_lapse`, `minimum_day_lapse`, `maximum_day_lapse`) [](https://github.com/Mapotempo/optimizer-api/pull/)
```
senhalil, line 10:
```suggestion
- Support relations (`order`, `same_route`, `sequence`, `shipment`) in split_solve algorithm (`max_split_size`) and partitions (`[:configuration][:preprocessing][:partitions]`) [PR145](https://github.com/Mapotempo/optimizer-api/pull/145)
- split_solve algorithm (`max_split_size`) respects relations (`vehicle_trips`, `meetup`, `minimum_duration_lapse`, `maximum_duration_lapse`, `minimum_day_lapse`, `maximum_day_lapse`) [PR145](https://github.com/Mapotempo/optimizer-api/pull/145)
```

models/relation.rb
```diff
@@ -31,5 +31,10 @@ class Relation < Base
     # validates_numericality_of :lapse, allow_nil: true
     # validates_numericality_of :periodicity, greater_than_or_equal_to: 1
     # validates_inclusion_of :type, :in => %i(same_route sequence order minimum_day_lapse maximum_day_lapse shipment meetup maximum_duration_lapse vehicle_group_duration vehicle_group_duration_on_weeks vehicle_group_duration_on_months vehicle_trips)
+
+    def self.create(hash)
```
fonsecadeline, line 35: Could we eventually use a coerce in grape model instead ?

test/test_helper.rb
```diff
@@ -96,6 +96,8 @@ def self.coerce(vrp)
 
     vrp.provide_original_ids unless vrp.is_a?(Hash) # TODO: re-dump with this modification
 
+    vrp[:relations]&.each{ |r| r[:type] = r[:type]&.to_sym } # TODO: re-dump with this modification
```
fonsecadeline, line 99: Is this meant to be in the PR ? (I mean the TODO)

fonsecadeline (approved):

models/relation.rb
senhalil: We don't need a coerce in the grape model actually. It is already corrected on the API level (and Grape does the string -> symbol conversion automatically) this is here to make sure that the internal model respects the model defined by the API.

test/test_helper.rb
senhalil: kinda, I didn't want to re-dump everything (and add another 100+ Mb) for something this small :) This coerce takes care of the conversion of old dumps but eventually we would re-dump I thought -- unless we get rid of the dumps completely before :crossed_fingers:

fonsecadeline: I will need to redump in one of my next commits (multi_tour removes trip field within vehicles) so maybe I can dump it at this time, if this modification is on dev.

senhalil: Sure, cheers Adeline! 

By the way, if it is something we can easily "coerce" via the test_helper:coerce like here, then it is better to not to dump because of the size of the repo :(

fonsecadeline: I know, but in my case I think we need to. Because one field was removed so for instance when we try to _duplicate_safe_ we look for a key that does not exist anymore.

We could also ignore keys when parameter is nil, but I guess it is cleaner to dump everything again.

senhalil: Yeah, if a field is removed then I am afraid we have to redump the vrp; otherwise, the methods associated with the deleted field will not exist but the fields will still be in the vrp (which eventually means bugs).

I resolve this convo, feel free to unresolve it if you want me to change something.

lib/interpreters/split_clustering.rb
```diff
@@ -340,6 +354,26 @@ def self.create_representative_vrp(split_solve_data)
         }
       }
 
+      relations = []
+      # go thorugh the original relations and force the services and vehicles to stay in the same sub-vrp if necessary
```
braktar, line 358:
```suggestion
      # go through the original relations and force the services and vehicles to stay in the same sub-vrp if necessary
```

braktar (approved):

PR #146, opened 2021-02-25
Relations and vehicles reference each other

senhalil:
@fonsecadeline I had to remove the vehicle relation from #144 because it is not compatible with the actual multi tour expansion logic. Since the expansion of the `trip` parameter is inside the optimizerwrapper and uses Marshal.dump/load the `base.rb` cannot "see" these new vehicles and the trip relation generated  inside this expansion function cannot point to these vehicles. We need to move the expansion of the `vehicle[:trips]` before the create function so that inside the code we only maintain one way -- trips as relations.

### DIFF
--- a/models/concerns/expand_data.rb
+++ b/models/concerns/expand_data.rb
@@ -50,6 +50,9 @@ module ExpandData
       relation.linked_services.each{ |service|
         service.relations << relation
       }
+      relation.linked_vehicles.each{ |vehicle|
+        vehicle.relations << relation
+      }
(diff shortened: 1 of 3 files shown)


senhalil:
Moved vehicle-relation referencing to #146 since it needs a change in the multi-tour expansion logic but the relation-service references are good to go.

senhalil:
- [ ] needs https://github.com/Mapotempo/optimizer-api/pull/123

PR #147, opened 2021-02-26
Vroom expects a dummy skill

braktar:
https://github.com/VROOM-Project/vroom/issues/460

- [x] Fix the issue
- [x] Add Test (moved to another PR)

### DIFF
--- a/wrappers/vroom.rb
+++ b/wrappers/vroom.rb
@@ -260,7 +260,7 @@ module Wrappers
           id: index,
           location_index: @points[service.activity.point_id].matrix_index,
           service: service.activity.duration.to_i,
-          skills: service.skills.flat_map{ |skill| vrp_skills.find_index{ |sk| sk == skill } }.compact + # undefined skills are ignored
+          skills: [vrp_skills.size] + service.skills.flat_map{ |skill| vrp_skills.find_index{ |sk| sk == skill } }.compact + # undefined skills are ignored
             service.sticky_vehicles.flat_map{ |sticky| vrp_skills.find_index{ |sk| sk == sticky.id } }.compact,
           priority: (100 * (8 - service.priority).to_f / 8).to_i, # Scale from 0 to 100 (higher is more important)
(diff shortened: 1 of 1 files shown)


fonsecadeline (approved):

senhalil:
No worries, I just saw your message !

wrappers/vroom.rb
```diff
@@ -260,7 +260,7 @@ def collect_jobs(vrp, vrp_skills, vrp_units)
           id: index,
           location_index: @points[service.activity.point_id].matrix_index,
           service: service.activity.duration.to_i,
-          skills: service.skills.flat_map{ |skill| vrp_skills.find_index{ |sk| sk == skill } }.compact + # undefined skills are ignored
+          skills: [vrp_skills.size] + service.skills.flat_map{ |skill| vrp_skills.find_index{ |sk| sk == skill } }.compact + # undefined skills are ignored
```
senhalil, line 263: It seems dangerous to me to use the array size as a default skill because it is going to be relatively "normal" number. For example if there are 5 skills in total and if the user has sent skills as the vehicle numbers `1`,`2`, etc then we break their skills completely.

I suggest doing something like `defaullt_skill = Digest::MD5.hexdigest vrp_skills.join` and then we use `default_skill` down below. This way there is no risk of erasing a skill.

senhalil (changes requested):

wrappers/vroom.rb
braktar: The skills send by the users are decorrelated from the skills send to VROOM

We collect all the skills in an array and convert them into their index in this array.

The dummy skill is the maximum skill index + 1

senhalil: ah, I didn't pay enough attention to the fact that we have indices here! 

On a separate note, why do we convert the skills to indices, vroom doesn't like string?

senhalil (approved):

wrappers/vroom.rb
braktar: Yes, Vroom skills are integers : https://github.com/VROOM-Project/vroom/blob/master/docs/API.md#jobs

PR #149, opened 2021-03-01
detect unconsistency between dates and indices

fonsecadeline:


models/vrp.rb
```diff
@@ -407,15 +407,26 @@ def self.deduce_unavailable_days(hash, element, start_index, end_index, type)
       collect_unavaible_day_indices(element, start_index, end_index)
     end
 
-    def self.generate_schedule_indices_from_date(hash)
-      return hash if !hash[:configuration] || !hash[:configuration][:schedule]
-
-      if hash[:configuration][:schedule][:range_indices] &&
-         (hash[:services]&.any?{ |s| s[:unavailable_date_ranges] || s[:unavailable_visit_day_date] })
+    def self.detect_date_indices_unconsistency(hash)
```
senhalil, line 410: unconsistency -> inconsistency

models/vrp.rb
```diff
@@ -407,15 +407,26 @@ def self.deduce_unavailable_days(hash, element, start_index, end_index, type)
...
         raise OptimizerWrapper::DiscordantProblemError.new(
           'Date intervals are not compatible with schedule range indices'
         )
       end
+    end
+
+    def self.generate_schedule_indices_from_date(hash)
+      return hash if !hash[:configuration] || !hash[:configuration][:schedule]
+
+      detect_date_indices_unconsistency(hash)
```
senhalil, line 429: What if there is no `hash[:configuration][:schedule]` given -- i.e., we don't reach the `detect_date_indices_unconsistency` function --  and there are `date` in the vehicles and `indices` in the services, do we crash or refuse the VRP?

test/models/vrp_consistency_test.rb
```diff
@@ -242,5 +242,14 @@ def test_available_intervals_compatibility
       vrp = TestHelper.create(vrp) # this should not raise
       assert_equal (4..7).to_a, vrp.services.first.unavailable_days.sort
     end
+
+    def test_dates_mixed_with_indices
```
senhalil, line 246:
```suggestion
    def test_dates_cannot_be_mixed_with_indices
```

senhalil (changes requested):

models/vrp.rb
fonsecadeline: If no schedule is provided then those fields are ignored (within create and whole algorithm). There is no way this produces on error from our side. 

I do not think we should return an error in this case because that would make it easier for customer to go from optimization on one week vs on one day for instance.

fonsecadeline: Does that make sense for you too ?

senhalil: :100: thanks Adeline!

senhalil: Brainstorming: I wonder if it makes sense to keep these fields -- e.g., `:unavailable_date_ranges`, indices etc. -- inside the internal model (vehicles, services) if there is no `schedule` :/ Can you imagine any use for these fields if there is no `hash[:configuration][:schedule]`

fonsecadeline: Indeed, I don't see who it could be used without schedule. 

But wouldn't it cost more to remove those fields if no schedule is provided than just ignoring them ? In fact, if no schedule is provided then we do not fill _unavailable_days_ which is new internal structure that groups all of these API fields. Therefore, in our internal model it corresponds to not having those fields.

senhalil: hash.delete(key) is a constant time `O(1)` operation so it would amount to nothing but if I understood it correctly these (date_ranges, indicies) don't "enter" the internal VRP model (is that it?).. So even if there is no `[:configuration][:schedule]` then effectively there is no risk to the internal model. 

I was afraid that these fields would appear inside vrp.vehicles/services :facepalm: but that is not the case! The main risk is in the verification process inside `vrp.create` itself -- which is not much.. Cheers Adeline!

models/vrp.rb
```diff
@@ -407,15 +407,26 @@ def self.deduce_unavailable_days(hash, element, start_index, end_index, type)
...
+    def self.detect_date_indices_inconsistency(hash)
+      missions = hash[:services].to_a + hash[:shipments].to_a
+      services_with_date = missions.any?{ |m| m[:unavailable_date_ranges] || m[:unavailable_work_date] }
+      services_with_index = missions.any?{ |m| m[:unavailable_index_ranges] || m[:unavailable_work_day_indices] }
+      vehicles_with_date = hash[:vehicles]&.any?{ |v| v[:unavailable_work_date] || v[:unavailable_date_ranges] }
+      vehicles_with_index = hash[:vehicles]&.any?{ |v| v[:unavailable_work_day_indices] || v[:unavailable_index_ranges] }
+      if (hash[:configuration][:schedule][:range_indices] &&
+         (services_with_date || vehicles_with_date)) ||
+         (hash[:configuration][:schedule][:range_date] &&
+         (services_with_index || vehicles_with_index))
```
senhalil, line 419: Maybe we can simplify the logic as follows:

```suggestion
      missions_and_vehicles = hash[:services].to_a + hash[:shipments].to_a + hash[:vehicles].to_a
      has_date = missions_and_vehicles.any?{ |m| m[:unavailable_date_ranges] || m[:unavailable_work_date] }
      has_index = missions_and_vehicles.any?{ |m| m[:unavailable_index_ranges] || m[:unavailable_work_day_indices] }
      if (hash[:configuration][:schedule][:range_indices] && has_date) ||
         (hash[:configuration][:schedule][:range_date] && has_index)
```

models/vrp.rb
```diff
@@ -407,15 +407,26 @@ def self.deduce_unavailable_days(hash, element, start_index, end_index, type)
       collect_unavaible_day_indices(element, start_index, end_index)
     end
 
-    def self.generate_schedule_indices_from_date(hash)
-      return hash if !hash[:configuration] || !hash[:configuration][:schedule]
-
-      if hash[:configuration][:schedule][:range_indices] &&
-         (hash[:services]&.any?{ |s| s[:unavailable_date_ranges] || s[:unavailable_visit_day_date] })
+    def self.detect_date_indices_inconsistency(hash)
```
senhalil, line 410: Lets make this function handle its corner case so that there is no future risk of `undefined method [] for nil:NilClass`

```suggestion
    def self.detect_date_indices_inconsistency(hash)
      return hash unless hash[:configuration] && hash[:configuration][:schedule]

```

senhalil (approved):

models/vrp.rb
fonsecadeline: I agree, even though this check is done in _generate_schedule_indices_from_date_ which is the only function that calls _detect_date_indices_inconsistency_. Should we check this twice ?

braktar (approved):

models/vrp.rb
senhalil: I think it is better if each function handles its corner case on its own. this fuction cannot know if the caller already checked or not.

PR #151, opened 2021-03-03
fix duplicate_safe function

fonsecadeline:
Latter modifications (https://github.com/Mapotempo/optimizer-api/pull/139) were not consistent but we still need some cleaning then.


senhalil (approved):
:ok_hand:  This was the modification I was suggesting (i.e., we shouldn't be needing to subtract stuff from the keys) in https://github.com/Mapotempo/optimizer-api/pull/139#discussion_r576327200

braktar (approved):

PR #32, opened 2020-06-18
CSV output columns consistent with provided language

fonsecadeline:

PR #153, opened 2021-03-05
Use present with VrpResult

fonsecadeline:
Some points need to be discussed : 

- when we collect with _deep_symbolize_keys_ we (kind of) duplicate object. Is that acceptable ? Do we want to have symbols from the start instead of symbolizing here ?
- we now output all keys, even if field is null. This can be confusing, for instance in the case of returning all _service_id_, _shipment_id_ and _rest_id_ for one single activity (with only one of these which is not nil). Still, this is JSON format so this should be an impact on our customers' process. Indeed, _rest_id_ will still be nil, even if key is present.


fonsecadeline:
Prerequisit for https://gitlab.com/mapotempo/optimizer-api/-/issues/635 (we want a field for each service to specify its corresponding cluster caracteristics).

api/v01/entities/vrp_result.rb
```diff
@@ -26,6 +26,7 @@ class VrpResultSolutionRouteActivityDetailTimewindows < Grape::Entity
 
     class VrpResultDetailQuantities < Grape::Entity
       expose :unit, documentation: { type: String, desc: '' }
+      expose :label, documentation: { type: String, desc: '' }
```
senhalil, line 29: what is unit? Isn't that the label of the unit? Or is `unit` the id? :thinking:

braktar:
Keys are not already all symbols ? Because, it should. Anyway, `deep_symbolize_keys` is temporary until we have a proper Solution model.

I have no particular point of view on the fact to return null values instead of missing fields. The fields are documented anyway, so the clients expect to receive them.

api/v01/entities/vrp_result.rb
```diff
@@ -45,16 +46,18 @@ class VrpResultSolutionRouteActivityDetails < Grape::Entity
...
+      expose :time, documentation: { type: Float, desc: 'Cost associated to the time dimension' }
+      expose :distance, documentation: { type: Float, desc: 'Cost associated to the distance dimension' }
+      expose :value, documentation: { type: Float, desc: 'Cost associated to the value dimension' }
+      expose :lateness, documentation: { type: Float, desc: 'Cost associated to late arrival' }
+      expose :overload, documentation: { type: Float, desc: 'Cost associated to quantities overload' }
     end
 
     class VrpResultSolutionRouteActivities < Grape::Entity
+      expose :day_week_num, documentation: { type: String, desc: '' }
+      expose :day_week, documentation: { type: String, desc: '' }
```
senhalil, line 60: Do we need both `day_week_num` and `day_week`?

braktar:
To avoid breaking change on client side with null values, it seems we can rely on `expose_nil` option https://github.com/ruby-grape/grape-entity#expose-nil

api/v01/entities/vrp_result.rb
```diff
@@ -104,6 +112,8 @@ class VrpResultSolution < Grape::Entity
       expose :iterations, documentation: { type: Integer, desc: 'Total number of iteration performed to obtain the current result' }
       expose :total_distance, documentation: { type: Integer, desc: 'cumulated distance of every route' }
       expose :total_time, documentation: { type: Integer, desc: 'Cumulated time of every route' }
+      expose :total_travel_time, documentation: { type: 'Travel time within the route'}
+      expose :total_waiting_time, documentation: { type: 'Idle time within the route' }
```
senhalil, line 116: I think these are not "within the route" but the "overall" or "total"

api/v01/vrp.rb
```diff
@@ -112,15 +112,14 @@ class Vrp < APIBase
...
                 status 201
-                present({ job: { id: ret, status: :queued }}, with: Grape::Presenters::Presenter)
+                present({ job: { id: ret, status: :queued }}, with: VrpResult)
               elsif ret.is_a?(Hash)
                 status 200
                 if vrp.restitution_csv
                   present(OptimizerWrapper.build_csv(ret.deep_stringify_keys), type: CSV)
                 else
-                  present({ solutions: [ret], job: { status: :completed }}, with: Grape::Presenters::Presenter)
+                  present({ solutions: [ret&.deep_symbolize_keys], job: { status: :completed }}, with: VrpResult)
```
senhalil, line 122: I belive this was your first question @fonsecadeline I think it is better if we fix this issue from the source so that the result hash always uses symbols in the first place instead of `deep_symbolize_keys`. This hash is one of the last hashes that is in the old style and it is just annoying to swithc to symbol/string in the middle of the code.

senhalil (changes requested):
I am not sure if we should return the `null`s because if a key doesn't exist it is `null` by default. From the user's point of view it should be the same. 

I don't know what other API's do but I suspect they don't return all keys that are `null` and only return the parameters that apply.

@fab-girard do you know what is the best practice here? Do we return the `null`s or let grape do its thing and return only the keys with values?

api/v01/entities/vrp_result.rb
braktar: That is a legacy typo from me :D `unit` should be `id`

api/v01/entities/vrp_result.rb
fonsecadeline: I think we do. They are both needed in CSV file to import in mapotempo. Today initial generates this CSV file from JSON file returned by API (which contains those fields).

fonsecadeline:
Just for you to know, I am planning to convert string keys into symbols within _build_csv_ function in https://github.com/Mapotempo/optimizer-api/pull/32. That is why I am temporarily using stringify_keys here.

fonsecadeline:
We decided that we can return key even if field is nil, unless there are 'conflicting' keys such as _shipment_id_ and _rest_id_ are 'conflicting' with _service_id_. In this case only one of these keys should be exposed in result.

senhalil (changes requested):
I think it is better if we avoid `deep_stringfy` and there might be more places to fix.
@fonsecadeline if there is no broken tests because of the missing `result[string/symbol]` replacements, can you add a test against that as well please?

api/v01/vrp.rb
```diff
@@ -217,27 +216,27 @@ class Vrp < APIBase
               solution = OptimizerWrapper::Result.get(id)
               status 202
               if solution && !solution.empty?
-                output_format = params[:format]&.to_sym || (solution['csv'] ? :csv : env['api.format'])
+                output_format = params[:format]&.to_sym || (solution[:csv] ? :csv : env['api.format'])
                 if output_format == :csv
-                  present(OptimizerWrapper.build_csv(solution['result']), type: CSV)
+                  present(OptimizerWrapper.build_csv(solution[:result]&.collect{ |r| r.deep_stringify_keys }), type: CSV) # TODO : edit build_csv function to consider symbolized keys
```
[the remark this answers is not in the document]
fonsecadeline: It is not very, but I might need to do it twice because of https://github.com/Mapotempo/optimizer-api/pull/32 (this PR moves build_csv logic to another file and clean functions).

But if this increases our memory issues then it may be worth to do it twice to avoid any problem !

api/v01/entities/vrp_result.rb
```diff
@@ -45,16 +46,18 @@ class VrpResultSolutionRouteActivityDetails < Grape::Entity
...
-      expose :value, documentation: { type: String, desc: 'Cost associated to the value dimension' }
-      expose :lateness, documentation: { type: String, desc: 'Cost associated to late arrival' }
-      expose :overload, documentation: { type: String, desc: 'Cost associated to quantities overload' }
+      expose :total, documentation: { type: Float, desc: 'Cumulated cost' }
+      expose :fixed, documentation: { type: Float, desc: 'Cost associated to the use of the vehicle' }
+      expose :time, documentation: { type: Float, desc: 'Cost associated to the time dimension' }
+      expose :distance, documentation: { type: Float, desc: 'Cost associated to the distance dimension' }
+      expose :value, documentation: { type: Float, desc: 'Cost associated to the value dimension' }
+      expose :lateness, documentation: { type: Float, desc: 'Cost associated to late arrival' }
+      expose :overload, documentation: { type: Float, desc: 'Cost associated to quantities overload' }
```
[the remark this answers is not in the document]
fonsecadeline: In fact https://github.com/Mapotempo/optimizer-api/pull/102 only rounds fields that were already specified as integers. Those fields are string so they are not rounded in that PR.

This PR might be merged before https://github.com/Mapotempo/optimizer-api/pull/102.
When it is done I can just round those fields and edit this documentation on https://github.com/Mapotempo/optimizer-api/pull/102 before we merge it.

api/v01/entities/vrp_result.rb
```diff
@@ -63,9 +66,13 @@ class VrpResultSolutionRouteActivities < Grape::Entity
       expose :begin_time, documentation: { type: Integer, desc: 'Time visit starts' }
       expose :end_time, documentation: { type: Integer, desc: 'Time visit ends' }
       expose :departure_time, documentation: { type: Integer, desc: '' }
-      expose :service_id, documentation: { type: String, desc: 'Internal reference of the service' }
-      expose :pickup_shipment_id, documentation: { type: String, desc: 'Internal reference of the shipment' }
-      expose :delivery_shipment_id, documentation: { type: String, desc: 'Internal reference of the shipment' }
+      expose :service_id, expose_nil: false, documentation: { type: String, desc: 'Internal reference of the service' }
+      # expose :original_service_id, expose_nil: false, documentation: { type: String, desc: 'Original reference of the service' } # TODO : rename to service_id and output when internal service id renamed (https://gitlab.com/mapotempo/optimizer-api/-/issues/537)
```
braktar, line 70: The next change of the API should be the v2. Which will merge all 'id' fields into a single one and one `original_id`

We should only mention `original_id` instead of o_service_id, o_shipment_id, o_rest_id and o_vehicle_id

api/v01/entities/vrp_result.rb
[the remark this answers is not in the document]
fonsecadeline: In fact I should just not change this documentation and fix it properly in PR102.

braktar (changes requested):

util/job_manager.rb
```diff
@@ -121,9 +121,10 @@ def self.set(key, value)
 
     def self.get(key)
       result = OptimizerWrapper::REDIS.get(key)
-      if result
-        JSON.parse(result.force_encoding(Encoding::UTF_8)) # On some env string is encoded as ASCII
-      end
+
+      return unless result
+
+      JSON.parse(result.force_encoding(Encoding::UTF_8), symbolize_names: true)
```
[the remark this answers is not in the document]
fonsecadeline: 1) In the "synchrone" case we do not need to GET result, it is returned by _OptimizerWrapper.wrapper_vrp_ which returns an hash with symbol keys. But you are right, some places need to be fixed in this file.

2) You are right, I missed those (don't really know how... :/)

api/v01/entities/vrp_result.rb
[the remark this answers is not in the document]
senhalil: I think it is better to give the correct type here in the first place. Also (you should test it but) i suspect grape will handle the rounding automatically, in that case we don't even need PR102. Plus since the rounding is done at the endpoint, we don't have to deal with all the issues I mentioned in PR102 (i.e., accumulating rounding errors)

util/job_manager.rb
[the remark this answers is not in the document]
senhalil: If there is no test, can you do a TDD?

api/v01/entities/vrp_result.rb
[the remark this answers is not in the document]
braktar: The costs are float, time and distance are integers.
With a cost per 100km of 6€ we should have a cost per meter of 0.00006€.
For a route of 56 km, the route has a cost 3.36€. Even if within the API we perform some scale change, the output should match the input expectation

api/v01/vrp.rb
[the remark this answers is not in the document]
senhalil: Hmm you are right there is quite a change in #32 and what is done here needs to be effectively duplicated there too :( 

I don't know what's the best but my gut feeling tells me that since replacing all string hash keys with symbol in the code is not that time consuming, it might worth to do it the right away.

api/v01/entities/vrp_result.rb
senhalil: We can even do just `id`, no?

api/v01/entities/vrp_result.rb
```diff
@@ -104,6 +113,8 @@ class VrpResultSolution < Grape::Entity
       expose :iterations, documentation: { type: Integer, desc: 'Total number of iteration performed to obtain the current result' }
       expose :total_distance, documentation: { type: Integer, desc: 'cumulated distance of every route' }
       expose :total_time, documentation: { type: Integer, desc: 'Cumulated time of every route' }
+      expose :total_travel_time, documentation: { type: 'Cumulated travel time of every route'}
+      expose :total_waiting_time, documentation: { type: 'Cumulated idle time of every route' }
```
senhalil, line 117:
```suggestion
      expose :total_travel_time, documentation: { type: Integer, desc: 'Cumulated travel time of every route'}
      expose :total_waiting_time, documentation: { type: Integer, desc: 'Cumulated idle time of every route' }
```

api/v01/entities/vrp_result.rb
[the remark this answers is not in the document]
senhalil: You are right I was confused the costs and stats ;) All costs effectively stay Float!

util/job_manager.rb
```diff
@@ -58,22 +58,22 @@ def perform
...
-          p = Result.get(self.uuid) || { 'graph' => [] }
-          p['graph'] = [] if !p.has_key?('graph')
-          p['csv'] = true if ask_restitution_csv
-          p['graph'] << { iteration: avancement, cost: cost, time: time }
-          p['result'] = solution if solution
+          p = Result.get(self.uuid) || { graph: [] }
+          p[:graph] ||= []
+          p[:csv] = true if ask_restitution_csv
+          p[:graph] << { iteration: avancement, cost: cost, time: time }
+          p[:result] = solution
```
braktar, line 65: The solution is not always transmitted if we have an intermediate cost. We should not override the result in this case

braktar (changes requested):

util/job_manager.rb
fonsecadeline: My mistake :) I thought p[:result] could not be different than nil if solution was nil.

optimizer_wrapper.rb
```diff
@@ -254,16 +254,14 @@ def self.solve(service_vrp, job = nil, block = nil)
...
               current_result[:csv] = true if cliqued_vrp[:restitution_csv]
-
-              current_result['pids'] = pids
+              current_result[:pids] = pids
 
               Result.set(job, current_result)
             }
           ) { |wrapper, avancement, total, _message, cost, _time, solution|
-            block&.call(wrapper, avancement, total, 'run optimization, iterations', cost, (Time.now - time_start) * 1000, solution.class.name == 'Hash' && solution) if dicho_level.nil? || dicho_level.zero?
+            block&.call(wrapper, avancement, total, 'run optimization, iterations', cost, (Time.now - time_start) * 1000, solution.class.name == 'Hash' ? solution : nil) if dicho_level.nil? || dicho_level.zero?
```
senhalil, line 264: Do we need this modification? Previous code is shorter and does the same thing.

util/job_manager.rb
```diff
@@ -58,22 +58,22 @@ def perform
         @wrapper = wrapper
         log "avancement: #{message} #{avancement}"
         if avancement && cost
-          p = Result.get(self.uuid) || { 'graph' => [] }
-          p['graph'] = [] if !p.has_key?('graph')
-          p['csv'] = true if ask_restitution_csv
-          p['graph'] << { iteration: avancement, cost: cost, time: time }
-          p['result'] = solution if solution
+          p = Result.get(self.uuid) || { graph: [] }
+          p[:graph] ||= []
```
senhalil, line 62: If for some reason, in other part of the code if we ever do `p[:graph] = nil` it will be overwritten here. I think the previous way was more secure -- or an improved version of it `unless p.key?(:graph)`.

util/job_manager.rb
```diff
@@ -58,22 +58,22 @@ def perform
...
       # Add values related to the current solve status
       p = Result.get(self.uuid) || {}
-      p['csv'] = true if ask_restitution_csv
-      p['result'] = result
-      if services_vrps.size == 1 && p && !p['result'].empty? && p['graph'] && !p['graph'].empty?
-        p['result'].first['iterations'] = p['graph'].last['iteration']
-        p['result'].first['elapsed'] = p['graph'].last['time']
+      p[:csv] = true if ask_restitution_csv
+      p[:result] = result
+      if services_vrps.size == 1 && p && !p[:result].empty? && p[:graph] && !p[:graph].empty?
```
senhalil, line 74: we can simplify this as follows I think
```suggestion
      if services_vrps.size == 1 && p && p[:result].any? && p[:graph]&.any?
```

senhalil (changes requested):

optimizer_wrapper.rb
fonsecadeline: In fact it does not do the same thing. This sometimes returned 'false' which produced an error when GET was called. 
In the case where solution.class.name == NilClass and solution is nil this returned false.

util/job_manager.rb
fonsecadeline: I just realized : 

if `p[:graph] = nil` then we will generate and error 2 lines after, when trying to add current iteration. I guess the aim of this `p['graph'] = [] if !p.has_key?('graph')` is to ensure `p['graph']` is never nil. 

In this case my edit would be ok, what do you think ?

braktar: Seems ok to me

braktar (approved):

senhalil (approved):

.rubocop.yml
```diff
@@ -10,6 +10,8 @@ Layout/SpaceBeforeBlockBraces:
 
 Layout/LineLength:
   Max: 120
+  Exclude:
+    - 'api/v01/entities/vrp_result.rb'
```
fab-girard, line 14: I don't think we can exclude some files like that

api/v01/entities/vrp_result.rb
```diff
@@ -63,16 +66,19 @@ class VrpResultSolutionRouteActivities < Grape::Entity
       expose :begin_time, documentation: { type: Integer, desc: 'Time visit starts' }
       expose :end_time, documentation: { type: Integer, desc: 'Time visit ends' }
       expose :departure_time, documentation: { type: Integer, desc: '' }
-      expose :service_id, documentation: { type: String, desc: 'Internal reference of the service' }
-      expose :pickup_shipment_id, documentation: { type: String, desc: 'Internal reference of the shipment' }
-      expose :delivery_shipment_id, documentation: { type: String, desc: 'Internal reference of the shipment' }
+      # TODO : v2 will provide an :id which will be equivalent to ID sent by used and :internal_id which will be equivalent to our current service_id/rest_id/pickup_shipment_id/delivery_shipment_id (https://gitlab.com/mapotempo/optimizer-api/-/issues/537)
```
fab-girard, line 69: no link to private project

api/v01/entities/vrp_result.rb
```diff
@@ -63,16 +66,19 @@ class VrpResultSolutionRouteActivities < Grape::Entity
...
+      expose :delivery_shipment_id, expose_nil: false, documentation: { type: String, desc: 'Internal reference of the shipment' }
+      expose :rest_id, expose_nil: false, documentation: { type: String, desc: 'Internal reference of the rest' }
       expose :detail, using: VrpResultSolutionRouteActivityDetails, documentation: { desc: '' }
       expose :type, documentation: { type: String, desc: 'depot, rest, service, pickup or delivery' }
       expose :current_distance, documentation: { type: Integer, desc: 'Travel distance from route start to current point' }
       expose :alternative, documentation: { type: Integer, desc: 'When one service has alternative activities, index of the chosen one' }
     end
 
     class VrpResultSolutionRoute < Grape::Entity
+      # TODO : v2 will provide an :vehicle_id which will be equivalent to ID sent by used + :internal_vehicle_id which will be equivalent to our current vehicle_id (https://gitlab.com/mapotempo/optimizer-api/-/issues/537)
```
fab-girard, line 81: idem

util/job_manager.rb
```diff
@@ -121,9 +121,10 @@ def self.set(key, value)
 
     def self.get(key)
       result = OptimizerWrapper::REDIS.get(key)
-      if result
-        JSON.parse(result.force_encoding(Encoding::UTF_8)) # On some env string is encoded as ASCII
-      end
+
```
fab-girard, line 124: you can keep empty line only after `return`

PR #156, opened 2021-03-10
ensure minimum_lapse smaller

fonsecadeline:
I don't think we can do this from grape side. Because we would need to use _given_ but then we should suppose that _maximum_lapse_ can only be provided when _minimum_lapse_ is provided, which is not the case. 

Or we should create several cases, but I think it is easier to do it within check_consistency function.
What do you think ?

### DIFF
--- a/models/vrp.rb
+++ b/models/vrp.rb
@@ -138,6 +138,9 @@ module Models
     end
 
     def self.check_consistency(hash)
+      hash[:services] ||= []
+      hash[:shipments] ||= []
+
       # vehicle time cost consistency
       if hash[:vehicles]&.any?{ |v| v[:cost_waiting_time_multiplier].to_f > (v[:cost_time_multiplier] || 1) }
         raise OptimizerWrapper::DiscordantProblemError, 'cost_waiting_time_multiplier cannot be greater than cost_time_multiplier'
@@ -174,9 +177,14 @@ module Models
         raise OptimizerWrapper::DiscordantProblemError, 'All matrices should have at least maximum(point[:matrix_index]) number of rows and columns' if matrix_not_big_enough
       end
 
+      # services consistency
+      if (hash[:services] + hash[:shipments]).any?{ |s| (s[:minimum_lapse] || 0) > (s[:maximum_lapse] || 2**32) }
+        raise OptimizerWrapper::DiscordantProblemError.new('Minimum lapse can not be bigger than maximum lapse')
+      end
+
       # shipment position consistency
       forbidden_position_pairs = [[:always_middle, :always_first], [:always_last, :always_middle], [:always_last, :always_first]]
-      hash[:shipments]&.each{ |shipment|
+      hash[:shipments].each{ |shipment|
         raise OptimizerWrapper::DiscordantProblemError, 'Unconsistent positions in shipments.' if forbidden_position_pairs.include?([shipment[:pickup][:position], shipment[:delivery][:position]])
       }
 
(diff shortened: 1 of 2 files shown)


models/vrp.rb
```diff
@@ -174,6 +174,11 @@ def self.check_consistency(hash)
         raise OptimizerWrapper::DiscordantProblemError, 'All matrices should have at least maximum(point[:matrix_index]) number of rows and columns' if matrix_not_big_enough
       end
 
+      # services consistency
+      if [hash[:services], hash[:shipments]].compact.flatten.any?{ |s| s[:minimum_lapse] && s[:maximum_lapse] && s[:minimum_lapse] > s[:maximum_lapse] }
```
braktar, line 178: Can't we allow minimum and maximum lapse to be equal ?
```suggestion
      if [hash[:services], hash[:shipments]].compact.flatten.any?{ |s| s[:minimum_lapse] && s[:maximum_lapse] && s[:minimum_lapse] >= s[:maximum_lapse] }
```

braktar (changes requested):

braktar:
As `given` method implies huge performance drop. It is indeed to put part of this logic out of grape

test/models/vrp_consistency_test.rb
```diff
@@ -311,5 +311,26 @@ def test_dates_cannot_be_mixed_with_indices
...
+      vrp[:services].first[:maximum_lapse] = 7
+      raised = false
+      begin
+        Models::Vrp.check_consistency(vrp)
+      rescue OptimizerWrapper::DiscordantProblemError => e
+        raised = true
+        assert_match 'Maximum lapse should always be bigger than minimum lapse', e.message
+      end
+
+      assert raised, 'This VRP is not consistent, it should raise an error.'
```
senhalil, line 333:
```suggestion
      error = assert_raises DiscordantProblemError do
        Models::Vrp.check_consistency(vrp)
      end
      assert_equal 'Maximum lapse should always be bigger than minimum lapse', error.message
```
assert_raises returns the error
assert_match is a regexp match and because if that its fail message is more complex

For example 
`assert_equal 'Centroid 0 is initialised with an incompatible service -- 1', error.message` returns
```
 FAIL["test_skills_taken_into_account", #<Minitest::Reporters::Suite:0x000055e8b06549f8 @name="SplitClusteringTest">, 1.7672128429403529]
 test_skills_taken_into_account#SplitClusteringTest (1.77s)
        --- expected
        +++ actual
        @@ -1 +1 @@
        -"Centroid 0 is initialised with an incompatible service -- 1"
        +"Centroid 0 is initialised with an incompatible service -- 0"
        /home/halil/dev/optimizer-api/test/lib/interpreters/split_clustering_test.rb:345:in `test_skills_taken_into_account'
```
while it returns the following with `assert_match`
```
 FAIL["test_skills_taken_into_account", #<Minitest::Reporters::Suite:0x0000562af0b90f98 @name="SplitClusteringTest">, 1.6732092190068215]
 test_skills_taken_into_account#SplitClusteringTest (1.67s)
        Expected /Centroid\ 0\ is\ initialised\ with\ an\ incompatible\ service\ \-\-\ 1/ to match "Centroid 0 is initialised with an incompatible service -- 0".
        /home/halil/dev/optimizer-api/test/lib/interpreters/split_clustering_test.rb:345:in `test_skills_taken_into_account'
```

senhalil (changes requested):

models/vrp.rb
fonsecadeline: We can not : most of our customers using scheduling use _minimum_lapse_ = _maximum_lapse_ . This means that lapse should be exactly equal to this value. Otherwise lapse can be infinite.

braktar (approved):

test/models/vrp_consistency_test.rb
senhalil: I think it is better to accept the suggestions instead of copy/pasting because it preservers code ownership. Unless the suggestion needs modification or if it is not a working code...

models/vrp.rb
```diff
@@ -174,17 +177,22 @@ def self.check_consistency(hash)
...
 
+      # services consistency
+      if (hash[:services] + hash[:shipments]).any?{ |s| (s[:minimum_lapse] || 0) > (s[:maximum_lapse] || 2**32) }
+        raise OptimizerWrapper::DiscordantProblemError.new('Minimum lapse can not be bigger than maximum lapse')
+      end
+
       # shipment position consistency
       forbidden_position_pairs = [[:always_middle, :always_first], [:always_last, :always_middle], [:always_last, :always_first]]
-      hash[:shipments]&.each{ |shipment|
+      hash[:shipments].each{ |shipment|
```
senhalil, line 187: what about the shipments that are given with relations? hmm this "duality" of  model:shipment vs relation:shipment is more problematic than we realised I think.

senhalil (approved):
Some checks are only implement vrp.shipments in mind and needs to be extended to relation.shipments.

PR #157, opened 2021-03-12
Sticky managed with skills

fonsecadeline:


fonsecadeline:
We finally did not need to rename original_skills/skills in VrpResult because present skills come from _build_detail_ function which is able to provide required skills. Providing "right" skills (original skills + clustering skills) through this function allows us to ensure we provide "right" skills in CSV format too. 

For record, we could have renamed fields from VrpResult with _as:_ option.

PR #158, opened 2021-03-15
Hot fix : sticky vehicles + clustering

fonsecadeline:
…sticky_vehicles

### DIFF
--- a/lib/interpreters/split_clustering.rb
+++ b/lib/interpreters/split_clustering.rb
@@ -792,7 +792,9 @@ module Interpreters
 
     def self.duplicate_vehicle(vehicle, timewindow, schedule)
       if timewindow.nil?
-        (schedule[:start]..[schedule[:end], 6].min).collect{ |day|
+        (0..6).flat_map{ |day|
+          next unless (schedule[:start]..schedule[:end]).any?{ |day_index| day_index % 7 == day }
+
           vehicle.skills = vehicle.skills.collect{ |sk_set|
             sk_set | [%w[mon tue wed thu fri sat sun][day]]
           }
--- a/lib/interpreters/split_clustering.rb
+++ b/lib/interpreters/split_clustering.rb
@@ -793,6 +793,9 @@ module Interpreters
     def self.duplicate_vehicle(vehicle, timewindow, schedule)
       if timewindow.nil?
         (schedule[:start]..[schedule[:end], 6].min).collect{ |day|
+          vehicle.skills = vehicle.skills.collect{ |sk_set|
+            sk_set | [%w[mon tue wed thu fri sat sun][day]]
+          }
           new_vehicle = Marshal.load(Marshal.dump(vehicle))
           new_vehicle.timewindow = Models::Timewindow.new(day_index: day)
           new_vehicle
@@ -800,6 +803,9 @@ module Interpreters
       elsif timewindow.day_index
         return [] unless (schedule[:start]..schedule[:end]).any?{ |day_index| day_index % 7 == timewindow.day_index }
 
+        vehicle.skills = vehicle.skills.collect{ |sk_set|
+          sk_set | [%w[mon tue wed thu fri sat sun][timewindow.day_index]]
+        }
         new_vehicle = Marshal.load(Marshal.dump(vehicle))
         new_vehicle.timewindow = timewindow
         new_vehicle.sequence_timewindows = nil
@@ -810,6 +816,9 @@ module Interpreters
 
           tw = Marshal.load(Marshal.dump(timewindow))
           tw.day_index = day
+          vehicle.skills = vehicle.skills.collect{ |sk_set|
+            sk_set | [%w[mon tue wed thu fri sat sun][day]]
+          }
           new_vehicle = Marshal.load(Marshal.dump(vehicle))
           new_vehicle.timewindow = tw
           new_vehicle.sequence_timewindows = nil
@@ -1225,7 +1234,6 @@ module Interpreters
           vrp.services.each{ |service|
             service.skills.insert(vehicle_id_in_skills ? 1 : 0, day_skill)
           }
-          vrp.vehicles.first.skills.first << day_skill
         end
 
         vrp
--- a/models/vrp.rb
+++ b/models/vrp.rb
@@ -219,10 +219,19 @@ module Models
         end
       end
 
-      # number of visits consistency
-      if !configuration[:schedule]
+      if configuration[:schedule]
+        if configuration[:schedule][:range_indices][:start] > 6
+          raise OptimizerWrapper::DiscordantProblemError.new('Api does not support schedule start index bigger than 6 yet')
+          # TODO : allow start bigger than 6 and make code consistent with this
+        end
+
+        if configuration[:schedule][:range_indices][:start] > configuration[:schedule][:range_indices][:end]
+          raise OptimizerWrapper::DiscordantProblemError.new('Schedule start index should be less than or equal to end')
+        end
+      else
         (hash[:services] + hash[:shipments]).each{ |s|
-          raise OptimizerWrapper::DiscordantProblemError, 'There can not be more than one visit if no schedule is provided' unless s[:visits_number].to_i <= 1
+          raise OptimizerWrapper::DiscordantProblemError.new(
+            'There can not be more than one visit if no schedule is provided') unless s[:visits_number].to_i <= 1
         }
       end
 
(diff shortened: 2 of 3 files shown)


lib/interpreters/split_clustering.rb
```diff
@@ -793,13 +793,19 @@ def self.collect_cluster_data(vrp, nb_clusters)
     def self.duplicate_vehicle(vehicle, timewindow, schedule)
       if timewindow.nil?
         (schedule[:start]..[schedule[:end], 6].min).collect{ |day|
+          vehicle.skills = vehicle.skills.collect{ |sk_set|
+            sk_set | [%w[mon tue wed thu fri sat sun][day]]
```
senhalil, line 797: Can we have these days in symbol `%i[]` if the rest of the change is not that painful? 

With string, we allocate memory for each comparison `(v_skills - s_skills).empty?` even inside the gem. For "words" that are general enough symbol is better than string because memory is allocated once.

lib/interpreters/split_clustering.rb
```diff
@@ -793,13 +793,19 @@ def self.collect_cluster_data(vrp, nb_clusters)
     def self.duplicate_vehicle(vehicle, timewindow, schedule)
       if timewindow.nil?
         (schedule[:start]..[schedule[:end], 6].min).collect{ |day|
```
senhalil, line 795: if schedule[:start] is greater than 6 this creates no TW, is that what we want? I feel like this logic needs `% 7`

lib/interpreters/split_clustering.rb
senhalil: On top of this, we can extract the `%i\w[mon tue wed thu fri sat sun]` to a class constant such as `WEEKDAYS = %i\w[mon tue wed thu fri sat sun].freeze` and use it whenever necessary, instead of creating a new array each time we need such a comparison.

senhalil (changes requested):

lib/interpreters/split_clustering.rb
fonsecadeline: For the sake of efficiency I suggest that I do these changes at the same time as I work on https://gitlab.com/mapotempo/optimizer-api/-/issues/635 (which I am actually working on right now). This will be great to differenciate internal and original skills.

lib/interpreters/split_clustering.rb
fonsecadeline: I would make no sense for user to provide a start bigger than 6 but you are right, nothing prevents this case.

lib/interpreters/split_clustering.rb
senhalil: In that dev, you can even turn general skills into symbol at the API level (since form the user point of view nothing changes -- string and symbol are the in JSON). We know that we cannot have a VRP with an infinite number of different skills. That change would make skill comparisons faster as well.

lib/interpreters/split_clustering.rb
fonsecadeline: In fact this is a tricky case. We should rather ensure start is always smaller or equal to 6 (because it would make no sense otherwise). Otherwise dealing with this case is quite hard : 

If end - start > 7 we need to generate vehicles for every index, otherwise we need to know if start and end are in the same week in order to know if we do (start % 7 .. end % 7) or (end % 7 .. start % 7).

Therefore I will just reject case where start > 6 because that makes no sense anyway.

senhalil: Why do you think it doesn't make sense? For example, a user who uses the day of year as their day index might want to use [182..365] as [start..end] to create a plan for the second half of year because all other indices for their vehicles, services etc comes from their ERP as day #.  If we prevent them, we are forcing them to modify all their day indices with their actual horizon[:start] to make the horizon[:start] to be in the range. 

It is not that hard, it would be similar to what we do for the general case:

```
(0..6).collect{ |day|
next unless (schedule[:start]..schedule[:end]).any?{ |day_index| day_index % 7 == day }

// some code
```

models/vrp.rb
```diff
@@ -220,9 +220,18 @@ def self.check_consistency(hash)
       end
 
       # number of visits consistency
-      if !configuration[:schedule]
+      if configuration[:schedule]
+        if configuration[:schedule][:range_indices][:start] > 6
+          raise OptimizerWrapper::DiscordantProblemError.new('Schedule start index should be less or equal to 6')
+        end
```
senhalil, line 226: I don't think this is necessary and it might be limiting for some users and it will prevent our internal code from being generic. https://github.com/Mapotempo/optimizer-api/pull/158/files#r594430415

models/vrp.rb
```diff
@@ -220,9 +220,18 @@ def self.check_consistency(hash)
...
+
+        if configuration[:schedule][:range_indices][:start] > configuration[:schedule][:range_indices][:end]
+          raise OptimizerWrapper::DiscordantProblemError.new('Schedule start index should be less than end')
```
senhalil, line 229: less than or equal to the end index?

lib/interpreters/split_clustering.rb
fonsecadeline: I think it makes no sense because range_indices are meant to be used by users that do not know exact date of schedule, but they know if it starts on a monday or friday and they know how long their period is.

If they know the date then they can just use range_date. I think that is how it was supposed to be, and the aim is that no user uses range_indices anyway.

(I thought you were suggesting something else, shorter, anyway). It is not hard to implement, but we have this logic in several parts of the code and it needs more lines than just (schedule[:start]..schedule[:end]).

models/vrp.rb
```diff
@@ -456,6 +465,12 @@ def self.detect_date_indices_inconsistency(hash)
     def self.generate_schedule_indices_from_date(hash)
       return hash if !hash[:configuration] || !hash[:configuration][:schedule]
 
+      schedule = hash[:configuration][:schedule]
+      if (!schedule[:range_indices] || [schedule[:range_indices][:start], schedule[:range_indices][:end]].compact.size < 2) ||
+         (!schedule[:range_date] || [schedule[:range_date][:start], schedule[:range_date][:end]].compact.size < 2)
```
senhalil, line 470: This condition is wrong. It raises if we have proper `schedule[:range_indices]` with `start` and `end` but no `schedule[:range_date]`, no?

I also suggest replacing `[].compact.size < 2` with `[].any?(&:nil?)` or just good old `schedule[:range_indices][:start].nil? || schedule[:range_indices][:end].nil?` to be more expressive -- i.e., "if any of them nil?" instead of "if the size of the array without the nils is less than two?"

senhalil (changes requested):

models/vrp.rb
fonsecadeline: If it is equal then it makes no sense to provide a schedule.

lib/interpreters/split_clustering.rb
senhalil: If we ignore the TW case here, I don't see why our code should work for [0, 182] but break for [183, 365]. At the end of the day, the latter case needs a preprocessing of `-183` for all indices and both cases are equal. That's why I can't understand why our code cannot handle these otherwise equivalent problems.

models/vrp.rb
fonsecadeline: I am not sure this is really limiting for our users. @braktar , @fab-girard can you give us a second opinion ?

lib/interpreters/split_clustering.rb
fonsecadeline: From my point of you we are getting off topic but we can discuss it on a Gitlab ticket. I agree we can work on making this more generic but at the same time I think we talked about making our customers always use dates so this will not be a problem in fact.

models/vrp.rb
senhalil: I meant your message is not in accord with your `if` condition! 

I don't want to get into an argument but I don't understand this "makes no sense" argument. If scheduling code can handle scheduling 2-days, it should be able to handle 1-day without any issues -- which is just calling OR-tools. It making no sense doesn't mean that our code shouldn't be generic. Generic code is easier to maintain, expand and plug/play with other code.

lib/interpreters/split_clustering.rb
senhalil: I don't remember we talked about that. I remember that we talked about the fact that we don't need both date and indices inside the codebase, and we should develop only with one and chose indices if I am not mistaken. Otherwise, I don't think it makes sense to prevent users from using indices in the future (because we don't know every software that will interface with optim-api).

senhalil, line 795: The following code is less error prone and correct compared to the previous code and handles those "makes no sense" cases naturally without fuss
```suggestion
        (0..6).flat_map{ |day|
          next unless (schedule[:start]..schedule[:end]).any?{ |day_index| day_index % 7 == day }
```

senhalil (changes requested):

models/vrp.rb
fonsecadeline: You are right, my message and error are not consistent. And we can deal with the case where start and end are equal. It is just not intuitive to me. But you are right, it is better if we want to avoid throwing useless errors.

models/vrp.rb
fonsecadeline: We can discuss it later : https://gitlab.com/mapotempo/optimizer-api/-/issues/701

braktar (approved):

senhalil (approved):

PR #160, opened 2021-03-17
Fix : bad value for range

fonsecadeline:
https://sentry.mapotempo.com/mapotempo/api-optimizer-prod/issues/15102/

### DIFF
--- a/models/vrp.rb
+++ b/models/vrp.rb
@@ -453,10 +453,10 @@ module Models
     def self.detect_date_indices_inconsistency(hash)
       missions_and_vehicles = hash[:services].to_a + hash[:shipments].to_a + hash[:vehicles].to_a
       has_date = missions_and_vehicles.any?{ |m|
-        !(m[:unavailable_date_ranges] || m[:unavailable_work_date]).to_a.empty?
+        (m[:unavailable_date_ranges] || m[:unavailable_work_date])&.any?
       }
       has_index = missions_and_vehicles.any?{ |m|
-        !(m[:unavailable_index_ranges] || m[:unavailable_work_day_indices]).to_a.empty?
+        (m[:unavailable_index_ranges] || m[:unavailable_work_day_indices])&.any?
       }
       if (hash[:configuration][:schedule][:range_indices] && has_date) ||
          (hash[:configuration][:schedule][:range_date] && has_index)
--- a/models/vrp.rb
+++ b/models/vrp.rb
@@ -452,8 +452,12 @@ module Models
 
     def self.detect_date_indices_inconsistency(hash)
       missions_and_vehicles = hash[:services].to_a + hash[:shipments].to_a + hash[:vehicles].to_a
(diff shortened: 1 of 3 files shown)


braktar (approved):

models/vrp.rb
```diff
@@ -452,8 +452,12 @@ def self.deduce_unavailable_days(hash, element, start_index, end_index, type)
...
     def self.detect_date_indices_inconsistency(hash)
       missions_and_vehicles = hash[:services].to_a + hash[:shipments].to_a + hash[:vehicles].to_a
-      has_date = missions_and_vehicles.any?{ |m| m[:unavailable_date_ranges] || m[:unavailable_work_date] }
-      has_index = missions_and_vehicles.any?{ |m| m[:unavailable_index_ranges] || m[:unavailable_work_day_indices] }
+      has_date = missions_and_vehicles.any?{ |m|
+        !(m[:unavailable_date_ranges] || m[:unavailable_work_date]).to_a.empty?
+      }
+      has_index = missions_and_vehicles.any?{ |m|
+        !(m[:unavailable_index_ranges] || m[:unavailable_work_day_indices]).to_a.empty?
+      }
```
senhalil, line 460: simpler is better
`&.any?` instead of `!.to_a.empty?`
```suggestion
      has_date = missions_and_vehicles.any?{ |m|
        (m[:unavailable_date_ranges] || m[:unavailable_work_date])&.any?
      }
      has_index = missions_and_vehicles.any?{ |m|
        (m[:unavailable_index_ranges] || m[:unavailable_work_day_indices])&.any?
      }
```

test/models/vrp_consistency_test.rb
```diff
@@ -305,7 +305,7 @@ def test_duplicated_ids_are_not_allowed
 
     def test_dates_cannot_be_mixed_with_indices
       vrp = VRP.scheduling # contains schedule[:range_indices]
-      vrp[:vehicles].first[:unavailable_work_date] = Date.new(2021, 2, 11)
+      vrp[:vehicles].first[:unavailable_work_date] = [Date.new(2021, 2, 11)]
```
senhalil, line 308: Comment out of the scope of this PR:

This parameter is an array, it should have been called `unavailable_work_dates` :facepalm: 

Is this the parameter we recently replaced with `unavailable_x_ranges`? Or this one does something different?

IF it is the one that is "replaced" and we want people to use the  `unavailable_date/index_ranges` logic then we can make `unavailable_work_day_indices` and `unavailable_work_date` hidden in the document. IF we want to keep it, we should hide/deprecate this one and add a `unavailable_work_dates` parameter to the public API.

@fonsecadeline I think you are more informed about the use-cases so I think you should decide... can you create a ticket accordingly?

senhalil (changes requested):

test/models/vrp_consistency_test.rb
fonsecadeline: https://gitlab.com/mapotempo/optimizer-api/-/issues/704

senhalil (approved):

PR #162, opened 2021-03-18
Move to Github Actions

alain-andre:
This moves CI from Travis to GitHub actions. I forced the tests to start only on dev branch like it was asked before.

I would prefer to launch tests. For example this PR does not show the tests results you can see on this [branch](https://github.com/alain-andre/optimizer-api/tree/dev)


fonsecadeline:
When you say 

> I forced the tests to start only on dev branch 

does it mean that if I push on a branch called _hot_fix_ then no integration tests will be run ? Is that what we want ?

README.md
```diff
@@ -1,4 +1,5 @@
-# Mapotempo Optimizer API [![Build Status](https://travis-ci.com/Mapotempo/optimizer-api.svg?branch=master)](https://travis-ci.com/Mapotempo/optimizer-api)
+# Mapotempo Optimizer API ![dev Build Status](https://github.com/Mapotempo/optimizer-api/actions/workflows/main.yml/badge.svg?branch=dev)
```
fab-girard, line 1: why do you replace master branch by dev branch?

.github/actions/tests.sh
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+
+CONTAINER=${PROJECT}_api.1.$(docker service ps -f "name=${PROJECT}_api.1" ${PROJECT}_api -q --no-trunc | head -n1)
+docker exec -i ${CONTAINER} bundle exec rake test ${TEST_ENV} ${OPTIONS}
```
fab-girard, line 4: what is this env var `TEST_ENV`?

senhalil:
If we have to choose, I think we should run the tests on PR's so that we don't merge a broken PR. 

Ideally we would run the tests on every branch and not just `dev` branch, this was what we did on Travis-CI. Since we do our developments on different branches (so that we can keep the PR's separate), it is important to run the tests on other branches.

README.md
alain-andre: I thought it was interesting to specify the dev, but I will move it back to master branch.

alain-andre:
> When you say
> 
> > I forced the tests to start only on dev branch
> 
> does it mean that if I push on a branch called _hot_fix_ then no integration tests will be run ? Is that what we want ?

As seen together, it was a misunderstanding, I change this to be run on any branch.

.github/actions/tests.sh
alain-andre: This comes from the old script, I remove it.

fonsecadeline (approved):
I do not have a great understanding of this but it seems good to me (mainly main.yml)

.github/workflows/main.yml
```diff
@@ -0,0 +1,145 @@
+name: Docker Image CI
+
+on:
+  push:
+
+env:
+  NB_SERVICES: ${{ secrets.NB_SERVICES }}
+  OPTIMIZER_ORTOOLS_VERSION: ${{ secrets.OPTIMIZER_ORTOOLS_VERSION }}
```
braktar, line 8: Do we have a way to put a default version ?

alain-andre: It uses the default version specified in Dockerfile. Those variables allow you to override the values.

senhalil: These secrets, are they like hidden environment variables on github side or they are defined inside our repo?

.github/workflows/main.yml
```diff
@@ -0,0 +1,145 @@
...
+      - uses: actions/checkout@v1
+      - name: Cache
+        id: cache
+        uses: actions/cache@v2
+        with:
+          path: vendor/bundle
+          key: ${{ runner.os }}
+      - name: Build image
+        run: ./.github/actions/build.sh
+        shell: bash
```
senhalil, line 28: Can't we create the image just once and launch the tests after? I mean that looks more like proper workflow unless it is complicated.

alain-andre: We cannot create the image and pass it to another job. At every job, or we build it, or we download it.

.github/workflows/main.yml
alain-andre: Secrets are defined inside your repo in the GitHub settings : `settings/secrets/actions`

.github/workflows/main.yml
senhalil: So there is no way to set a common "environment" for all jobs :/ weird. Thanks Alain!

.github/actions/build.sh
```diff
@@ -0,0 +1,42 @@
...
+  nbps=$(docker service ls | grep 1/1 | awk '{print $4}' | wc -l)
+  if [ ${nbps} -eq ${nb_services} ];
+  then
+    echo "All services up, starting tests"
+    break;
+  fi
+
+  echo "Waiting for all services (${nbps}/${nb_services}) to start ($cpt secondes)."
+  sleep 1
+done
```
senhalil, line 42: Do we need this "try/fail" logic for Github as well?

senhalil (approved):

.github/actions/build.sh
alain-andre: Yes we need it to be sure all services are up before starting tests.

braktar (approved):

PR #163, opened 2021-03-23
Small fixes and clean

fonsecadeline:


test/lib/interpreters/split_clustering_test.rb
```diff
@@ -948,6 +948,27 @@ def test_list_vehicles
...
+    def test_split_with_vehicle_alternative_skills
+      vrp = TestHelper.create(VRP.lat_lon_scheduling_two_vehicles)
+
+      # nothing expected to be raised :
+      [[[]], [['skill']], [['skill'], []]].each{ |skill_set|
+        vrp.vehicles.first.skills = skill_set
+        Interpreters::SplitClustering.split_balanced_kmeans(
+          { vrp: vrp, service: :demo }, vrp.vehicles.size,
+          ntity: :vehicle, restarts: 1, max_iterations: 1)
+      }
```
senhalil, line 961: This is a general comment: it is dangerous to modify the VRP object directly like this in the tests (unless it is necessary or the test is about that) because we skip the validations done in coerce and create function. 

Changes in these functions will not be represented in the tests and we risk having a test that will not fail when it should or vice-versa, a test that breaks when it really shouldn't. 

The safe option is to always modify the json and call the create on the modified problem when it is necessary 
```suggestion
      problem = VRP.lat_lon_scheduling_two_vehicles

      # nothing expected to be raised :
      [[[]], [['skill']], [['skill'], []]].each{ |skill_set|
        problem[:vehicles].first[:skills] = skill_set
        Interpreters::SplitClustering.split_balanced_kmeans(
          { vrp: TestHelper.create(problem), service: :demo }, problem[:vehicles].size,
          entity: :vehicle, restarts: 1, max_iterations: 1)
      }
```

senhalil (changes requested):

test/lib/interpreters/split_clustering_test.rb
```diff
@@ -948,6 +948,27 @@ def test_list_vehicles
...
+
+      # only this case should raise, because we do have alternative skills :
+      vrp.vehicles.first.skills = [['skill'], ['other_skill']]
+      error = assert_raises OptimizerWrapper::UnsupportedProblemError do
+        Interpreters::SplitClustering.split_balanced_kmeans(
+          { vrp: vrp, service: :demo }, vrp.vehicles.size,
+          ntity: :vehicle, restarts: 1, max_iterations: 1)
+      end
+      assert_equal 'Cannot use balanced kmeans if there are vehicles with alternative skills', error.message
```
senhalil, line 970: I thought it was clear that my suggestion was just half of the test and the second half needed fixing as well. Maybe I shouldn't use suggestion unless it is "acceptable" without modification or add a "warning" :thinking: 
```suggestion
      problem[:vehicles].first[:skills] = [['skill'], ['other_skill']]
      error = assert_raises OptimizerWrapper::UnsupportedProblemError do
        Interpreters::SplitClustering.split_balanced_kmeans(
          { vrp: TestHelper.create(problem), service: :demo }, problem[:vehicles].size,
          entity: :vehicle, restarts: 1, max_iterations: 1)
      end
      assert_equal 'Cannot use balanced kmeans if there are vehicles with alternative skills', error.message
```

senhalil (changes requested):
It needs the last suggestion for the test to work. Otherwise, it is okay.

test/lib/interpreters/split_clustering_test.rb
fonsecadeline: Sorry, in fact when I click link from email it shows me your suggestion in may page. I thought you had changed it all.

senhalil (approved):

PR #165, opened 2021-03-25
Visualize results in geojson

fonsecadeline:


fonsecadeline:
This PR requires https://github.com/Mapotempo/optimizer-api/pull/153

fonsecadeline:
- [ ] Add test
- [ ] Validate with GithubActions

fonsecadeline:
This is only a first version, I will work on merging OutputHelper::Clustering and OutputHelper::Result functions (collect_hulls mainly).

CHANGELOG.md
```diff
@@ -15,6 +16,7 @@
 - Bump VROOM to v1.8.0 and start using the features integrated since v1.3.0 [#107](https://github.com/Mapotempo/optimizer-api/pull/107)
 - Bump OR-Tools v7.8 [#107](https://github.com/Mapotempo/optimizer-api/pull/107)
 - VROOM were previously always called synchronously, it is now reserved to a set of effective `router_mode` (:car, :truck_medium) within a limit of points (<200). [#107](https://github.com/Mapotempo/optimizer-api/pull/107)
+- Geometry_polylines parameter should now be provided through geometry parameter [#165](https://github.com/Mapotempo/optimizer-api/pull/165)
```
braktar, line 19: It should be in the Deprecated section
```suggestion
- `geometry_polyline` parameter now be provided through `geometry` parameter [#165](https://github.com/Mapotempo/optimizer-api/pull/165)
```

api/v01/entities/vrp_input.rb
```diff
@@ -192,8 +192,9 @@ module VrpConfiguration
   end
 
   params :vrp_request_restitution do
-    optional(:geometry, type: Boolean, desc: 'Allow to return the MultiLineString of each route')
-    optional(:geometry_polyline, type: Boolean, desc: 'Encode the MultiLineString')
+    optional(:geometry, type: Array[Symbol], default: [], coerce_with: ->(value) { GeometryType.type_cast(value) },
+                        desc: 'Specifies which geometry needs to be returned. You can choose among points, polylines and partitions or use boolean to output all or no geometry.')
```
braktar, line 196:
```suggestion
                        desc: 'Specifies which geometry needs to be returned. You can choose among points, polylines and partitions or use boolean to output all or none geometry.')
```

api/v01/entities/vrp_input.rb
```diff
@@ -279,7 +280,9 @@ module VrpMisc
     optional(:router_mode, type: String, desc: '`car`, `truck`, `bicycle`, etc... See the Router Wrapper API doc')
     optional(:router_dimension, type: String, values: ['time', 'distance'], desc: 'time or dimension, choose between a matrix based on minimal route duration or on minimal route distance')
     optional(:speed_multiplier, type: Float, default: 1.0, desc: 'multiply the current modality speed, default : 1.0')
-    optional(:skills, type: Array[String], desc: 'Particular abilities required by a vehicle to perform this subtour', coerce_with: ->(val) { val.is_a?(String) ? val.split(/,/) : val })
+    optional(:skills, type: Array[Symbol],
```
braktar, line 283: Have you looked what swagger displays for symbol type ?

api/v01/entities/vrp_result.rb
```diff
@@ -129,8 +129,20 @@ class VrpResultJob < Grape::Entity
...
 
+    class VrpResultVisualClusters < Grape::Entity
+      expose :vehicle, expose_nil: false, documentation: { desc: 'Vehicle partition visual' }
+      expose :work_day, expose_nil: false, documentation: { desc: 'Work_day partition visual' }
+    end
+
+    class VrpResultVisual < Grape::Entity
+      expose :partitions, expose_nil: false, using: VrpResultVisualClusters, documentation: { is_array: true, desc: 'According to specified geometry and partitions parameter, geojsons representing each partition' }
+      expose :points, expose_nil: false, documentation: { desc: 'Points visualization' }
+      expose :polylines, expose_nil: false, documentation: { desc: 'Polylines visualization' }
```
braktar, line 140: Are Points and polylines arrays ?

lib/output_helper.rb
```diff
@@ -90,6 +90,192 @@ def self.csv_line(vrp, service, cluster_index, two_stages = false)
...
+      all_skills = elements.collect{ |a| a[:detail][:internal_skills] }.flatten.uniq
+
+      has_vehicle_partition = all_skills.any?{ |skill| skill.to_s.include?('vehicle_partition_') }
+      has_work_day_partition = all_skills.any?{ |skill| skill.to_s.include?('work_day_partition_') }
+      return unless has_vehicle_partition || has_work_day_partition
+
+      result[:geojsons][0][:partitions] ||= {}
+      if has_vehicle_partition
+        @vehicle_partitioned = true
+        result[:geojsons][0][:partitions][:vehicle] = draw_cluster(elements, :vehicle)
```
braktar, line 127: :partitions is expected in the output to be an array, but it is a hash here

api/v01/entities/vrp_result.rb
```diff
@@ -129,8 +129,20 @@ class VrpResultJob < Grape::Entity
       expose :graph, using: VrpResultJobGraphItem, documentation: { is_array: true, desc: 'Items to plot cost evolution' }
     end
 
+    class VrpResultVisualClusters < Grape::Entity
+      expose :vehicle, expose_nil: false, documentation: { desc: 'Vehicle partition visual' }
+      expose :work_day, expose_nil: false, documentation: { desc: 'Work_day partition visual' }
```
braktar, line 134: my assumption is that a result should have a collection of partitions, which each have a name, a type and a feature collection

braktar (changes requested):

api/v01/entities/vrp_result.rb
fonsecadeline: No

optimizer_wrapper.rb
```diff
@@ -98,22 +99,22 @@ def self.wrapper_vrp(api_key, services, vrp, checksum, job_id = nil)
 
     if services_vrps.any?{ |sv| !sv[:service] }
       raise UnsupportedProblemError.new('Cannot apply any of the solver services', inapplicable_services)
-    elsif vrp.restitution_geometry && !vrp.points.all?{ |point| point[:location] }
-      raise DiscordantProblemError, 'Geometry is not available if locations are not defined'
+    elsif vrp.restitution_geometry.size.positive? && !vrp.points.all?{ |point| point[:location] }
```
braktar, line 102:
```suggestion
    elsif vrp.restitution_geometry.any? && !vrp.points.all?{ |point| point.location }
```

wrappers/ortools.rb
```diff
@@ -118,7 +118,7 @@ def solve(vrp, job, thread_proc = nil, &block)
       services_positions = { always_first: [], always_last: [], never_first: [], never_last: [] }
       vrp.services.each_with_index{ |service, service_index|
         vehicles_indices =
-          if !service[:skills].empty? && (vrp.vehicles.all? { |vehicle| vehicle.skills.empty? }) &&
+          if !service.skills.empty? && (vrp.vehicles.all? { |vehicle| vehicle.skills.empty? }) &&
```
braktar, line 121:
```suggestion
          if service.skills.any? && (vrp.vehicles.all? { |vehicle| vehicle.skills.empty? }) &&
```

wrappers/ortools.rb
```diff
@@ -201,7 +201,7 @@ def solve(vrp, job, thread_proc = nil, &block)
         end
       }
       vrp.shipments.each_with_index{ |shipment, shipment_index|
-        vehicles_indices = if !shipment[:skills].empty? && (vrp.vehicles.all? { |vehicle| vehicle.skills.empty? })
+        vehicles_indices = if !shipment.skills.empty? && (vrp.vehicles.all? { |vehicle| vehicle.skills.empty? })
```
braktar, line 204:
```suggestion
        vehicles_indices = if shipment.skills.any? && (vrp.vehicles.all? { |vehicle| vehicle.skills.empty? })
```

braktar (changes requested):

api/v01/vrp.rb
```diff
@@ -227,7 +237,8 @@ class Vrp < APIBase
                       status: :killed,
                       avancement: job.message,
                       graph: solution[:graph]
-                    }
+                    },
+                    geojsons: solution[:result] ? [solution[:result]].flatten(1).first[:geojsons] : nil
```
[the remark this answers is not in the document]
fonsecadeline: This would not work, we would rather need something like

```suggestion
                    geojsons: solution[:result]&.first&[:geojsons]
```

but this returns _false_ if there is no solution[:result], we want to return _nil_ in this case.

api/v01/entities/vrp_result.rb
fonsecadeline: In fact this 'collection' is represented by the fact that geojson is an array. Therefore we have different partitions according to each repetition. 

Each partition contains _vehicle_ and _work_day_ keys to represent each type of partition.

api/v01/entities/vrp_input.rb
fonsecadeline: I am not sure how to access swagger from my local code.

api/v01/vrp.rb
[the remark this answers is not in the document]
braktar: we can add `|| nil` to force the false to become nil

api/v01/entities/vrp_input.rb
fonsecadeline: imagen

braktar: And on the up to date swagger-ui ?

fonsecadeline: This already is the up to date swagger-ui. I opened the file with this code running on server/worker.

optimizer_wrapper.rb
```diff
@@ -808,9 +814,7 @@ def self.apply_zones(vrp)
 
         next unless zone.inside(activity_loc.lat, activity_loc.lon)
 
-        service.sticky_vehicles += zone.vehicles
-        service.sticky_vehicles.uniq!
-        service.skills += [zone[:id]]
+        service.skills += [zone[:id].to_sym]
```
senhalil, line 817: can't we make the zone ID's symbol at the API level so that we don't have this to_sym business?

optimizer_wrapper.rb
```diff
@@ -796,7 +802,7 @@ def self.apply_zones(vrp)
       next if zone.vehicles.compact.empty?
 
       zone.vehicles.each{ |vehicle|
-        vehicle.skills.each{ |skillset| skillset << zone[:id] }
+        vehicle.skills.each{ |skillset| skillset << zone[:id].to_sym }
```
senhalil, line 805: we need a discordantproblem raise here for vehicles and services in case the zone[:id] already exists as a skill. and a consistency check at the vrp.create level to refuse it as discordantproblem similar to what we do for reserved_internal_skill clash.

test/api/v01/output_test.rb
```diff
@@ -74,6 +74,17 @@ def test_no_day_week_num
...
   end
 
+  def test_returned_skills
+    vrp = VRP.basic
+    vrp[:vehicles].first[:skills] = [['skill_to_output']]
+    vrp[:services].first[:skills] = ['skill_to_output']
+
+    response = post '/0.1/vrp/submit', { api_key: 'demo', vrp: vrp }.to_json, 'CONTENT_TYPE' => 'application/json'
+    result = JSON.parse(response.body)['solutions'].first
+    activities = result['routes'].collect{ |r| r['activities'] }.flatten
```
senhalil, line 84: flat_map

senhalil: can you check all collect{}.flatten's you added and replace them with flat_map?

I don't understand why rubocop doesn't bug you with this :thinking:

test/lib/interpreters/split_clustering_test.rb
```diff
@@ -315,6 +316,9 @@ def test_unavailable_days_taken_into_account_vehicle_work_day
       vrp[:services][0][:activity][:timewindows] = [{ start: 0, end: 10, day_index: 0 }]
       vrp[:services][3][:activity][:timewindows] = [{ start: 0, end: 10, day_index: 1 }]
       service_vrp = { vrp: TestHelper.create(vrp), service: :demo }
+      service_vrp[:vrp].services.each{ |s|
+        s.skills = []
+      }
```
senhalil, line 321: unless you have a specific reason?
```suggestion
      service_vrp[:vrp].services.each{ |s| s.skills = [] }
```

test/models/vrp_test.rb
```diff
@@ -304,5 +304,22 @@ def test_no_lapse_in_relation
...
+
+    def test_original_skills_and_skills_are_equal_after_create
+      vrp = VRP.basic
+      vrp[:vehicles].first[:skills] = [['skill_to_output']]
+      vrp[:services].first[:skills] = ['skill_to_output']
+
+      created_vrp = Models::Vrp.create(vrp)
+      assert_equal 1, created_vrp.services.first.skills.size
+      assert_equal created_vrp.services.first.original_skills.size, created_vrp.services.first.skills.size
+    end
```
senhalil, line 316: I feel like I just saw this test, maybe in output_test. Is it a duplicate?

test/models/vrp_test.rb
```diff
@@ -304,5 +304,22 @@ def test_no_lapse_in_relation
...
+
+    def test_geometry_provided
+      vrp = VRP.toy
+      vrp[:configuration][:restitution] = { geometry: %i[partitions points polylines], geometry_polyline: true }
+      Models::Vrp.ensure_retrocompatibility(vrp)
+      assert_equal %i[polylines], vrp[:configuration][:restitution][:geometry]
```
senhalil, line 322: This should raise. Geometry is a new field, if someone is starting to use the new field, they can stop using the old field. 

If only the old field `geometry_polyline: true` is sent, it should continue to work as expected (as if they sent `geometry: %i[polylines]`) but if they sent the two, we should raise like the `solver_parameter`. Otherwise, we are erasing the info sent by the user which might be a mistake so we should fail "loudly"

util/job_manager.rb
```diff
@@ -48,6 +48,7 @@ def perform
       OptimizerWrapper::REDIS.set(Resque::Plugins::Status::Hash.status_key(self.uuid), Resque::Plugins::Status::Hash.encode(value))
 
       ask_restitution_csv = services_vrps.any?{ |s_v| s_v[:vrp].restitution_csv }
+      ask_restitution_geojson = services_vrps.collect{ |s_v| s_v[:vrp].restitution_geometry }.flatten.uniq
```
senhalil, line 51: flat_map

senhalil: in fact, `any?` should work as the way it works for `ask_restitution_csv`, no?

wrappers/ortools.rb
```diff
@@ -118,7 +118,7 @@ def solve(vrp, job, thread_proc = nil, &block)
       services_positions = { always_first: [], always_last: [], never_first: [], never_last: [] }
       vrp.services.each_with_index{ |service, service_index|
         vehicles_indices =
-          if !service[:skills].empty? && (vrp.vehicles.all? { |vehicle| vehicle.skills.empty? }) &&
+          if service.skills.any? && (vrp.vehicles.all? { |vehicle| vehicle.skills.empty? }) &&
```
senhalil, line 121:
```suggestion
          if service.skills.any? && vrp.vehicles.all?{ |vehicle| vehicle.skills.empty? } &&
```

wrappers/wrapper.rb
```diff
@@ -501,11 +501,29 @@ def build_rest(rest, vehicle = nil)
       }
     end
 
+    def build_skills(job)
```
senhalil, line 504: build_skill_detail ? otherwise, it doesn't sound related to output/detail

wrappers/wrapper.rb
```diff
@@ -501,11 +501,29 @@ def build_rest(rest, vehicle = nil)
...
+
+      skills_to_output.flatten
+    end
+
     def build_detail(job, activity, point, day_index, job_load, vehicle, delivery = nil)
       {
         lat: point&.location&.lat,
         lon: point&.location&.lon,
-        skills: job&.skills,
+        skills: build_skills(job),
```
senhalil, line 525: can't we just call the `original_skills` ?

senhalil (changes requested):
Phew, this was a long one :sweat_smile: thanks @fonsecadeline ! 

My main comment is that we can make the `generate_geometry` function return the generated geometries and keep nothing -- like we do for csv functionality. That was the main point of doing this at the end point. That way when we make a change (improvement/extension/correction) to the geometry output/functionality, solutions that are already GET'ed/dumped can be re-GET'ed to have the latest geometry without an issue -- like the way it is possible for us to add a new column to the CSV output and new GETs would show this new column.

Otherwise, there are some small fixes to do but I think, the PR is mainly there.

models/types/geometry_type.rb
```diff
@@ -0,0 +1,39 @@
...
+#
+
+class GeometryType
+  def self.type_cast(value)
+    all_types = %i[points polylines partitions]
+    if value.is_a?(FalseClass)
+      []
+    elsif value.is_a?(TrueClass)
+      all_types
+    elsif value.is_a?(Array)
```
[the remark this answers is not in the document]
fonsecadeline: You are right, I did not deal with the case where we provide a string. I will make this consistent with other parameters that do accept strings.

util/job_manager.rb
fonsecadeline: s_v[:vrp].restitution_csv is a boolean while s_v[:vrp].restitution_geometry can be an empty array and, in this case, we do not want to call function.

Well in fact that was initial idea. Now we always want to provide, at least, points. So in fact we do not even need this anymore.

api/v01/entities/vrp_input.rb
```diff
@@ -192,8 +192,9 @@ module VrpConfiguration
   end
 
   params :vrp_request_restitution do
-    optional(:geometry, type: Boolean, desc: 'Allow to return the MultiLineString of each route')
-    optional(:geometry_polyline, type: Boolean, desc: 'Encode the MultiLineString')
+    optional(:geometry, type: Array[Symbol], default: [], coerce_with: ->(value) { GeometryType.type_cast(value) },
```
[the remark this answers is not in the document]
fonsecadeline: This should work ! But GitHub Actions will confirm.

api/v01/entities/vrp_input.rb
```diff
@@ -489,8 +496,13 @@ module VrpVehicles
     optional(:distance, type: Integer, desc: 'Maximum tour distance. Not available with periodic heuristic.')
     optional(:maximum_ride_time, type: Integer, desc: 'Maximum ride duration between two route activities')
     optional(:maximum_ride_distance, type: Integer, desc: 'Maximum ride distance between two route activities')
-    optional :skills, type: Array[Array[String]], desc: 'Particular abilities which could be handle by the vehicle. This parameter is a set of alternative skills, and must be defined as an Array[Array[String]]. Not available with periodic heuristic.',
-                      coerce_with: ->(val) { val.is_a?(String) ? [val.split(/,/).map(&:strip)] : val } # TODO : Create custom coerce to consider multiple alternatives
+    optional :skills, type: Array[Array[Symbol]],
+                      coerce_with: ->(val) {
+                        val.is_a?(String) ?
+                        [val.split(/,/).map(&:strip).map(&:to_sym)] :
```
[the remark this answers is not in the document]
fonsecadeline: I guess we can do it for both.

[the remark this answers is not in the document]
fonsecadeline: I will maybe do it in another PR in order not to change any additional behaviour .

api/v01/entities/vrp_result.rb
```diff
@@ -129,8 +129,20 @@ class VrpResultJob < Grape::Entity
...
+
+    class VrpResultVisual < Grape::Entity
+      expose :partitions, expose_nil: false, using: VrpResultVisualClusters, documentation: { desc: 'According to specified geometry and partitions parameter, geojsons representing each partition' }
+      expose :points, expose_nil: false, documentation: { desc: 'Points visualization' }
+      expose :polylines, expose_nil: false, documentation: { desc: 'Polylines visualization' }
+    end
+
     class VrpResult < Grape::Entity
       expose :solutions, using: VrpResultSolution, documentation: { is_array: true, desc: 'The current best solution' }
+      expose :geojsons, using: VrpResultVisual, documentation: { is_array: true, desc: 'If required through geometry VRP parameter, set of geojsons generated' }
```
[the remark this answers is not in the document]
fonsecadeline: In fact partitions were initially documented as being an array, which was a mistake. I think after discussing in one of our daily @braktar was okay with this.

geojsons are array to allow us to return one geojsons structure per repetition or per method.

api/v01/vrp.rb
```diff
@@ -174,7 +178,8 @@ class Vrp < APIBase
                   status: job&.status&.to_sym || :completed, # :queued, :working, :completed, :failed
                   avancement: job&.message,
                   graph: solution[:graph]
-                }
+                },
+                geojsons: solution[:result]&.first&[:geojsons] || nil
```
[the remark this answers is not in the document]
fonsecadeline: It is not written like this because there are several checks to make before calling OutputHelper::Result.generate_geometry() but we do call this every time we call get (from the time job is completed).

lib/interpreters/split_clustering.rb
```diff
@@ -1203,15 +1207,17 @@ def add_corresponding_entity_skills(entity, vrp)
         case entity
         when :vehicle
           vrp.services.each{ |service|
-            service.skills.insert(0, vrp.vehicles.first.id)
+            service.skills.insert(0, "vehicle_partition_#{vrp.vehicles.first.id}".to_sym)
+          }
+          vrp.vehicles.each{ |v|
+            v.skills.each{ |set| set << "vehicle_partition_#{vrp.vehicles.first.id}".to_sym }
```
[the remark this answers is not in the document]
fonsecadeline: Same comment here, it is not possible to have this skill from original skills, neither from partition by vehicle twice because partition by vehicle generates cluster with only one vehicle so if we call vehicle partition twice, there will be no clustering second time.

lib/interpreters/split_clustering.rb
```diff
@@ -1203,15 +1207,17 @@ def add_corresponding_entity_skills(entity, vrp)
...
           }
-          vrp.vehicles.first.skills.first << vrp.vehicles.first.id
         when :work_day
           vehicle_id_in_skills = vrp.services.any?{ |s| s.skills.include?(vrp.vehicles.first.id) }
           cluster_day = (vrp.vehicles.first.timewindow || vrp.vehicles.first.sequence_timewindows.first).day_index
-          day_skill = %w[mon tue wed thu fri sat sun][cluster_day]
+          day_skill = OptimizerWrapper::WEEKDAYS[cluster_day]
           vrp.services.each{ |service|
-            service.skills.insert(vehicle_id_in_skills ? 1 : 0, day_skill)
+            service.skills.insert(vehicle_id_in_skills ? 1 : 0, "work_day_partition_#{day_skill}".to_sym)
```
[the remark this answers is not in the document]
fonsecadeline: Me too :p

lib/output_helper.rb
```diff
@@ -90,6 +90,190 @@ def self.csv_line(vrp, service, cluster_index, two_stages = false)
...
+      @work_day_partitioned = true
+      result[:geojsons][0][:partitions][:work_day] = draw_cluster(elements, :work_day)
+    end
+
+    def self.draw_cluster(elements, entity)
+      polygons = elements.group_by{ |element|
+        entity == :vehicle ?
+                  [element[:detail][:internal_skills].find{ |sk| sk.to_s.include?('vehicle_partition_') }] :
+                  [element[:detail][:internal_skills].find{ |sk| sk.to_s.include?('work_day_partition_') },
+                   element[:detail][:skills].find{ |sk| sk.include?('cluster ') }]
```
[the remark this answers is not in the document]
fonsecadeline: We need _find_ because internal skills contain all service skills. Therefore we can not group_by skills/internal_skills but we need to select interesting skills.

We had decided that internal skills are original_skills were we add more data (instead of generating specific structure for clustering skills).

lib/output_helper.rb
```diff
@@ -90,6 +90,190 @@ def self.csv_line(vrp, service, cluster_index, two_stages = false)
...
+
+    def self.find_color(vehicle_id, day, index)
+      if @work_day_partitioned
+        @cluster_colors["work_day_partition_#{OptimizerWrapper::WEEKDAYS[day % 7]}"]
+      elsif @vehicle_partitioned
+        @cluster_colors["vehicle_partition_#{vehicle_id}"]
+      else
+        @colors[index % @colors.size]
+      end
+    end
```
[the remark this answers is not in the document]
fonsecadeline: _@colors_ and_@cluster_colors_ are initialized in _generate_geometry_ function. I made every other function private, like this we make sure that colors are initialized before calling find_color function.

models/concerns/expand_data.rb
```diff
@@ -90,4 +90,10 @@ def provide_original_ids
       element.original_id = element.id
     }
   end
+
+  def provide_original_skills
+    (self.services + self.shipments).each{ |element|
+      element.original_skills = element.skills
```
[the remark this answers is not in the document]
fonsecadeline: It does work like this. I modified _test_returned_skills_ to double check this.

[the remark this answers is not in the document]
fonsecadeline: vehicle skills are not returned but like you say, it does not hurt.

wrappers/wrapper.rb
fonsecadeline: No because we want to return skills corresponding to clusters.

wrappers/wrapper.rb
fonsecadeline: we already have _build_rest_, _build_timewindows_ ... called within _build_details_.

optimizer_wrapper.rb
```diff
@@ -612,26 +619,15 @@ def self.compute_route_total_dimensions(vrp, route, matrix)
...
     route[:activities].each{ |activity|
-      # TODO: This next operation is expensive for big instances. Is there a better way?
-      point_id = activity[:point_id]
-      point_id ||= if activity[:service_id]
-        vrp.services.find{ |s| s.id == activity[:service_id] }.activity.point_id
-      elsif activity[:pickup_shipment_id]
-        vrp.shipments.find{ |s| s.id == activity[:pickup_shipment_id] }.pickup.point_id
-      elsif activity[:delivery_shipment_id]
-        vrp.shipments.find{ |s| s.id == activity[:delivery_shipment_id] }.pickup.point_id
+      point = vrp.points.find{ |p| p.id == activity[:point_id] }&.matrix_index
```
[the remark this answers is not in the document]
fonsecadeline: This is a route[:activity], it has no point for now

optimizer_wrapper.rb
```diff
@@ -709,63 +705,73 @@ def self.route_details(vrp, route, vehicle)
...
-        current = previous
-      end
-      segment = if previous && current
-        [current[:location][:lat], current[:location][:lon], previous[:location][:lat], previous[:location][:lon]]
-      end
+      current =
+        if activity[:rest_id]
+          previous
+        else
+          vrp.points.find{ |point| point.id == activity[:point_id] }
```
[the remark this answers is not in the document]
fonsecadeline: this is a route[:activitiy], it has no point

test/models/vrp_test.rb
fonsecadeline: Geometry is not a new field, I changed behaviour because previously if _geometry_ was equivalent to _geometry_polylines_. It is therefore useless to keep both.

Users used to send { geometry: true, geometry_polylines: true } to get what we now call { geometry: [polylines] }

lib/interpreters/split_clustering.rb
```diff
@@ -790,16 +790,20 @@ def self.collect_cluster_data(vrp, nb_clusters)
       vehicles
     end
 
+    def self.provide_work_day_skill(vehicle, day)
+      vehicle.skills.each{ |sk_set|
+        sk_set << "work_day_partition_#{OptimizerWrapper::WEEKDAYS[day]}".to_sym
+      }
+    end
```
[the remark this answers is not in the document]
fonsecadeline: I did it in this PR : https://github.com/Mapotempo/optimizer-api/pull/165/files#diff-1eed82c376c276daf88072fb9cf8ee2337419c79922b8a9ff8e0857d6f23e55dR259-R272

optimizer_wrapper.rb
```diff
@@ -709,63 +708,71 @@ def self.route_details(vrp, route, vehicle)
...
 
-  def self.parse_result(vrp, result)
-    tic_parse_result = Time.now
-    result[:routes].each{ |r|
-      details = nil
-      v = vrp.vehicles.find{ |vehicle| vehicle.id == r[:vehicle_id] }
-      if r[:end_time] && r[:start_time]
-        r[:total_time] = r[:end_time] - r[:start_time]
-      end
+  def self.compute_route_travel_times(vrp, matrix, route, vehicle)
```
braktar, line 737: 'travel_times' implies to compute something relative to 'time' not a 'distance' ?

braktar (changes requested):

optimizer_wrapper.rb
fonsecadeline: Indeed..!

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@
 
 - Corresponding vehicle_id is returned within each service's skills if problem is partitioned with vehicle entity [#110](https://github.com/Mapotempo/optimizer-api/pull/110)
 - Support initial routes and skills in split_solve (`max_split_size`) algorithm [#140](https://github.com/Mapotempo/optimizer-api/pull/140)
+- Return `geojsons` structure according to geometry parameter [#165](https://github.com/Mapotempo/optimizer-api/pull/165)
```
fonsecadeline, line 9:
```suggestion
- Return `geojsons` structure according to `geometry` parameter [#165](https://github.com/Mapotempo/optimizer-api/pull/165)
```

wrappers/wrapper.rb
senhalil: Oh in that case why don't we return the skills with the internal masking? Don't we want to make sure that the cluster skills are separate from non-clustering skills at the output.

If we remove the maskings `work_day_partition_X` and return `X`, if the user has skill `X` then it will clash with that. I don't think we have a verification for that. Then the user cannot know which is skill is due to clustering and which skill is due to their own skills.

senhalil (changes requested):

models/types/geometry_type.rb
```diff
@@ -0,0 +1,44 @@
...
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+# or FITNESS FOR A PARTICULAR PURPOSE.  See the Licenses for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with Mapotempo. If not, see:
+# <http://www.gnu.org/licenses/agpl.html>
+#
+
+class GeometryType
+  ALL_TYPES = %i[polylines partitions].freeze
```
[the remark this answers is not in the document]
fonsecadeline: We decided that we can always return points because it is not costly (and in fact front page would look weird if we do not return points at least). 

We can change this in a second version, while adjusting with front behaviour but I think it is okay to always return points.

test/models/vrp_test.rb
fonsecadeline: You are right. I started like this and then changed it because it was dirty but you are right, we need to know if user is mixing old and new parameters.

This implies : that _geometry_ can be either an array of symbol, or a boolean.
If it is false we just return _geometry_ = [], otherwise we need to keep _true_ value until we get to vrp.create and detect if we used _geometry_polylines_ or not. 

Do you see a better way of doing this ?

wrappers/wrapper.rb
fonsecadeline: I am only returning _X_ because that was one of our user's need, to have the exact id of corresponding vehicle.

It seems like we wanted to make it easier, returning only _skills_ but we are making it harder. We should have _clustering_skills_ for every service and shipment, as we initially said (and add a new column with vehicle_cluster skills and day_cluster skills specifically in the CSV. 

@senhalil @braktar What do you think ?  Are we okay to increase number of attributes within service/shipment model ? I think it would simplify the code. But then we discussed about using internal skills for sticky_vehicles and that is why we decided to make everythink in _skills_.

fonsecadeline: https://gitlab.com/mapotempo/optimizer-api/-/issues/697

models/concerns/expand_data.rb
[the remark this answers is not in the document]
fonsecadeline: I was talking about _test_returned_skills_.

In fact since today skills are only modified after clustering then we do Marshal.load and there is no problem. As you suggested I added .dup to avoid any error in the future if we edit skills without using Marshal.load.

models/vrp.rb
```diff
@@ -341,12 +356,20 @@ def self.convert_route_indice_into_index(hash)
       }
     end
 
+    def self.adjust_geometry_according_to_geometry_polylines(hash)
+      return unless hash[:configuration] && hash[:configuration][:restitution] && hash[:configuration][:restitution][:geometry_polyline] == true
+
+      hash[:configuration][:restitution][:geometry] = %i[polylines]
```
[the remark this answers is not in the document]
fonsecadeline: Related to https://github.com/Mapotempo/optimizer-api/pull/165/files#r606205605

api/v01/entities/vrp_input.rb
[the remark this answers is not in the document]
fonsecadeline: Since we expect Array[Symbol] it did not work and I had to remove those values. I will allow boolean also (https://github.com/Mapotempo/optimizer-api/pull/165/files#r606205605) and we should be able to use it.

api/v01/entities/vrp_result.rb
fonsecadeline: In fact partitions were initially documented as being an array, which was a mistake. I think after discussing in one of our daily @braktar was okay with this.

geojsons are array to allow us to return one geojsons structure per repetition or per method.

(this comment appeared on wrong thread apparently)

lib/output_helper.rb
```diff
@@ -90,6 +90,190 @@ def self.csv_line(vrp, service, cluster_index, two_stages = false)
...
+
+      result[:geojsons][0][:partitions] ||= {}
+      if has_vehicle_partition
+        @vehicle_partitioned = true
+        result[:geojsons][0][:partitions][:vehicle] = draw_cluster(elements, :vehicle)
+      end
+
+      return unless has_work_day_partition
+
+      @work_day_partitioned = true
```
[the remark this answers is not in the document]
fonsecadeline: We can remove _find_color_ function but we still need to know which partitioning was used in order to colorized regarding work_days or regarding vehicles.

test/models/vrp_test.rb
senhalil: No, not really. It'll be again `ensure_retrocompatibility`

wrappers/wrapper.rb
senhalil: By the way do we know if they will not be okay with  `vehicle_cluster_vehicleid`? Maybe we are making it harder for no reason. These partition skills are internally generated so it is quite normal that they would have internal masks. 

If we decide to have separate columns in the CSV, why do we need clustering skills for the internal code? We can keep the current code (and not remove the skill mask for cluster skills) and CSV parser can parse `vehicle_cluster_X` into a separate column :thinking:

lib/output_helper.rb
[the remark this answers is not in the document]
senhalil: I think the function is fine, I have trouble understanding the need for @work_day_partitioned and @vehicle_partitioned class variables. 

Does it really matter if we had vehicle partition or not before creating the output for work_day partition? Can't we already have one color per day by default in any case?

wrappers/wrapper.rb
fonsecadeline: We do know they can not work with _ vehicle_cluster_vehicleid_, it was like that and they still had troubles with it :/

We first did not want to return specific column for clustering but I guess that would not hurt. It is strange to mix internal and original skills.

models/vrp.rb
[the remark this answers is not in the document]
senhalil: I resolve this and lets follow up on the other one then

api/v01/entities/vrp_input.rb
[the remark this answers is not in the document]
fonsecadeline: In fact it is tricky to use Boolean here : 

- if I use 
```suggestion
    optional(:geometry, type: [Array[Symbol], Boolean], default: [], coerce_with: ->(value) { GeometryType.type_cast(value) },
```
then it does not call coerce function when value is false. I think I need to use current line and, when value is true, return like [:to_fill]. In this case I can use values but can not output _to_fill_ in the documentation.

lib/output_helper.rb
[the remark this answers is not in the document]
fonsecadeline: After trying to convince you, you convinced me that we can always color with day color.

The idea was to make it easier to see vehicle partitions when there was no work_day partitioning. But then users can only output vehicle clusters polygons :)

braktar (approved):

braktar (approved):

models/types/geometry_type.rb
```diff
@@ -0,0 +1,48 @@
...
+
+    if value.is_a?(FalseClass)
+      []
+    elsif value.is_a?(TrueClass)
+      ALL_TYPES
+    elsif value.is_a?(Array)
+      to_return = []
+      value.each{ |geometry_type|
+        unless ALL_TYPES.include?(geometry_type.to_sym)
+          raise ArgumentError.new("Invalid geometry value : #{geometry_type}")
```
senhalil, line 33:
```suggestion
          raise ArgumentError.new("Invalid geometry value: #{geometry_type}")
```

models/types/geometry_type.rb
```diff
@@ -0,0 +1,48 @@
...
+        end
+
+        to_return << geometry_type.to_sym
+      }
+
+      if (to_return & [:polylines, :encoded_polylines]).size == 2
+        raise ArgumentError.new('Invalid type for geometry value : polylines and encoded polylines are not compatible')
```
senhalil, line 40:
```suggestion
        raise ArgumentError.new('Invalid geometry value: polylines and encoded_polylines options are not compatible')
```

optimizer_wrapper.rb
```diff
@@ -98,22 +99,26 @@ def self.wrapper_vrp(api_key, services, vrp, checksum, job_id = nil)
 
     if services_vrps.any?{ |sv| !sv[:service] }
       raise UnsupportedProblemError.new('Cannot apply any of the solver services', inapplicable_services)
-    elsif vrp.restitution_geometry && !vrp.points.all?{ |point| point[:location] }
-      raise DiscordantProblemError, 'Geometry is not available if locations are not defined'
+    elsif vrp.restitution_geometry.any? && !vrp.points.all?(&:location)
+      raise DiscordantProblemError.new('Geometry is not available if locations are not defined')
```
senhalil, line 103: Can't we move this to the check_consistency? It seems out of order here. We can also change the error type to an ArgumentError because we don't know if there is an issue with the problem.

optimizer_wrapper.rb
```diff
@@ -98,22 +99,26 @@ def self.wrapper_vrp(api_key, services, vrp, checksum, job_id = nil)
...
-        JobList.add(api_key, job_id)
-        job_id
-      end
+      # Delegate the job to a worker
+      job_id = Job.enqueue_to(services[:queue], Job, services_vrps: Base64.encode64(Marshal.dump(services_vrps)),
+                                                     api_key: api_key,
+                                                     checksum: checksum,
+                                                     pids: [],
+                                                     configuration: { csv: services_vrps.any?{ |s| s[:vrp].restitution_csv },
+                                                                      geometry: [] })
```
senhalil, line 119: Do we need this `configuration:` hash in job?

senhalil (commented):
few questions

optimizer_wrapper.rb
fonsecadeline: One test failed without it, but it does not anymore. I can remove this eventually, test has probably been fixed when we modified call to OutputHelper::Result.generate_geometry(solution) which now handles correctly when solution is empty (and hence no configuration is provided).

optimizer_wrapper.rb
senhalil: Sure, if we have that logic, you can do it like that.

I was under the impression that we return DiscordantProblemError when the optimisation problem itself has "disaggreeing/discordant" params (like, visits_number > schedule_range[end]-[begin], because the problem has a disagreement in it) but when the issue is due to a config like this then it is a misuse of an Argument  -- i.e., asking a geometry without locations. I might be wrong, in any case, there still is a disaccord/discordance so it is okay.

models/vrp.rb
```diff
@@ -227,6 +228,14 @@ def self.check_consistency(hash)
         end
       end
 
+      if configuration[:restitution]
+        if configuration[:restitution][:geometry].to_a.any? &&
```
senhalil, line 232: do we need `to_a` here? isn't geometry an array by default -- can it be nil?

senhalil (approved):

models/vrp.rb
fonsecadeline: It is sometimes in the tests. I can deal with this case in test_helper maybe.

senhalil: yeah, test_helper coerce might be better suited for this if we have tests which have restitution but no geometry in them.

fonsecadeline: https://github.com/Mapotempo/optimizer-api/pull/165/commits/8dd8ebc18be97feb092c2b84697bc632837ceda9

PR #166, opened 2021-03-26
Fix test timeout and and PR managment

alain-andre:


fonsecadeline (approved):

fonsecadeline:
Will this be enough to prevent basis to run undefinetely AND provide results from basis ? Won't this just cut basis, and we will not know if tests are ok or not?

senhalil (approved):

alain-andre:
> Will this be enough to prevent basis to run undefinetely AND provide results from basis ? Won't this just cut basis, and we will not know if tests are ok or not?

@fonsecadeline this is based on passed runs; basis takes like 6 minutes normally. If the script is not ended after 10 minutes this will trigger an error

fonsecadeline:
> > Will this be enough to prevent basis to run undefinetely AND provide results from basis ? Won't this just cut basis, and we will not know if tests are ok or not?
> 
> @fonsecadeline this is based on passed runs; basis takes like 6 minutes normally. If the script is not ended after 10 minutes this will trigger an error

But then we might get an error for every run. Because this seems to be running undefinetely. We need to understand why to avoid this error.

alain-andre:
> > > Will this be enough to prevent basis to run undefinetely AND provide results from basis ? Won't this just cut basis, and we will not know if tests are ok or not?
> > 
> > 
> > @fonsecadeline this is based on passed runs; basis takes like 6 minutes normally. If the script is not ended after 10 minutes this will trigger an error
> 
> But then we might get an error for every run. Because this seems to be running undefinetely. We need to understand why to avoid this error.

I agree we need to understand why it runs indefinitely to fix the problem, but this PR is not about fixing this error which seems random (as on this PR tests passed). This PR allows to save time on indefinitely running tests and to show results in PR. Nor does it solve the router access error in test_split_clustering.

braktar (approved):

PR #168, opened 2021-03-30
Warn instead for raising when date used with day indices

fonsecadeline:
Temporary modification in order to let our users time to adapt their process.


fonsecadeline:
This may not be necessary anymore. I sent an email to our user to check.

PR #169, opened 2021-03-31
Avoid loop in failing test

fonsecadeline:


test/api/v01/request_helper.rb
```diff
@@ -118,6 +119,7 @@ def asynchronously(options = {start_worker: false})
     OptimizerWrapper.config[:solve][:synchronously] = false
     old_resque_inline = Resque.inline
     Resque.inline = false
+    sleep 0.5
```
senhalil, line 122: why do we need an extra sleep here?

braktar, line 122: It has been removed once by @senhalil . Do you remember why ?

test/api/v01/request_helper.rb
```diff
@@ -26,6 +26,7 @@ def wait_status(job_id, status, options)
       get "0.1/vrp/jobs/#{job_id}.json", options
 
       assert_equal 200, last_response.status, last_response.body
+      assert_includes ['queued', 'working', 'completed'], JSON.parse(last_response.body)['job']['status']
```
senhalil, line 29: Can we use a similar check in `wait_status_csv` or is there no risk there?

fonsecadeline: You are right, we can add it in `wait_status_csv` too.

test/api/v01/request_helper.rb
fonsecadeline: Maybe this is not the best solution. Some tests fail because of "job_id not found". I run some (small) tests and when adding a sleep here we reduce number of times this occurs : 
Without sleep            test fails 8/11 times because of this, 
With sleep                 test fails 3/11 times because of this.

I assumed that we try to get result too fast because it has not arrived to worker yet. But now that I look a little closer to this, I guess sleep should rather be in wait status.

fonsecadeline: In fact we do have one in `wait_status_csv` but not in `wait_status`

senhalil: I specifically looked into the sleep problem not long ago. The sleep shouldn't be needed. 

If you are doing this because of those tests with `job_id=xx not found` problem, they are related to a "zombie" worker at the background, if you check the xx job_id that can't be found, that job is already launched by another "zombie" worker and then "cleaned" by the worker who supposed to do the job.

If this is because of another test, lets find the real culprit, sleep is just hiding the issue.

senhalil: because it's not necessary. It was also me (I believe) who added it in the first place.

senhalil: Back then we had race condition in the public API, so it was a way to avoid them if my memory serves well.

fonsecadeline: Thank you, I wasn't really sure that was needed.

test/api/v01/request_helper.rb
```diff
@@ -43,6 +44,9 @@ def wait_status_csv(job_id, status, options)
       get "0.1/vrp/jobs/#{job_id}", options
 
       assert_equal 200, last_response.status, last_response.body
+      unless last_response.body.is_a?(String)
+        assert_includes ['queued', 'working', 'completed'], JSON.parse(last_response.body)['job']['status']
+      end
 
       break if last_response.content_type == 'text/csv;'
```
senhalil, line 51: would this work? I think if the content is not `text/csv` then we should be able to parse the response without any issue
```suggestion

      break if last_response.content_type == 'text/csv;'

      assert_includes ['queued', 'working', 'completed'], JSON.parse(last_response.body)['job']['status']
```

senhalil (approved):

PR #174, opened 2021-04-06
Rename scheduling to periodic

fonsecadeline:


api/v01/entities/vrp_input.rb
```diff
@@ -176,11 +176,11 @@ module VrpConfiguration
...
-    optional(:minimize_days_worked, type: Boolean, default: false, desc: '(Scheduling heuristic only) Starts filling earlier days of the period first and minimizes the total number of days worked. Available only if first_solution_strategy is \'periodic\'. Not available with ORtools.')
-    optional(:same_point_day, type: Boolean, desc: '(Scheduling only) Forces all services with the same point_id to take place on the same days. Available only if first_solution_strategy is \'periodic\'. Not available ORtools.')
-    optional(:allow_partial_assignment, type: Boolean, default: true, desc: '(Scheduling heuristic only) Considers a solution as valid even if only a subset of the visits of a service is performed. If disabled, a service can only appear fully assigned or fully unassigned in the solution. Not available with ORtools.')
+    optional(:minimize_days_worked, type: Boolean, default: false, desc: '(Periodic heuristic only) Starts filling earlier days of the period first and minimizes the total number of days worked. Available only if first_solution_strategy is \'periodic\'. Not available with ORtools.')
```
senhalil, line 179: I think `Periodic heuristic only` is more expressive but in the remaining we always use `Periodic only`. Which one do we prefer `Periodic heuristic only` or `Periodic only`? 

Since we are renaming all, may be we can do the rest `Periodic heuristic only` as well?

senhalil:
@fonsecadeline can you merge all the commits of this PR so that renaming is done in one single commit?

api/v01/entities/vrp_input.rb
fonsecadeline: `Periodic only` and `Periodic heuristic only` have a different meaning. 

`Periodic only`means only if you have a schedule. `Periodic heuristic` means only if you have a schedule and you specifically call heuristic.

fonsecadeline: I guess if we have _schedule_ but all service have one visit then we should rather say `Scheduling only`. I am not sure if our customers will understand the difference.

senhalil: I understand better now. I am not sure the difference `Periodic only` and `Periodic heuristic only` is clear enough but I don't have a better suggestion :/ 

Are we renaming `config[:schedule]` into `config[:periodic]` as well? If that is not the case then `Schedule only` should stay as it is, no? 

If the parameter is called Schedule and the heuristic is called Periodic heuristic then  `Periodic heuristic only` and  `Schedule only` are intuitive for me but I am afraid `Periodic only` might be confusing. What do you think @fonsecadeline?

fonsecadeline: You are right ! I guess we can use _Periodic heurisitc_ and _Schedule_only_.
_Periodic only_ is very specific and can be replaced by _Schedule_only_ because in both cases we call ORtools with schedule (only difference is that _Periodic only_ implies that services have more than 1 visit).

lib/heuristics/periodic_heuristic.rb
```diff
@@ -298,8 +305,10 @@ def update_route(route_data, first_index, first_start = nil)
         previous_end = stop[:end]
       }
 
-      raise OptimizerWrapper::SchedulingHeuristicError, 'Vehicle end violated after updating route' if route.size.positive? &&
-                                                                                                       route.last[:end] + matrix(route_data, route.last[:point_id], route_data[:end_point_id]) > route_data[:tw_end]
+      if route.size.positive? &&
```
braktar, line 308:
```suggestion
      if route.any? &&
```

braktar (changes requested):

util/job_manager.rb
```diff
@@ -112,7 +112,7 @@ class UnsupportedProblemError     < StandardErrorWithData; end
   class ClusteringError             < StandardError; end
   class DiscordantProblemError      < StandardError; end
   class JobKilledError              < StandardError; end
-  class SchedulingHeuristicError    < StandardError; end
+  class PeriodicHeuristicError    < StandardError; end
```
braktar, line 115: Could you align the `<` with the other lines ?

braktar (changes requested):

braktar (approved):

senhalil (approved):

fonsecadeline:
- [ ] https://github.com/Mapotempo/optimizer-api/pull/196 needs to be merged first

fonsecadeline:
@braktar @senhalil should I mention in CHANGELOG that we renamed scheduling heuristic to periodic heuristic ?

senhalil:
> @braktar @senhalil should I mention in CHANGELOG that we renamed scheduling heuristic to periodic heuristic ?

I don't know, the heuristic was always called periodic on the user side, so there is no change from their point of view. From the dev point of view, it doesn't sound like it deserves to be on the changelog but either way is fine for me.

senhalil (approved):

PR #176, opened 2021-04-07
Fix in find_best_heuristic

fonsecadeline:
I am not sure we do need to test this.

### DIFF
--- a/lib/interpreters/compute_several_solutions.rb
+++ b/lib/interpreters/compute_several_solutions.rb
@@ -191,7 +191,7 @@ module Interpreters
           heuristic_solution
         }
 
-        raise RuntimeError, 'No solution found during heuristic selection' if first_results.all?(&:nil?)
+        raise 'No solution found during heuristic selection' if first_results.all?(&:nil?)
 
         synthesis = []
         first_results.each_with_index{ |result, i|
@@ -214,7 +214,7 @@ module Interpreters
             }.compact
             next if mission_ids.empty?
 
-            Models::Route.new(vehicle: vrp.vehicles.find{ |v| v[:id] = route[:vehicle_id] }, mission_ids: mission_ids)
+            Models::Route.new(vehicle: vrp.vehicles.find{ |v| v[:id] == route[:vehicle_id] }, mission_ids: mission_ids)
           }.compact
         end
 
--- a/test/wrappers/ortools_test.rb
+++ b/test/wrappers/ortools_test.rb
@@ -5777,4 +5777,39 @@ class Wrappers::OrtoolsTest < Minitest::Test
   def test_no_nil_in_corresponding_mission_ids
     assert_empty OptimizerWrapper.config[:services][:ortools].send(:corresponding_mission_ids, ['only_id'], ['non_id'])
   end
+
+  def test_self_selection_should_not_change_vehicle_ids
+    vrp = VRP.independent_skills
+    vrp[:configuration][:preprocessing] = { first_solution_strategy: ['self_selection'] }
+
+    heuristic_counter = 0
+    OptimizerWrapper.stub(
+      :solve,
+      lambda{ |_|
+        heuristic_counter += 1
+        if heuristic_counter == 1
+          { cost: 27648.0, solvers: ['ortools'], elapsed: 0.94, routes: [{ vehicle_id: 'vehicle_0', activities: [] }, { vehicle_id: 'vehicle_1', activities: [] }, { vehicle_id: 'vehicle_2', activities: [] }] }
+        else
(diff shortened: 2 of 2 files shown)


braktar (approved):

lib/interpreters/compute_several_solutions.rb
```diff
@@ -214,7 +214,7 @@ def self.find_best_heuristic(service_vrp)
             }.compact
             next if mission_ids.empty?
 
-            Models::Route.new(vehicle: vrp.vehicles.find{ |v| v[:id] = route[:vehicle_id] }, mission_ids: mission_ids)
+            Models::Route.new(vehicle: vrp.vehicles.find{ |v| v[:id] == route[:vehicle_id] }, mission_ids: mission_ids)
```
senhalil, line 217: :face_with_head_bandage:  this is the second time I introduced this exact same error and it ended up on beta :(

test/wrappers/ortools_test.rb
```diff
@@ -5777,4 +5777,26 @@ def test_pause_should_be_last_if_possible
...
+    vrp[:configuration][:preprocessing] = { first_solution_strategy: ['self_selection'] }
+
+    Interpreters::SeveralSolutions.stub(
+      :collect_heuristics,
+      lambda { |_, _|
+        %w[parallel_cheapest_insertion global_cheapest_arc local_cheapest_insertion savings]
+      }
+    ) do
+      service_vrp = Interpreters::SeveralSolutions.find_best_heuristic({ service: :ortools, vrp: TestHelper.create(vrp) })
+      assert_equal 3, service_vrp[:vrp].vehicles.collect(&:id).uniq.size, 'We expect same number of IDS as initial vehicles in problem'
```
senhalil, line 5799: Can we also stub the `OptimizerWrapper.solve` and return empty result to skip solution process altogether?

senhalil, line 5799: IDS -> IDs

senhalil (changes requested):

test/wrappers/ortools_test.rb
fonsecadeline: If we return empty result that will not generate the error. We need one non empty route. We could eventually stub and return always same solution but this test needs 0.28s, I don't think we need to spend more time on it.

fonsecadeline:
```suggestion
      assert_equal 3, service_vrp[:vrp].vehicles.collect(&:id).uniq.size, 'We expect same number of IDs as initial vehicles in problem'
```

senhalil: Knowing what we test here, it shouldn't even take 0.01 seconds. I think your idea, returning a fixed solution, is a good idea in fact.

test/wrappers/ortools_test.rb
```diff
@@ -5777,4 +5777,39 @@ def test_pause_should_be_last_if_possible
...
+    ) do
+      Interpreters::SeveralSolutions.stub(
+        :collect_heuristics,
+        lambda{ |_, _|
+          %w[global_cheapest_arc parallel_cheapest_insertion local_cheapest_insertion savings]
+        }
+      ) do
+        service_vrp = Interpreters::SeveralSolutions.find_best_heuristic({ service: :ortools, vrp: TestHelper.create(vrp) })
+        assert_equal vrp[:vehicles].size, service_vrp[:vrp].vehicles.collect(&:id).uniq.size, 'We expect same number of IDS as initial vehicles in problem'
+        assert_equal vrp[:services].size, service_vrp[:vrp].services.collect(&:id).uniq.size, 'We expect same number of IDS as initial services in problem'
```
senhalil, line 5809: find_best_heuristic sets the `route`. Can we also verify that the route is as expected -- like we do with vehicles and services? 

At the moment find_best_heurictic doesn't touch the `service_vrp[:vrp].vehicles` nor `service_vrp[:vrp].services` but we can keep them for good measure.

IDS - IDs

senhalil (changes requested):

test/wrappers/ortools_test.rb
fonsecadeline: find_best_heuristic is not SUPPOSED to touch. But the bug we had is that calling this function changed IDs in service_vrp[:vrp].vehicles. So in fact this is the way to test there is no regression about the bug we've had.

senhalil: That is why I want to also test the structure (`:routes`) it touches. It also affected the `routes` and it can happen without touching the `vehicles` or `services` in the future.

fonsecadeline: Yes, I changed the test already

senhalil (approved):

PR #178, opened 2021-04-08
Return discordant when rejected by create

fonsecadeline:
Fix for sentry https://sentry.mapotempo.com/mapotempo/api-optimizer-beta/issues/15291/?query=is:unresolved

### DIFF
--- a/models/vrp.rb
+++ b/models/vrp.rb
@@ -148,7 +148,7 @@ module Models
         raise OptimizerWrapper::UnsupportedProblemError.new(
           'Services can appear in at most one shipment relation. '\
           'Following services appear in multiple shipment relations',
-          shipment_relations.detect{ |id| shipment_relations.count(id) > 1 }
+          [shipment_relations.select{ |id| shipment_relations.count(id) > 1 }.uniq]
         )
       end
 


models/vrp.rb
```diff
@@ -144,10 +144,10 @@ def self.check_consistency(hash)
       # shipment relation consistency
       shipment_relations = hash[:relations]&.select{ |r| r[:type] == :shipment }&.flat_map{ |r| r[:linked_ids] }.to_a
       unless shipment_relations.size == shipment_relations.uniq.size
-        raise OptimizerWrapper::UnsupportedProblemError.new(
+        raise OptimizerWrapper::DiscordantProblemError.new(
```
senhalil, line 147: This is not a discordant problem. It is a valid VRP, it is just we don't support it. 

The fix is to give the collected ids to the UnsupportedProblemError in an `Array[Array[String]]` form if I am no mistaken.

models/vrp.rb
```diff
@@ -407,6 +407,12 @@ def self.remove_unnecessary_relations(hash)
            vehicle_group_duration_on_months vehicle_group_number]
 
       hash[:relations].delete_if{ |r| r[:lapse].nil? && types_with_duration.include?(r[:type]) }
+
+      uniq_relations = []
+      hash[:relations].group_by{ |r| r[:type] }.each{ |_type, relations_set|
+        uniq_relations += relations_set.uniq
+      }
+      hash[:relations] = uniq_relations
```
senhalil, line 415: what is the goal here, if there is a duplicate relation, then it is an invalid vrp (i.e., DiscordantProblemError)? Like the way, it is invlaid when there is a duplicate service.

senhalil (changes requested):

models/vrp.rb
fonsecadeline: From my point of view if it is twice the exact same relation I do not think we need to reject it because if we can satisfy it once we can satisfy it twice. Result will be consistent with what data provided by user. 

But at the same time it could help user notice he provided 'weird' data.

senhalil: I think your last phrase is more important that is why I think we should raise otherwise the user can't know.

models/vrp.rb
```diff
@@ -147,7 +147,7 @@ def self.check_consistency(hash)
         raise OptimizerWrapper::UnsupportedProblemError.new(
           'Services can appear in at most one shipment relation. '\
           'Following services appear in multiple shipment relations',
-          shipment_relations.detect{ |id| shipment_relations.count(id) > 1 }
+          [[shipment_relations.detect{ |id| shipment_relations.count(id) > 1 }]]
```
senhalil, line 150: I didn't verify this by the way, does it work as expected?

senhalil (changes requested):

models/vrp.rb
fonsecadeline: It does

models/vrp.rb
```diff
@@ -148,7 +148,7 @@ def self.check_consistency(hash)
...
```
fonsecadeline, line 151:
```suggestion
          [[shipment_relations.select{ |id| shipment_relations.count(id) > 1 }]]
```

models/vrp.rb
```diff
@@ -148,7 +148,7 @@ def self.check_consistency(hash)
         raise OptimizerWrapper::UnsupportedProblemError.new(
           'Services can appear in at most one shipment relation. '\
           'Following services appear in multiple shipment relations',
-          shipment_relations.detect{ |id| shipment_relations.count(id) > 1 }
+          [[shipment_relations.select{ |id| shipment_relations.count(id) > 1 }.uniq]]
```
fonsecadeline, line 151:
```suggestion
          [shipment_relations.select{ |id| shipment_relations.count(id) > 1 }.uniq]
```

senhalil (approved):

PR #179, opened 2021-04-09
Parse intermediate results

fonsecadeline:


wrappers/ortools.rb
```diff
@@ -514,8 +514,7 @@ def parse_output(vrp, _services, points, _matrix_indices, _cost, _iterations, ou
 
       collected_indices = []
       vehicle_rest_ids = Hash.new([])
-      solution_cost = Models::CostDetails.new({})
-      {
+      OptimizerWrapper.parse_result(vrp, {
```
braktar, line 517: I would prefer to put the hash within a variable and then transmit it to the method : `OptimizerWrapper.parse_result(...)`

Having a multi-line argument is pretty hard to understand

braktar (changes requested):

braktar (approved):

senhalil (approved):

PR #181, opened 2021-04-09
Move consistency checkers to a class

fonsecadeline:
VRP validation was getting too long and mixed. I think we had discussed doing this in the meantime of [I do not remember what].

In any case that would help to have a cleaner base for future developments

### DIFF
--- a/models/vrp.rb
+++ b/models/vrp.rb
@@ -18,12 +18,14 @@
 require './lib/tsp_helper.rb'
 require './models/base'
 require './models/concerns/distance_matrix'
+require './models/concerns/validate_data'
 require './models/concerns/expand_data'
 require './models/concerns/periodic_service'
 
 module Models
   class Vrp < Base
     include DistanceMatrix
+    include ValidateData
     include ExpandData
     include PeriodicService
 
@@ -124,7 +126,7 @@ module Models
       vrp = super({})
       self.filter(hash) # TODO : add filters.rb here
       # moved filter here to make sure we do have schedule_indices (not date) to do work_day check with lapses
-      self.check_consistency(hash)
+      vrp.check_consistency(hash)
       self.ensure_retrocompatibility(hash)
       [:name, :matrices, :units, :points, :rests, :zones, :capacities, :quantities, :timewindows,
        :vehicles, :services, :shipments, :relations, :subtours, :routes, :configuration].each{ |key|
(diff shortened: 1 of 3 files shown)


models/hash_validator.rb
```diff
@@ -0,0 +1,273 @@
...
+# or FITNESS FOR A PARTICULAR PURPOSE.  See the Licenses for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with Mapotempo. If not, see:
+# <http://www.gnu.org/licenses/agpl.html>
+#
+
+require './models/vrp'
+
+module Models
```
senhalil, line 21: Do we need this to be in the Models module? There is only real models in this module.

models/hash_validator.rb
```diff
@@ -0,0 +1,273 @@
...
+  class HashValidator < Vrp
```
senhalil, line 22: Why do we need to inherit from VRP class to have a Hash validator?

models/hash_validator.rb
```diff
@@ -0,0 +1,273 @@
...
+    def initialize(hash, internal_call)
+      @hash = hash
+      @internal_call = internal_call
+    end
+
+    def check_hash
```
senhalil, line 28: Do we really need a class with initialiser and a function? We can have the initialiser do the work and be done with it.

models/hash_validator.rb
```diff
@@ -0,0 +1,273 @@
...
+# <http://www.gnu.org/licenses/agpl.html>
+#
+
+require './models/vrp'
+
+module Models
+  class HashValidator < Vrp
+    def initialize(hash, internal_call)
+      @hash = hash
+      @internal_call = internal_call
```
senhalil, line 25: I don't think we should have a different behaviour for internal_calls. It is dangerous, we might be able to create invalid VRPs in the code if the validators is not the same.

fonsecadeline: That is right. But then we need to edit _max_split_ behaviour and/or validation regarding number of visits :

- In vrp.create we should reject VRP if service has more than one visit but no schedule. Indeed, if we have no schedule then we do not expand service visits so user will have an inconsistent result.
- In new _max_split_ we edit number of services visits (if I remember well it is in the case were several services have same lat/lon then we generate one service with several visits) but we have no schedule. 

We need to find a way to use _visits_number_ in a more generic way to be able to deal with both cases.

models/hash_validator.rb
fonsecadeline: I felt it could be weird to 'initialize' a validator and in fact it has already validated.

But I agree that we are always going to call them consecutively.

models/hash_validator.rb
fonsecadeline: I replaced _models/hash_validator.rb_ by _lib/validators.rb_ which contains _Validators_ module.
What do you think ?

models/hash_validator.rb
fonsecadeline: I think I forgot to remove it :/

models/hash_validator.rb
senhalil: In fact, I ask myself if we need a class for this logic. Because it is a hash anyways (i.e., the hash has no methods coming from its class that we call). Even though I don't like concerns, we already have several of them, we can have this logic in a "concerned" function without any class. What do you think @fonsecadeline, wouldn't it be more simple like that ?

fonsecadeline: Concern works for me !

models/hash_validator.rb
senhalil: - Oh, I thought if we have no `:schedule`, we just ignore the visits_number. If that is the case I don't see how the new split_solve worked in the first place :confused: 
- I think I understand where you mean. Do you mean [here](https://github.com/Mapotempo/optimizer-api/blob/8ae741327b2a4559eecde0f1bef9f607929a3be5/lib/interpreters/split_clustering.rb#L349) where we represent all services coming from the same vehicle "group" with a single point ? This code used to work with the old validators even though we didn't have a separate logic for internal calls, did we change the validators?

In any case, it looks like my usage of `:visits_number` is not completely "appropriate" and I am not happy that we modify the validators because of it.

I can only think of modifying the split_solve so that it first creates the representative_vrp and then sets the visits_number for the visits (which is still an inappropriate usage but at least we don't need to modify the validators because of it). Can you think of another way to avoid this?

models/hash_validator.rb
fonsecadeline: The only thing is that if we use concern then we need to call it with something like 
```vrp.check_consistency``` 
whereas we want to check consistency on _hash_, not on _vrp_. This might be confusing don't you think ? Am I missing something?

File: models/hash_validator.rb (hunk 4 above)
Comment (fonsecadeline) [reply models/hash_validator.rb]: You are right, this check (there can not be more than one visit without a schedule) is new, that is why it used to work. I guess I added this check in parallel of your changes of split_solve, that is why we never saw the problem. 

As you said, visits_number used to be ignored if no schedule was provided. Maybe it is okay to keep it like that. Documentation about _visits_number_ specifies `Scheduling only`. This check could be helpful for users but no one ever complained about this. Maybe the easier would be to simply remove the check ?

Comment (senhalil) [reply models/hash_validator.rb]: Make sense :+1: 

I don't know. In that case we ignore a parameter and the user might not aware of that they made an error :/ or  (I doubt it but) one day we might want to have a "meaning" for `visits_number sans scheduling` and the behaviour would change for json with visits_number without scheduling. I think it's better if you keep the check and remove the internal_call logic and I will fix the split_solve calls (your PR will have a couple of broken tests but I think we can live with that for few days :crossed_fingers:)

File: models/hash_validator.rb (hunk 3 above)
Comment (senhalil) [reply models/hash_validator.rb]: Since vrp.create(hash) is also called with an hash, it doesn't shock me if vrp.check_consistency(hash) is called with an hash to do the vrp validation.

File: models/vrp.rb [hunk 5]
```diff
@@ -124,7 +126,7 @@ def self.create(hash, delete = true)
       vrp = super({})
       self.filter(hash) # TODO : add filters.rb here
       # moved filter here to make sure we do have schedule_indices (not date) to do work_day check with lapses
-      self.check_consistency(hash)
```
braktar, line 127: It doesn't really concern the VRP, but the hash. I am not convinced at all that it should be a concern.

fonsecadeline:
Split_clustering_tests fail because we added a check : there can not be more than one visit if there is no schedule. Visits are considered a little differently in max_split algo so 7 tests fail now.

fonsecadeline:
Real scheduling cases should be okay after we merge https://github.com/Mapotempo/optimizer-api/pull/191

fonsecadeline:
- [ ] needs https://gitlab.com/mapotempo/optimizer-api/-/issues/767

models/vrp.rb
fonsecadeline: What would you suggest ? We had previously created a class that validates Hash.

braktar: We have in input a JSON received by grape, which converts it into an hash, this hash is then converted into a vrp model using active_hash.

My assumption is that, the intermediate hash should have is own class which inherit from hash and add the concern to it.
so we could directly call hash.check_consistency

senhalil: @fonsecadeline the class you proposed was completely different -- it was in Models module and inhereted from Vrp class, -- the class @braktar suggests is a lot straightforward and it is kind of a wrapper around `Hash` class so that you can add the concern to that class.

I am still not convinced that it worth the effort because the main reason we have these validations like this (i.e., manually) is that the validation functionality `validates_x_of_y` of active_hash(?) was broken in the previous Ruby version. We still haven't tested if it is still broken or not in the current Ruby version. This functionality also lets us define custom validators and it calls them automatically, and (if I am not mistaken) it gives clear fail messages with indexing etc. so our manual validators are temporary at best. 

This would make more sense if we could use this custom Hash to tell Grape to use as validator (kinda like presenter + validation) but I don't think we can do that.

fonsecadeline: I agree that this is meant to be temporary and maybe we do not need to spend too much time trying to make it in the best way.

What should I do then ? are we okay to keep it like that ? (https://github.com/Mapotempo/optimizer-api/pull/192)

PR #182, opened 2021-04-12
Fix self selection when routes

fonsecadeline:


lib/interpreters/compute_several_solutions.rb
```diff
@@ -216,6 +216,8 @@ def self.find_best_heuristic(service_vrp)
 
             Models::Route.new(vehicle: vrp.vehicles.find{ |v| v[:id] == route[:vehicle_id] }, mission_ids: mission_ids)
           }.compact
+        else
+          best[:heuristic] = nil
```
senhalil, line 220: If we do it like this, the `synthesis` output will not show that the supplied routes are used as first solution. I suggest implementing the logic directly at the line below without going through `nil`.

lib/interpreters/compute_several_solutions.rb
```diff
@@ -226,7 +228,7 @@ def self.find_best_heuristic(service_vrp)
         }
         synthesis.each{ |synth| synth.delete(:solution) }
         vrp.resolution_batch_heuristic = nil
-        vrp.preprocessing_first_solution_strategy = [best[:heuristic]]
+        vrp.preprocessing_first_solution_strategy = best[:heuristic] ? [verified(best[:heuristic])] : []
```
senhalil, line 231:
```suggestion
        vrp.preprocessing_first_solution_strategy = best[:heuristic] != 'supplied_initial_routes' ? [verified(best[:heuristic])] : []
```

senhalil (changes requested):

senhalil (approved):

test/wrappers/ortools_test.rb
```diff
@@ -5041,20 +5000,19 @@ def test_self_selection_computes_saving_only_once_if_rest
       }],
       duration: 1
     }]
-    vrp.vehicles.first.rest_ids = ['rest_0']
+    vrp[:vehicles].first[:rest_ids] = ['rest_0']
```
fab-girard, line 5003: why don't you use accessor anymore? (best practice: use accessor as possible)

fonsecadeline: We've had a discussion about this with @senhalil : it is safer to call TestHelper.create on vrp we do want to use for the test. Like this we can check consistency of VRP generated for the test.

senhalil: @fonsecadeline I think you are answering a different question but that's true :) @fab-girard this test used to create the Vrp object from the hash first and then modify the Vrp object manually and call an internal function after. But this is risky (unless it is intentional) because we can create an invalid Vrp and call the internal function with an invalid Vrp (so test might not break when it should or vice-versa). So now we first modify the hash to have the problem we need and then create the Vrp object.

I believe Fabien asked this because it is not clear that `vrp` variable is a `Hash` and not a `Vrp` object because of its name. We should probably rename it, it is confusing like this.

In some tests, this is already the case and we do something as follows:
``` ruby
problem = VRP.lat_lon
problem[:vehicles] << problem[:vehicles].first.dup # first modify the problem into shape
problem[:vehicles].last[:id] += '_dup'

# then create and use the vrp
Interpreters::SplitClustering.generate_split_vrps(vrp: TestHelper.create(problem), service: :demo) 

# or if the vrp object is reused/needed, we assign it to a variable 
vrp = TestHelper.create(problem)
```

fab-girard (approved):

PR #183, opened 2021-04-13
Fix split_solve and tests

senhalil:
Split_solve should not ignore the existing matrix for transferred cars
Increase github actions time for split_tests

### DIFF
--- a/test/lib/interpreters/split_clustering_test.rb
+++ b/test/lib/interpreters/split_clustering_test.rb
@@ -536,7 +536,10 @@ class SplitClusteringTest < Minitest::Test
 
       problem = TestHelper.create(vrp)
       check_vrp_services_size = problem.services.size
-      result = OptimizerWrapper.wrapper_vrp('ortools', { services: { vrp: [:ortools] }}, problem, nil)
+      error = proc{ raise 'Split_solve should not demand matrix for a problem which has the complete matrix' }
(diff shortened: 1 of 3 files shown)


test/lib/interpreters/split_clustering_test.rb
```diff
@@ -536,7 +536,10 @@ def test_max_split_size_with_empty_fill
 
       problem = TestHelper.create(vrp)
       check_vrp_services_size = problem.services.size
-      result = OptimizerWrapper.wrapper_vrp('ortools', { services: { vrp: [:ortools] }}, problem, nil)
+      error = proc{ raise 'Split_solve should not demand matrix for a problem which has the complete matrix' }
```
fonsecadeline, line 539: Do you know if there is any way to do this for every test? Like having a global stub that raises any time we call _OptimizerWrapper.router.matrix_ function ?

fonsecadeline (approved):
Just one question but I may be too 'greedy' :)

test/lib/interpreters/split_clustering_test.rb
senhalil: we need to stub it at the setup phase and it would work for all tests (I guess) but then the tests which are already stubbing matrix to return a fixed matrix info wouldn't work :/ so we might eventually do that but it needs a redesign of sorts.

PR #185, opened 2021-04-13
Return nil in skills if none provided

fonsecadeline:


api/v01/entities/vrp_input.rb
```diff
@@ -498,10 +498,10 @@ module VrpVehicles
     optional(:maximum_ride_time, type: Integer, desc: 'Maximum ride duration between two route activities')
     optional(:maximum_ride_distance, type: Integer, desc: 'Maximum ride distance between two route activities')
     optional :skills, type: Array[Array[Symbol]],
-                      coerce_with: ->(val) {
+                      coerce_with: -> (val) {
```
senhalil, line 501: on my computer, rubocop suggest the contrary, is this a typo?

Do not use spaces between `->` and `(` in lambda literals. (convention:Layout/SpaceInLambdaLiteral)

senhalil (approved):
there is a space typo; otherwise :heavy_check_mark:

api/v01/entities/vrp_input.rb
fonsecadeline: Hum, my rubocoop put an alert here ... (oh, I just checked, alert does not ask for a space in fact :/)

senhalil (approved):

PR #187, opened 2021-04-19
Fixing randomly breaking tests

senhalil:


braktar (approved):

.github/actions/tests.sh
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env bash
 
 CONTAINER=${PROJECT}_api.1.$(docker service ps -f "name=${PROJECT}_api.1" ${PROJECT}_api -q --no-trunc | head -n1)
-docker exec -i ${CONTAINER} bundle exec rake test ${OPTIONS}
+docker exec -i ${CONTAINER} rake test TESTOPTS="${TESTOPTS}" ${OPTIONS}
```
fonsecadeline, line 4: Do we want to keep this ? I guess that allowed you to run only part of tests. 

If it is meant to stay, can you please tell us how to use this ?

senhalil: It is used to set test options -- for example, to set the seed to launch the test with the same random numbers, etc.

`TESTOPTS` can be defined just like `OPTIONS` variable which is defined in main.yml. 

First one applies to all "jobs" otherwise, it can be defined separately for each job.
Screenshot from 2021-04-20 11-41-59

fonsecadeline: Thank you :)

fonsecadeline (approved):

PR #190, opened 2021-04-21
Fix reading of JSON at csv format

braktar:


api/v01/entities/vrp_input.rb
```diff
@@ -544,7 +544,7 @@ module VrpVehicles
     mutually_exclusive :capacity_ids, :capacities
 
     optional(:sequence_timewindow_ids, type: Array[String], documentation: { hidden: true }, desc: 'Sequence timewindows to consider, CSV front only',
-                                       coerce_with: ->(val) { val.split(/,/).map(&:strip) })
+                                       coerce_with: ->(val) { val.is_a?(Array) ? val : val.split(/,/).map(&:strip) })
```
fonsecadeline, line 547: I would rather do it like that to ensure we know what we are doing : if it is a String then we know we want to split, otherwise we assume it is an Array. If it is not then that would raise.

```suggestion
                                       coerce_with: ->(val) { val.is_a?(String) ? val.split(/,/).map(&:strip) : val })
```

If it is an integer we do not want to try to split, we just want to reject this value.

fonsecadeline (changes requested):

fonsecadeline (approved):

senhalil (approved):

PR #191, opened 2021-04-21
Clean scheduling tests

fonsecadeline:
Clean tests core and removed or moved some tests when that was consistent to.


fonsecadeline:
| Time | Action
-- | -- | --
test_performance_13vl | 66.25s | kept
test_fill_days_and_post_processing | 342.15s | kept
test_instance_andalucia2 | 0.88s | kept
test_instance_baleares2 | 0.95s | kept
test_instance_clustered | 305.15s | Removed because clustering + sched already tested several times
test_treatment_site | 77.44s | kept
test_instance_andalucia1_two_vehicles | 1.78s | kept
test_instance_same_point_day | 74.27s | kept
test_route_initialisation | 1.86s | kept
test_performance_12vl | 164.61s | Removed because also in real_cases_solver_test
test_vrp_allow_partial_assigment_false | 64.38s | kept
test_without_same_point_day | 8.24s | Moved to real_cases_solver_test
test_instance_baleares2_with_priority | 0.96s | Merged with baleares2 test
test_quality_with_minimum_stops_in_route | 1.06s | kept
test_minimum_stop_in_route | 3.20s | kept
test_two_phases_clustering_sched_with_freq_and_same_point_day_5veh | 262.20s | Removed because also in real_cases_solver_test

braktar (approved):

senhalil (approved):
I don't see any reason for `test_instance_same_point_day` to break but it is broken, @fonsecadeline any idea?

fonsecadeline:
> I don't see any reason for `test_instance_same_point_day` to break but it is broken, @fonsecadeline any idea?

I edited the test to check performance and check same_point_day consistency. I guess it fails because post-processing can violate this _same_point_day_ option. I will edit the test accordingly.

senhalil:
> I edited the test to [...] check same_point_day consistency. 

Consistency of the result or the problem? I thought we already validated the same_point_day consistency of the result in this test.

fonsecadeline:
> > I edited the test to [...] check same_point_day consistency.
> 
> Consistency of the result or the problem? I thought we already validated the same_point_day consistency of the result in this test.

No we did not use to check this, we only checked quantities (which was not very consistent). In fact here we have a specific case : 1 service with 2 visits and lapse of 42, 1 service at same location with 3 visits and lapse of 28. So in fact those can not be grouped at same days. That is why the test fails, but in fact it is good that the algorithm assigns visits like this.

PR #192, opened 2021-04-21
Move consistency checkers to a class (2)

fonsecadeline:
This PR replaces https://github.com/Mapotempo/optimizer-api/pull/181

### DIFF
--- a/models/concerns/validate_data.rb
+++ b/models/concerns/validate_data.rb
@@ -1,0 +1,265 @@
+# Copyright © Mapotempo, 2021
+#
+# This file is part of Mapotempo.
+#
+# Mapotempo is free software. You can redistribute it and/or
+# modify since you respect the terms of the GNU Affero General
+# Public License as published by the Free Software Foundation,
+# either version 3 of the License, or (at your option) any later version.
+#
+# Mapotempo is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+# or FITNESS FOR A PARTICULAR PURPOSE.  See the Licenses for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with Mapotempo. If not, see:
+# <http://www.gnu.org/licenses/agpl.html>
+#
+require 'active_support/concern'
+
+# Expands provided data
+module ValidateData
+  extend ActiveSupport::Concern
+
+  def check_consistency(hash)
+    hash[:services] ||= []
+    hash[:shipments] ||= []
+    @hash = hash
+
+    ensure_uniq_ids
+    ensure_no_conflicting_skills
+
+    configuration = @hash[:configuration]
+    schedule = configuration[:schedule] if configuration
+    periodic_heuristic = configuration &&
+                         configuration[:preprocessing] &&
+                         configuration[:preprocessing][:first_solution_strategy].to_a.include?('periodic')
+    check_matrices
+    check_vehicles
+    check_relations(periodic_heuristic)
+    # TODO : this should be replaced by schedule when max_split does not use visits_number > 1 without schedule anymore
+    # indeed, no configuration implies no schedule and there should be no visits_number > 1 in this case
+    # check_services_and_shipments(schedule)
+    check_services_and_shipments(configuration, schedule)
+    check_shipments_specificities
+
+    check_routes(periodic_heuristic)
+    check_configuration(configuration, periodic_heuristic) if configuration
+  end
+
+  def ensure_uniq_ids
+    # TODO: Active Hash should be checking this
+    [:matrices, :units, :points, :rests, :zones, :timewindows,
+     :vehicles, :services, :shipments, :subtours].each{ |key|
+      next if @hash[key]&.collect{ |v| v[:id] }&.uniq!.nil?
+
+      raise OptimizerWrapper::DiscordantProblemError.new("#{key} IDs should be unique")
+    }
+  end
+
+  def ensure_no_conflicting_skills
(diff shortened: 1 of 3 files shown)


fonsecadeline:
Needs : 
- [ ] merge of https://gitlab.com/mapotempo/optimizer-api/-/issues/767
- [ ] Real scheduling cases should be okay after we merge #191

models/vrp.rb
```diff
@@ -124,7 +126,7 @@ def self.create(hash, delete = true)
       vrp = super({})
       self.filter(hash) # TODO : add filters.rb here
       # moved filter here to make sure we do have schedule_indices (not date) to do work_day check with lapses
-      self.check_consistency(hash)
+      vrp.check_consistency(hash)
```
fonsecadeline, line 129: See conversation https://github.com/Mapotempo/optimizer-api/pull/181#discussion_r617300541
(I am sorry, I force pushed the branch when previous PR was closed and I could not open PR181 again!)

senhalil:
@fonsecadeline is it me or a test that needs to be introduced in PR https://github.com/Mapotempo/optimizer-api/pull/193 stayed on the PR https://github.com/Mapotempo/optimizer-api/pull/192 side after the split?

fonsecadeline:
@senhalil I think I had added no test, but I will on https://github.com/Mapotempo/optimizer-api/pull/193

senhalil:
I don't remember but I think I saw modified tests and thought that this PR adds tests by mistake (instead of just refacto) but now I checked again, I don't see any modified tests. Maybe I check the PR during rebase or something.

fonsecadeline:
> I don't remember but I think I saw modified tests and thought that this PR adds tests by mistake (instead of just refacto) but now I checked again, I don't see any modified tests. Maybe I check the PR during rebase or something.

For my part I thought I had added some tests, just because _max_split_ tests where failing but in fact it is just that refactoring implies checks that we did not check before : https://gitlab.com/mapotempo/optimizer-api/-/issues/767

If I sum up, we did not use to check that services/shipments' visits_number was == 1 if no configuration was provided. Now, when no configuration is provided we assume no schedule is provided (which is correct) and reject if services.visits_number > 1.

senhalil:
Thanks for the sum up. I forgot that this was that PR ! To not to break mt/dev, we can skip the offending tests, we already have a ticket that mentions these tests, we want to treat this ticket as soon as possible so that way we wouldn't creak out CI for everyone.

Do we know why the "move" of checkers leads to broken tests in this PR? I am asking because we add the check `visits_number > 1` in 193 (so it shouldn't break here), right ? (maybe I am confused of the order of the PRs and which one is based on which :thinking: )

fonsecadeline:
> Thanks for the sum up. I forgot that this was that PR ! To not to break mt/dev, we can skip the offending tests, we already have a ticket that mentions these tests, we want to treat this ticket as soon as possible so that way we wouldn't creak out CI for everyone.
> 
> Do we know why the "move" of checkers leads to broken tests in this PR? I am asking because we add the check `visits_number > 1` in 193 (so it shouldn't break here), right ? (maybe I am confused of the order of the PRs and which one is based on which thinking )

You can see more details on the GitLab ticket but it is because before refacto we did not reach this check if we had no _configuration_ in VRP (when coming from _max_split_ there is apparently no configuration so it was okay). But now no configuration implies no schedule so the check fails).

senhalil:
> You can see more details on the GitLab ticket but it is because before refacto we did not reach this check if we had no _configuration_ in VRP (when coming from _max_split_ there is apparently no configuration so it was okay). But now no configuration implies no schedule so the check fails).

Thanks for the explanation @fonsecadeline  Sorry by the way, I didn't know that it was already explained in the ticket :(  Next time you can just copy/paste the ticket link saying that it is explained -> HERE :smile:  It is hard to know what's discussed in the PR, what's discussed in the ticket

fonsecadeline:
> > You can see more details on the GitLab ticket but it is because before refacto we did not reach this check if we had no _configuration_ in VRP (when coming from _max_split_ there is apparently no configuration so it was okay). But now no configuration implies no schedule so the check fails).
> 
> Thanks for the explanation @fonsecadeline Sorry by the way, I didn't know that it was already explained in the ticket :( Next time you can just copy/paste the ticket link saying that it is explained -> HERE smile It is hard to know what's discussed in the PR, what's discussed in the ticket

OK ! Yes, it got a little bit messy :)

models/vrp.rb
fonsecadeline: @senhalil @braktar did we agree on keeping this as is because it is meant to be changed anyway ?

models/concerns/validate_data.rb
```diff
@@ -0,0 +1,250 @@
...
+
+  def check_services_and_shipments(configuration, schedule)
+    (@hash[:services].to_a + @hash[:shipments].to_a).each{ |mission|
+      if (mission[:minimum_lapse] || 0) > (mission[:maximum_lapse] || 2**32)
+        raise OptimizerWrapper::DiscordantProblemError.new('Minimum lapse can not be bigger than maximum lapse')
+      end
+
+      # TODO : this should be replaced next line when max_split does not use visits_number > 1 without schedule anymore
+      # next if schedule && schedule[:range_indices] || mission[:visits_number].to_i <= 1
+      next if configuration.nil? || schedule && schedule[:range_indices] || mission[:visits_number].to_i <= 1
```
fonsecadeline, line 124: @braktar @senhalil I changed here : for now if we have no configuration (coming from max_split in general) then we do not do this check. 

PR is ready to make this check more general after we addit max split : https://github.com/Mapotempo/optimizer-api/pull/193

senhalil (approved):

braktar (approved):

PR #193, opened 2021-04-21
Add validation on visits number

fonsecadeline:
Needs edit in max_split : visits_number should not be used if no schedule is provided (or vrp.create should be done before).


fonsecadeline:
This is done in https://github.com/Mapotempo/optimizer-api/pull/243

PR #194, opened 2021-04-21
Skills in scheduling

fonsecadeline:
- [x] merge https://github.com/Mapotempo/optimizer-api/pull/186


CHANGELOG.md
```diff
@@ -25,6 +25,7 @@
 - Support relations (`order`, `same_route`, `sequence`, `shipment`) in split_solve algorithm (`max_split_size`) and partitions (`[:configuration][:preprocessing][:partitions]`) [#145](https://github.com/Mapotempo/optimizer-api/pull/145)
 - split_solve algorithm (`max_split_size`) respects relations (`vehicle_trips`, `meetup`, `minimum_duration_lapse`, `maximum_duration_lapse`, `minimum_day_lapse`, `maximum_day_lapse`) [#145](https://github.com/Mapotempo/optimizer-api/pull/145)
 - Return `geojsons` structure according to `geometry` parameter [#165](https://github.com/Mapotempo/optimizer-api/pull/165)
+- Support skills in periodic heuristic [#194](https://github.com/Mapotempo/optimizer-api/pull/194)
```
senhalil, line 28:
```suggestion
- Support skills in periodic heuristic (`first_solution_strategy='periodic'`) [#194](https://github.com/Mapotempo/optimizer-api/pull/194)
```

lib/heuristics/concerns/scheduling_data_initialisation.rb
```diff
@@ -21,26 +21,21 @@
...
       @candidate_routes[vehicle.original_id][vehicle.global_day_index] = {
-        vehicle_id: vehicle.id,
-        global_day_index: vehicle.global_day_index,
-        tw_start: (vehicle.timewindow.start < 84600) ? vehicle.timewindow.start : vehicle.timewindow.start - vehicle.global_day_index * 86400,
-        tw_end: (vehicle.timewindow.end < 84600) ? vehicle.timewindow.end : vehicle.timewindow.end - vehicle.global_day_index * 86400,
-        start_point_id: vehicle.start_point&.id,
-        end_point_id: vehicle.end_point&.id,
+        vehicle: vehicle,
+        tw_start: reduce_to_24h(vehicle.timewindow.start, vehicle.global_day_index),
+        tw_end: reduce_to_24h(vehicle.timewindow.end, vehicle.global_day_index),
```
senhalil, line 35: why not just do `vehicle.timewindow.start % 84600` we wouldn't even need a function that way.

lib/heuristics/concerns/scheduling_end_phase.rb
```diff
@@ -239,7 +239,8 @@ def empty_underfilled
 
           all_empty = false
 
-          next if route_data[:stops].sum{ |stop| @services_data[stop[:id]][:raw].exclusion_cost || 0 } >= route_data[:cost_fixed]
+          next if route_data[:stops].sum{ |stop| @services_data[stop[:id]][:raw].exclusion_cost || 0 } >=
+                  route_data[:vehicle].cost_fixed
```
senhalil, line 243: this looks weird, rubocop should have marked it.

in any case it is better to keep the conditional operator (==, >=, etc) at the same line with part of the left hand side and right hand side at the same time (to clearly see what gets compared with what) like
```suggestion
          next if route_data[:stops].sum{ |stop| 
                    @services_data[stop[:id]][:raw].exclusion_cost || 0
                  } >= route_data[:vehicle].cost_fixed
```
otherwise, t looks like a separate condition on its own.

(`|| 0` can be replaced with `.to_i` or `.to_f` depending on if exclusion_cost is a float or int)

test/lib/heuristics/scheduling_test.rb
```diff
@@ -553,6 +553,20 @@ def test_sticky_in_scheduling
       refute_includes result[:routes].find{ |r| r[:activities].any?{ |stop| stop[:service_id] == 'service_6_1_1' } }[:vehicle_id], 'vehicle_0_'
     end
 
+    def test_skills_in_scheduling
+      vrp = VRP.lat_lon_scheduling_two_vehicles
+      result = OptimizerWrapper.wrapper_vrp('ortools', { services: { vrp: [:ortools] }}, TestHelper.create(vrp), nil)
+      assigned_route = result[:routes].find{ |r| r[:activities].any?{ |stop| stop[:service_id] == 'service_6_1_1' } }
+      assert_includes assigned_route[:vehicle_id], 'vehicle_0_' # default result
```
senhalil, line 560: do we know why vehicle_0_ is "default" when there is no skills? is there a chance that changing the heuristic would assign this service to another vehicle? then this test would give a false negative, right?

test/lib/heuristics/scheduling_test.rb
```diff
@@ -553,6 +553,20 @@ def test_sticky_in_scheduling
...
+
+      vrp = VRP.lat_lon_scheduling_two_vehicles
+      vrp[:vehicles][1][:skills] = [[:compatible]]
+      vrp[:services].find{ |s| s[:id] == 'service_6' }[:skills] = [:compatible]
+      result = OptimizerWrapper.wrapper_vrp('ortools', { services: { vrp: [:ortools] }}, TestHelper.create(vrp), nil)
+      assigned_route = result[:routes].find{ |r| r[:activities].any?{ |stop| stop[:service_id] == 'service_6_1_1' } }
+      refute_includes assigned_route[:vehicle_id], 'vehicle_0_'
```
senhalil, line 567: instead of refuting vehicle_0 (or at least additional to that), it would make more sense to check if the assigned route is the vehicle with the correct skill.

senhalil: if there is a chance that this service would get assigned to vehicle_1 even without the skills then this test would give false positives, right? 

(maybe this is already the case but I don't see it clearly, if that is the case a small note at the beginning of the test would be very useful) it might make sense to force the service to on a vehicle by "convenience" but then force it to be on the other "inconvenient" vehicle (with more cost) due to skills.

lib/heuristics/scheduling_heuristic.rb
```diff
@@ -400,18 +413,18 @@ def reorder_stops(vrp)
...
+          scheduling_route_time = back_to_depot - route_data[:stops].first[:start]
+          solver_route_time = (result[:routes].first[:activities].last[:begin_time] -
+                              result[:routes].first[:activities].first[:begin_time]) # last activity is vehicle depot
 
-          next if scheduling_route_time - solver_route_time < @candidate_services_ids.flat_map{ |s| @services_data[s][:durations] }.min ||
-                  result[:routes].first[:activities].collect{ |stop| @indices[stop[:service_id]] }.compact == route_data[:stops].collect{ |s| @indices[s[:id]] } # we did not change our points order
+          next if scheduling_route_time - solver_route_time <
+                  @candidate_services_ids.flat_map{ |s| @services_data[s][:durations] }.min ||
+                  result[:routes].first[:activities].collect{ |stop| @indices[stop[:service_id]] }.compact ==
+                  route_data[:stops].collect{ |s| @indices[s[:id]] } # we did not change our points order
```
senhalil, line 425: Can you check other conditions so that they respect the rubocop and they are not splitted in half by the condition?

senhalil (changes requested):

lib/heuristics/scheduling_heuristic.rb
```diff
@@ -842,8 +842,13 @@ def compatible_days(service_id, day)
       !@services_data[service_id][:raw].unavailable_days.include?(day)
     end
 
+    def compatible_vehicle(service_id, route_data)
+      route_data[:vehicle].skills.any?{ |skill_set| (@services_data[service_id][:raw].skills - skill_set).empty? }
+    end
```
senhalil, line 870: Like this we risk two incompatible services get assigned to the same vehicle because one needs first skill_set, the other needs the second skill_set but they return true. However, skills_sets are not cumulatif (if I am not mistaken). 

Either we need a way to distinguish which skills left "eliminated" or we need to add an assert so that periodic heuristic can handle at most one skill_set.

It is not hard to achieve the first but it is a bit delicate (if the user doesn't expect to have multiple skill_sets, I suggest we add a note here and an assert to the wrapper.rb)

senhalil (changes requested):
current implementation cannot handle multiple skill_sets

lib/heuristics/scheduling_heuristic.rb
fonsecadeline: I agree with this suggestion :) especially because alternative skills are not needed in periodic_heuristic yet. But thank you for noticing, I do not think you are mistaken.

test/lib/heuristics/scheduling_test.rb
fonsecadeline: Maybe we need to add a comment. The aim is exactly that, if we change heuristic such that we do not assign 'vehicle_0_' then we need to change the test. Because skills will not be the reason why we affected this vehicle to this service anymore.

test/lib/heuristics/scheduling_test.rb
fonsecadeline: That is exactly why we have "default result" (see your previous comment) : to ensure we are using this vehicle because of skills, and no other reason.

lib/heuristics/scheduling_heuristic.rb
fonsecadeline: I do not think this is a rule we have defined in rubocoop, I do not have any alert. Maybe we should add it in common rules ?

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## [v1.8.0-dev] - Unreleased
 
 ### Added
+- Support skills in periodic heuristic (`first_solution_strategy='periodic'`) [#194](https://github.com/Mapotempo/optimizer-api/pull/194)
```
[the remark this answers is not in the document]
senhalil: I believe it was me who suggested that part but I must have copy/pasted the `first_solution_strategy='periodic'` suggestion from the Changelog. If we prefer `:`, should fix both (and others if exists) so that we can prevent this back and forth.

braktar (approved):

wrappers/wrapper.rb
```diff
@@ -61,9 +61,12 @@ def assert_vehicles_no_capacity_initial(vrp)
...
-      vrp.vehicles.empty? || vrp.vehicles.none?{ |vehicle|
-        !vehicle.skills || vehicle.skills.size > 1
-      }
+      vrp.vehicles.none?{ |vehicle| vehicle.skills.size > 1 }
+    end
+
+    def assert_vehicles_no_alternative_skills_with_periodic_heuristic(vrp)
+      !vrp.preprocessing_first_solution_strategy.include?('periodic') ||
+        vrp.vehicles.none?{ |vehicle| vehicle.skills.size > 1 }
     end
```
senhalil, line 70: Can't we move this assert to vrp.create? I feel like in this case we refuse to solve the VRP, right? So there is no need to create the VRP.

senhalil (changes requested):

braktar (approved):

senhalil (approved):

braktar (changes requested):
Both : 
`test/lib/interpreters/split_clustering_test.rb:501:in 'test_fail_when_alternative_skills' `
and
`test/lib/interpreters/split_clustering_test.rb:962:in 'test_split_with_vehicle_alternative_skills'`
fail with the error : `Minitest::UnexpectedError:         OptimizerWrapper::DiscordantProblemError: Periodic heuristic does not support vehicle alternative skills`

braktar (approved):

PR #198, opened 2021-04-27
Fix vroom solution reading

braktar:


test/wrappers/vroom_test.rb
```diff
@@ -756,4 +756,17 @@ def test_mixed_shipments_and_services
...
+
+  def test_correct_route_collection
+    problem = VRP.lat_lon_two_vehicles
+    problem[:services].each{ |service|
+      service[:skills] = ['s1']
+    }
+    problem[:vehicles].last[:skills] = [['s1']]
+
+    result = OptimizerWrapper.wrapper_vrp('demo', { services: { vrp: [:vroom] }}, TestHelper.create(problem), nil)
+    assert result
```
fonsecadeline, line 768: I think we can stop checking this. If we have no result then next line will fail and while debugging it will be easy to see result is missing.

test/wrappers/vroom_test.rb
```diff
@@ -756,4 +756,17 @@ def test_mixed_shipments_and_services
...
+    assert_equal 2, result[:routes].size
+    assert_equal problem[:services].size + 2, result[:routes].find{ |route| route[:vehicle_id] == problem[:vehicles].last[:id] }[:activities].size
```
fonsecadeline, line 770:
```suggestion
    assert_equal problem[:services].size, result[:routes].find{ |route| route[:vehicle_id] == problem[:vehicles].last[:id] }[:activities].count{ |activity| activity[:service_id }
```

fonsecadeline (changes requested):

test/wrappers/vroom_test.rb
senhalil: Unless we want to make sure that the depots are there or not?

senhalil (approved):

test/wrappers/vroom_test.rb
fonsecadeline: If so, I suggest we make it more explicit.

braktar: Figuring out if depots are there or not is not the purpose of this test, so I will follow the recommendations of @fonsecadeline

fonsecadeline (approved):

senhalil (approved):

PR #199, opened 2021-04-28
Improve performance

fonsecadeline:
Result can be either a hash or a table of hash.

It will return a hash in most of the cases, and a table of hash if we use _batch_heuristic_.

Two options : 
- do this check in vrp.rb
- do this check when writing result on redis

### DIFF
--- a/api/v01/vrp.rb
+++ b/api/v01/vrp.rb
@@ -117,7 +117,7 @@ module Api
               elsif ret.is_a?(Hash)
                 status 200
                 if vrp.restitution_csv
-                  present(OptimizerWrapper.build_csv(ret), type: CSV)
+                  present(OptimizerWrapper.build_csv([ret]), type: CSV)
                 else
                   present({ solutions: [ret], job: { status: :completed }}, with: VrpResult)
                 end
@@ -165,11 +165,17 @@ module Api
 
             status 200
 
+            unless solution.nil? || solution[:result].nil? || solution[:result].is_a?(Array)
+            # TODO: solution[:result] should always be an array, find out why it is not and
+            # remove this check and other similar ones -- i.e.,  [solution[:result]].flatten(1).
+              solution[:result] = [solution[:result]]
+            end
+
             if output_format == :csv && (job.nil? || job.completed?) # At this step, if the job is nil then it has already been retrieved into the result store
               present(OptimizerWrapper.build_csv(solution[:result]), type: CSV)
             else
               present({
-                solutions: [solution[:result]].flatten(1),
+                solutions: solution[:result],
                 job: {
                   id: id,
                   status: job&.status&.to_sym || :completed, # :queued, :working, :completed, :failed
--- a/lib/output_helper.rb
+++ b/lib/output_helper.rb
@@ -92,7 +92,7 @@ module OutputHelper
 
   class Result
     def self.generate_geometry(solution)
-      return nil unless solution.to_h[:result]
+      return nil if solution.nil? || solution[:result].nil? || solution[:result].empty?
 
       @colors = ['#DD0000', '#FFBB00', '#CC1882', '#00CC00', '#558800', '#009EFF', '#9000EE',
                  '#0077A3', '#000000', '#003880', '#BEE562']
@@ -101,7 +101,7 @@ module OutputHelper
       return nil unless expected_geometry.any?
 
       expected_geometry.map!(&:to_sym)
-      [solution[:result]].flatten(1).collect{ |result|
+      solution[:result].collect{ |result|
         geojson = {}
 
         geojson[:partitions] = generate_partitions_geometry(result) if expected_geometry.include?(:partitions)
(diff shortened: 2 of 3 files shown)


braktar:
Does it really have an impact on performances ?

fonsecadeline:
> Does it really have an impact on performances ?

Yes : 
`solution&.key?(:result)` is instantaneous whereas `solution.to_h[:result]` is not.

+ I would say that avoiding _flatten_ operation on a _solution_ (which can be quite big) is more performant. I was debugging OOM issue, I do not have the numbers.

api/v01/vrp.rb
```diff
@@ -169,7 +169,7 @@ class Vrp < APIBase
               present(OptimizerWrapper.build_csv(solution[:result]), type: CSV)
             else
               present({
-                solutions: [solution[:result]].flatten(1),
+                solutions: solution[:result].is_a?(Array) ? solution[:result] : [solution[:result]],
```
senhalil, line 172: I am glad this improved the performance as expected. @fonsecadeline if we don't have a ticket can you create one about this issue please, solution[:result] should be always an array whether coming from Redis.get, wrapper_vrp or else.

lib/output_helper.rb
```diff
@@ -92,15 +92,16 @@ def self.csv_line(vrp, service, cluster_index, two_stages = false)
 
   class Result
     def self.generate_geometry(solution)
-      return nil unless solution.to_h[:result]
+      return nil unless solution&.key?(:result)
```
senhalil, line 95: What if the hash has the key `:result` but it is `nil`? 

I think your first implementation `solution && solution[:result]` was better.

or you can be more expressive and do `return nil if solution.nil? || solution[:result].nil? || solution[:result].empty?` because this reads literally the conditions we care.

lib/output_helper.rb
```diff
@@ -92,15 +92,16 @@ def self.csv_line(vrp, service, cluster_index, two_stages = false)
...
 
       @colors = ['#DD0000', '#FFBB00', '#CC1882', '#00CC00', '#558800', '#009EFF', '#9000EE',
                  '#0077A3', '#000000', '#003880', '#BEE562']
 
       expected_geometry = solution[:configuration].to_h[:geometry].to_a.map!(&:to_sym) # TODO : investigate how it is possible that no configuration is returned
+      return nil unless expected_geometry.any?
 
       expected_geometry.map!(&:to_sym)
-      [solution[:result]].flatten(1).collect{ |result|
+      (solution[:result].is_a?(Array) ? solution[:result] : [solution[:result]]).collect{ |result|
```
senhalil, line 104: `solution[:result] = [solution[:result]] unless solution.nil? || solution[:result].nil? || solution[:result].is_a?(Array)`* before calling `present` inside the else condition so that this function can work without any problems and we don't need to repeat this operation in two places.

*: or something like this.

senhalil:
At the end, we always return a table of hash to the user (even without batch_heuristic option), we should not have to do any checks but the result should always be a table of hash even if we don't call `batch_heuristic`. 

So this means, even if there is no `batch_heuristic`, we still should have a `collect` or an array conversion at the correct location

senhalil:
I just digged a bit and it looks like we do something like that  -- i.e., even if there bath_heuristic is false, the service_vrp enters into an array 
```ruby
    def self.expand_several_solutions(service_vrps)
      several_solutions(service_vrps).flat_map{ |each_service_vrps|
        if service_vrps[0][:vrp].resolution_batch_heuristic
          batch_heuristic(each_service_vrps)
        else
          [each_service_vrps]
        end
      }
    end
 ```
 and these service_vrps (or a single service_vrp) is solved and its results are collected in `define_main_process`
 ```ruby
  def self.define_main_process(services_vrps, job = nil, &block)
    several_service_vrps = Interpreters::SeveralSolutions.expand_several_solutions(services_vrps)
    several_results = several_service_vrps.collect.with_index{ |current_service_vrps, solution_index|
    }
    several_results
  ensure
    log "<-- define_main_process elapsed: #{(Time.now - tic).round(2)} sec", level: :info
  end
```
which means we should be always returning an array.

This suggests me that there might be something weird going on during the `Result.set`. (we should give the link of this PR in the ticket)

api/v01/vrp.rb
fonsecadeline: https://gitlab.com/mapotempo/optimizer-api/-/issues/782

lib/output_helper.rb
fonsecadeline: I will go with your last suggestion.

api/v01/vrp.rb
```diff
@@ -165,11 +165,15 @@ class Vrp < APIBase
 
             status 200
 
+            unless solution.nil? || solution[:result].nil? || solution[:result].is_a?(Array)
```
senhalil, line 168:
```suggestion
            unless solution.nil? || solution[:result].nil? || solution[:result].is_a?(Array)
            # TODO: solution[:result] should always be an array, find out why it is not and 
            # remove this check and other similar ones -- i.e.,  [solution[:result]].flatten(1).
```

senhalil (approved):
It might be better to add a TODO to show that this a temporary band-aid

braktar (approved):

PR #201, opened 2021-04-30
Do not output polylines through geometry parameter

fonsecadeline:


models/types/geometry_type.rb
```diff
@@ -25,15 +25,17 @@ def self.type_cast(value)
...
     elsif value.is_a?(Array)
       to_return = []
       value.each{ |geometry_type|
         unless ALL_TYPES.include?(geometry_type.to_sym)
           raise ArgumentError.new("Invalid geometry value: #{geometry_type}")
         end
 
-        to_return << geometry_type.to_sym
+        # to_return << geometry_type.to_sym
+        to_return << geometry_type.to_sym unless %i[polylines encoded_polylines].include?(geometry_type.to_sym)
```
senhalil, line 38: We might want to put a limit on the size of the problem (vehicles/services/number of days) instead of completely preventing it. 

But if we want to completely prevent polylines, I suggest we modify `ALL_TYPES` so that if someone demands polylines it is not silently ignored.

senhalil (approved):

senhalil (commented):
I put aproved by mistake instead of comment

models/types/geometry_type.rb
senhalil: @fonsecadeline I meant (I put aproved by mistake instead of comment) but I have this :arrow_up:  comment.

fonsecadeline: We are in a tricky situation now : 
* if we change ALL_TYPES then we might return an error in cases that we told customers that it should work. That is, we will not ensure retrocompatibility. I thought we had discussed something different : just ignore this for now because we know that, even for small data, we are duplicating data in result. For me, since this value is only used by a set of users that do not really need it we can just filter this _polylines_. 

- [ ] edit data so we do not expose _polylines_

* we can filter only if services / routes number exceed a given number. In this case it will still be silently ignored in some cases (but for now none of our users that do need geometry will satisfy this limit I think).

fonsecadeline: > @fonsecadeline I meant (I put aproved by mistake instead of comment) but I have this arrow_up comment.

I saw, I was answering :)

senhalil: hmm you are right, their POSTs will be refused, but lets not forget the `geojsons` code is still not "released" it is not public. There is no "retro-compatibility", no? 

It feels silly to expose something that will be ignored (since this param is not exposed yet, exposing only the working parts sounds more appropriate). 

Still, I understand your concern that we don't want to bother the user who are testing the  unexposed parameter already. 

I guess this leaves us only "ignoring" option... If you think it is not necessary to have a size limit (to not allow under certain size) then you code is good. Lets make sure that there is no one else, that needs polylines though ! If there is someone we can find a limit between their size and 110K visits :D

senhalil: In fact except the clients that you know, no one could use `geosjson = [:polyline]`, because no one knows the eoption yet, right? If that is the case, I have one last question and we can resolve this thread, this modif doesn't affect the users of the old parameter, right? If there is a client getting polylines with the old parameter, they should continue to get their polylines.

fonsecadeline: You are right, we need to make a different modification if we want to make sure users that use old version of _geometry_ have expect result. I am fixing this right now.

senhalil (approved):
except the typos it is good to go imo :ok_hand:

test/api/v01/vrp_test.rb
```diff
@@ -368,27 +368,32 @@ def test_submit_without_schedule_start_should_break
   end
 
   def test_ask_for_geometry
-    [false, true, # boolean
+    [false, true, true, # boolean
      ['polylines'], [:polylines], [:polylines, 'partitions'], ['unexistant'], # array of string or symbol
-     'partitions', 'polylines,partitions', 'polylines, encoded_polylines' # string to be converted into an array
+     'partitions', 'polylines,partitions', 'polylines, encoded_polylines'  # string to be converted into an array
```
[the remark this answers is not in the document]
fonsecadeline: Space is a type but "true" is not, I added case with encoded_polylines to check previous behaviour

PR #203, opened 2021-04-30
Reduce vroom time thread consumption

fonsecadeline:
Some tests are still running to chose right value for X : 

- x 0 - 5, 6 minutes (Solution cost: 515723 & unassigned: 0)
- x 1 - 9 minutes (Solution cost: 515723 & unassigned: 0)
- x 2 - 14 minutes (Solution cost: 515723 & unassigned: 0)

I suggest using 0 not to take any risks without testing (we only tested on one instance)

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -36,7 +36,7 @@ module OptimizerWrapper
   HEURISTICS = %w[path_cheapest_arc global_cheapest_arc local_cheapest_insertion savings parallel_cheapest_insertion first_unbound christofides].freeze
   WEEKDAYS = %i[mon tue wed thu fri sat sun].freeze
   DEMO = Wrappers::Demo.new(tmp_dir: TMP_DIR, threads: 4)
-  VROOM = Wrappers::Vroom.new(tmp_dir: TMP_DIR, threads: 4)
+  VROOM = Wrappers::Vroom.new(tmp_dir: TMP_DIR, threads: 1)
   # if dependencies don't exist (libprotobuf10 on debian) provide or-tools dependencies location
   ORTOOLS_EXEC = 'LD_LIBRARY_PATH=../or-tools/dependencies/install/lib/:../or-tools/lib/ ../optimizer-ortools/tsp_simple'.freeze
   ORTOOLS = Wrappers::Ortools.new(tmp_dir: TMP_DIR, exec_ortools: ORTOOLS_EXEC, threads: 4)
--- a/wrappers/vroom.rb
+++ b/wrappers/vroom.rb
@@ -25,6 +25,7 @@ module Wrappers
     def initialize(hash = {})
       super(hash)
       @exec_vroom = hash[:exec_vroom] || '../vroom/bin/vroom'
+      @exec_vroom += " -t #{hash[:threads]}" if hash[:threads]
     end
 
     def solver_constraints
--- a/wrappers/vroom.rb
+++ b/wrappers/vroom.rb
@@ -386,7 +386,8 @@ module Wrappers
       output = Tempfile.new('optimize-vroom-output', @tmp_dir)
       output.close
 
-      cmd = "#{@exec_vroom} -i '#{input.path}' -o '#{output.path}'"
+      # TODO : find best value for x https://github.com/Mapotempo/optimizer-api/pull/203
+      cmd = "#{@exec_vroom} -i '#{input.path}' -o '#{output.path}' -x 0"
       log cmd
       system(cmd)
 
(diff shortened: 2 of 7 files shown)


wrappers/vroom.rb
```diff
@@ -25,6 +25,9 @@ class Vroom < Wrapper
     def initialize(hash = {})
       super(hash)
       @exec_vroom = hash[:exec_vroom] || '../vroom/bin/vroom'
+      if hash[:threads]
+        @exec_vroom += " -t #{hash[:threads]}"
+      end
```
senhalil, line 30:
```suggestion
      @exec_vroom += " -t #{hash[:threads]}" if hash[:threads]
```

wrappers/wrapper.rb
```diff
@@ -449,6 +449,13 @@ def assert_no_setup_duration(vrp)
         }
     end
 
+    def assert_max_split_not_exceeded(vrp)
+      return true unless vrp.preprocessing_max_split_size
+
+      empties_or_fills = vrp.services.select{ |s| s.quantities.any?(&:fill) || s.quantities.any?(&:empty) }
+      (vrp.services.size - empties_or_fills.size) <= vrp.preprocessing_max_split_size
+    end
```
senhalil, line 457: Can you call `Interpreters::SplitClustering.split_solve_candidate?({ vrp: vrp })` instead of manual check (because the max_split conditions are more complicated then this (for exampl there is only one vehicle, even if vrp.size > max_split_size, we should be able to use Vroom (if other conditions are satisfed) because we can't do split_solve.

I didn't tested it but something like this should work:
```suggestion
    def assert_max_split_is_not_active(vrp) # in this case we can rename the assert as follows
    def assert_not_a_split_solve_candidate(vrp) # or like this..
      !Interpreters::SplitClustering.split_solve_candidate?({ vrp: vrp })
    end
```

wrappers/vroom.rb
```diff
@@ -384,7 +388,7 @@ def run_vroom(vrp, dimensions, _job)
       output = Tempfile.new('optimize-vroom-output', @tmp_dir)
       output.close
 
-      cmd = "#{@exec_vroom} -i '#{input.path}' -o '#{output.path}'"
+      cmd = "#{@exec_vroom} -i '#{input.path}' -o '#{output.path}' -x 0"
```
senhalil, line 391: Can you add a `TODO:` and share the link of this PR in the ticket? It looks like we will move the ticket to the next milestone and make a full performance test..

senhalil: I didn't check but I suspect Flore's instance has no TWs which renders "everything possible" which degrades the performance of Vroom , because it looses time by trying useless combinations :( but these "exploration" operators might be really useful in more realistic instances with timewindows etc.

config/environments/production.rb
```diff
@@ -36,7 +36,7 @@ module OptimizerWrapper
   HEURISTICS = %w[path_cheapest_arc global_cheapest_arc local_cheapest_insertion savings parallel_cheapest_insertion first_unbound christofides].freeze
   WEEKDAYS = %i[mon tue wed thu fri sat sun].freeze
   DEMO = Wrappers::Demo.new(tmp_dir: TMP_DIR, threads: 4)
-  VROOM = Wrappers::Vroom.new(tmp_dir: TMP_DIR, threads: 4)
+  VROOM = Wrappers::Vroom.new(tmp_dir: TMP_DIR, threads: 1)
```
senhalil, line 39: Do you want to leave it 4 for tests and dev (or is it already 0)?

I would make the same thing for other environments so that we don't have a different (even better) performance when it is not production.

senhalil (changes requested):

wrappers/wrapper.rb
fonsecadeline: I thought we only want to limit on services. Are you saying VROOM is more efficient than ORtools for big instances that do not call split_solve ? The aim was initially to add a limit on number of services and we then decided to use max_split_size as limit value.

If we use `Interpreters::SplitClustering.split_solve_candidate?({ vrp: vrp })` instead of comparing with _max_split_size_ then we might still try to solve big problems with VROOM, I am not sure that is what we discussed.

config/environments/production.rb
fonsecadeline: We did not specify number of threads in test and development but it would make sense to have same limit everywhere, yes.

wrappers/vroom.rb
senhalil: both or-tools and vroom (and MIP as well for that matter) has troubles when there is no TWs (or when the TWs are "permissively" large).

If 87% of the services have TWs are longer than 4 hours (almost half of which even longer than 6.5 hours...) yeah, that is almost like no TW. it might even be harder since solver needs to check the TW constraints for violation but they don't really decrease the size of the search space so it might be even worse.

senhalil (approved):

PR #204, opened 2021-04-30
Generic handling of geojson for visualization

AntoineCAMILLE:
Before :  optimizer_old  
After :  optimizer


fonsecadeline (approved):
I do not know about JS, but front behaviour is as expected :)

public/js/result.js
```diff
@@ -241,11 +201,8 @@ function initializeMap(id) {
 }
 
 function initializeHandlers() {
-  $('#get-job').on('click', function() { getJob($('#job-id').val(), queryParams.api_key); });
+  $('#get-job').on('click', function() { console.log('CLIKED BATARD'); getJob($('#job-id').val(), queryParams.api_key); });
```
fab-girard, line 204: ?

PR #207, opened 2021-05-03
Fix front scheduling to display errors

braktar:


fonsecadeline (approved):

senhalil (approved):
@braktar I assume you tested the front with these modifications and it works right. I approve but I am not fluent in js.

PR #209, opened 2021-05-04
Prepare v1.7.1

braktar:


api/api_v01.rb
```diff
@@ -58,7 +58,7 @@ class ApiV01 < Grape::API
         'application/json; charset=UTF-8',
         'application/xml',
       ],
-      doc_version: '1.7.0',
+      doc_version: '1.7.1-rc',
```
senhalil, line 61: I think that `rc` is not the correct terminology here. 

We can use `unreleased` or `dev`. The reason is that we are still in the `dev` cycle actually. In fact, we are literally starting this dev cycle just now, there is no code to be a "candidate" to be "released". 

`rc` marks the end of an active dev cycle and the fact that the code is stable enough to be a candidate to be released and that it is actively being tested to be released. Most of the time, the code that will be actually published is very similar to the code marked as `rc` (even the same code if the testing phase doesn't show any problems).

If anything, we should mark the code `rc` when we start the massive_test procedure, up-to that point it is `dev` or `unreleased` imo. That way we would have a commit trace of when we started the massive_test procedure. What do you guys think @fonsecadeline @fab-girard @braktar ?

fonsecadeline (approved):

senhalil (approved):

api/api_v01.rb
fonsecadeline: I like the idea of marking what was runed in massive tests (I hope we remember to do that, I will add it in next massive test ticket for v1.8.0).

I have to say I was just following the original naming, I like with your suggestion :)

PR #212, opened 2021-05-05
Clean travis related

fonsecadeline:
I am not sure what we should do in _docker-compose.yml_ file.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -2,7 +2,6 @@ version: '3.3'
 services:
   redis:
     image: redis:${REDIS_VERSION:-3.2-alpine}
-    hostname: travis-redis
     volumes:
       - ../redis:/data
     command: redis-server --appendonly yes
@@ -11,7 +10,6 @@ services:
 
   redis-cache:
     image: redis:${REDIS_VERSION:-3.2-alpine}
-    hostname: travis-redis-cache
     volumes:
       - ../redis:/data
     command: redis-server --save ""
@@ -32,7 +30,6 @@ services:
 
   api:
     image: registry.test.com/mapotempo/optimizer-api:latest
-    hostname: travis-api
     environment:
       PASSENGER_APP_ENV: test
       REDIS_COUNT_HOST: redis-count
@@ -47,7 +44,6 @@ services:
   resque-default:
     image: registry.test.com/mapotempo/optimizer-api:latest
     working_dir: /srv/app
-    hostname: travis-resque-default
     environment:
       APP_ENV: test
       COUNT: 5
--- a/test/lib/heuristics/dichotomious_test.rb
+++ b/test/lib/heuristics/dichotomious_test.rb
@@ -23,11 +23,6 @@ class DichotomiousTest < Minitest::Test
       vrp = TestHelper.load_vrp(self)
       vrp.resolution_dicho_algorithm_service_limit = 457 # There are 458 services in the instance. TODO: Remove it once the dicho contions are stabilized
 
-      if ENV['TRAVIS'] # Compansate for travis performance
-        vrp.resolution_duration *= 1.60
-        vrp.resolution_minimum_duration *= 1.60
-      end
-
       t1 = Time.now
       result = OptimizerWrapper.wrapper_vrp('ortools', { services: { vrp: [:ortools] }}, vrp, nil)
       t2 = Time.now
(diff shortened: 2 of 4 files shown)


docker/docker-compose.yml
```diff
@@ -2,7 +2,7 @@ version: '3.3'
 services:
   redis:
     image: redis:${REDIS_VERSION:-3.2-alpine}
-    hostname: travis-redis
+    hostname: travis-redis # should we rename ? remove ?
```
braktar, line 5: Did you try to remove the hostname ? If so, does it have an impact on the CI ?

alain-andre: There is no impact. This was only to specify this compose file was for CI tests only.

test/lib/heuristics/dichotomious_test.rb
```diff
@@ -23,11 +23,6 @@ def test_dichotomious_approach
       vrp = TestHelper.load_vrp(self)
       vrp.resolution_dicho_algorithm_service_limit = 457 # There are 458 services in the instance. TODO: Remove it once the dicho contions are stabilized
 
-      if ENV['TRAVIS'] # Compansate for travis performance
```
alain-andre, line 26: Do we need specifying durations on CI Github Action ?  We can use `ENV['CI']` in place of `ENV['TRAVIS']`

test/rubocop_test.rb
```diff
@@ -19,7 +19,7 @@
 
 class RubocopTest < Minitest::Test
   def test_no_error_in_files
-    parallel = ENV['RUBOCOP_PARALLEL'] || ENV['TRAVIS'] ? '--parallel' : nil
```
alain-andre, line 22: I think running rubocop in parallel on `ENV['CI']` should be a good idea.

test/rubocop_test.rb
```diff
@@ -28,7 +28,7 @@ def test_no_error_in_files
   end
 
   def test_no_warning_in_tests
-    parallel = ENV['RUBOCOP_PARALLEL'] || ENV['TRAVIS'] ? '--parallel' : nil
```
alain-andre, line 31: I think running rubocop in parallel on `ENV['CI']` should be a good idea.

test/real_cases_dichotomious_test.rb
```diff
@@ -18,7 +18,7 @@
 require './test/test_helper'
 
 class RealCasesTest < Minitest::Test
-  if !ENV['SKIP_REAL_DICHO'] && !ENV['SKIP_DICHO'] && !ENV['TRAVIS']
```
alain-andre, line 21: If we need to skip this test on CI we can replace with `ENV['CI']`

test/lib/heuristics/dichotomious_test.rb
fonsecadeline: Tests work as is right now so I don't think we need anything like that for now.

docker/docker-compose.yml
alain-andre: Yes we can remove hostnames on each service. I would better change `travis` by `ci`

.github/workflows/main.yml
```diff
@@ -16,7 +16,7 @@ jobs:
   test_basis:
     runs-on: ubuntu-latest
     env:
-      OPTIONS: "COV=false LOG_LEVEL=info SKIP_DICHO=true SKIP_REAL_CASES=true SKIP_PERIODIC=true SKIP_SPLIT_CLUSTERING=true"
+      OPTIONS: "CI=true COV=false LOG_LEVEL=info SKIP_DICHO=true SKIP_REAL_CASES=true SKIP_PERIODIC=true SKIP_SPLIT_CLUSTERING=true"
```
alain-andre, line 19: This is not necessary, the environment variable *CI* is always true in Github Action : https://docs.github.com/en/actions/reference/environment-variables#default-environment-variables Sorry if I wasn't clear. You can remove for all jobs.

test/real_cases_dichotomious_test.rb
alain-andre: *RealCasesTest* must be launched in pipeline ? If not we must add `&& !ENV['CI']`

alain-andre (changes requested):

braktar (approved):

senhalil (approved):
Intensive option increases the test duration, timeout limits need to be corrected accordingly

alain-andre (approved):

senhalil:
@fonsecadeline it looks like this needs manual rebase since workflow/main.yml is modified. Can you do the rebase and force push? We can merge this if the time-out limits are alright :crossed_fingers:

PR #213, opened 2021-05-05
Edit quantity precision for vroom

braktar:


test/wrappers/ortools_test.rb
```diff
@@ -5770,4 +5770,27 @@ def test_self_selection_should_not_change_vehicle_ids
...
+    problem = VRP.lat_lon_two_vehicles
+    problem[:services].each{ |service|
+      service[:quantities] = [{ unit_id: 'kg', value: 1.001 }]
+    }
+    problem[:vehicles].each{ |vehicle|
+      vehicle[:capacities] = [{ unit_id: 'kg', limit: 13 }]
+    }
+
+    result = OptimizerWrapper.wrapper_vrp('demo', { services: { vrp: [:ortools] }}, TestHelper.create(problem), nil)
+    assert_equal 2, result[:routes].size
```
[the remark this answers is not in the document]
senhalil: Actually we can remove this check, what if the solver returns 1 route with some services unassigned... the test will give false negatives because of this check.

All we care in this tests is the check you already did with the number of activities and maybe we can add a check from the `problem` point of view to be sure that we didn't mess the quantity values internally on the solution we return and the following is still correct (but your check even covers this case, I think) 

`route[:activities].select{ |a| a[:service_id] }.sum{ |a| problem[:services].find{ |s| s[:id] == a[:id]}[:quantities].first[:value] } < 13`

wrappers/vroom.rb
```diff
@@ -267,8 +269,14 @@ def collect_jobs(vrp, vrp_skills, vrp_units)
             service.sticky_vehicles.flat_map{ |sticky| vrp_skills.find_index{ |sk| sk == sticky.id } }.compact,
           priority: (100 * (8 - service.priority).to_f / 8).to_i, # Scale from 0 to 100 (higher is more important)
           time_windows: service.activity.timewindows.map{ |timewindow| [timewindow.start, timewindow.end || 2**30] },
-          delivery: vrp_units.map{ |unit| service.quantities.find{ |quantity| quantity.unit.id == unit.id && quantity.value.negative? }&.value&.to_i || 0 },
-          pickup: vrp_units.map{ |unit| service.quantities.find{ |quantity| quantity.unit.id == unit.id && quantity.value.positive? }&.value&.to_i || 0 }
+          delivery: vrp_units.map{ |unit|
+            q = service.quantities.find{ |quantity| quantity.unit.id == unit.id && quantity.value.negative? }
+            ((q&.value || 0) * CUSTOM_QUANTITY_BIGNUM).round
```
fonsecadeline, line 274: Shouldn't we use _ceil_ rather than _round_ ? If every quantity rounding is equivalent to _quantity.floor_ it might generate quantity overruns.

test/wrappers/vroom_test.rb
```diff
@@ -770,4 +770,27 @@ def test_correct_route_collection
...
+    problem = VRP.lat_lon_two_vehicles
+    problem[:services].each{ |service|
+      service[:quantities] = [{ unit_id: 'kg', value: 1.001 }]
+    }
+    problem[:vehicles].each{ |vehicle|
+      vehicle[:capacities] = [{ unit_id: 'kg', limit: 13 }]
+    }
+
+    result = OptimizerWrapper.wrapper_vrp('demo', { services: { vrp: [:vroom] }}, TestHelper.create(problem), nil)
+    assert_equal 2, result[:routes].size
```
senhalil, line 784: the same comments I made for the ortools test

wrappers/vroom.rb
braktar: We currently only provide a 1e-3 precision. below this limit a ceil will surely ensure the feasibility but may impeach some quantities combinations. Anyway, it is below our precision scope

The closest choice we have is rounding for time matrix in optimizer-ortools, we also chose a rounding over a ceil.

senhalil: I agree, to make sure, we might ned to do something but `ceil` is not a good idea, the issue is floating point precision. 

```ruby
irb(main):017:0> 2.011 * 1e3
=> 2011.0000000000002
irb(main):018:0> (2.011 * 1e3).ceil
=> 2012
```
Clearly we don't want 2012 here. 

I think you are misinterpreting the `round` , the`round` is there to prevent floating-point errors (and making sure we get 2011 for the above case). 

With Gwen's dev we are supporting up to 3 decimal places, so the rest of the decimals are dropped and as you said, we might generate violation if the user really uses more than 3 decimal places in their quantities/capacities -- but we can't guarantee what we don't support.

To use ceil, we need to do wizardry 
```ruby
irb(main):021:0> ((2.011 * 1e6).round / 1e3).ceil
=> 2011
```
but I am not sure if it worth the hassle.

test/wrappers/ortools_test.rb
```diff
@@ -5770,4 +5770,31 @@ def test_self_selection_should_not_change_vehicle_ids
...
+      service[:quantities] = [{ unit_id: 'kg', value: 1.001 }]
+    }
+    problem[:vehicles].each{ |vehicle|
+      vehicle[:capacities] = [{ unit_id: 'kg', limit: 13 }]
+    }
+
+    result = OptimizerWrapper.wrapper_vrp('demo', { services: { vrp: [:ortools] }}, TestHelper.create(problem), nil)
+    assert_equal 13, result[:routes].map{ |route| route[:activities].count{ |activity| activity[:service_id] } }.sum,
+                 'The problem and result are expected to contain 13 services'
+
```
senhalil, line 5786: We don't care about this in this test, if (for any reason, like performance of the machine etc) or-tools or vroom returns a single woute with 12 services planned and one unplanned, it is not violating the `test_quantity_precision` so the test should pass. We have proper performance tests for this.

test/wrappers/ortools_test.rb
```diff
@@ -5770,4 +5770,31 @@ def test_self_selection_should_not_change_vehicle_ids
...
+      assert_operator route[:activities].count{ |activity| activity[:service_id] }, :<=, 12,
+                      'Routes cannot load more than 12 services'
+      collected_quantities = route[:activities].map{ |activity|
+        next unless activity[:service_id]
+
+        assert_equal 1.001, activity[:detail][:quantities].first[:value]
+        activity[:detail][:quantities].first[:value]
+      }.compact.sum
+      assert_operator collected_quantities, :<, 13,
+                      'Routes cannot load more than 13 kg'
```
senhalil, line 5797: last assert is not necessary it is a double of the first asserts. 

if there are at most 12 activities and all are equal to 1.001 then the sum cannot be more than 12 * 1.001 (which is always < 13)

senhalil:
@braktar can you update the doc so that the description of `value` for quantity and capacity reflects the fact that we support maximum 3 decimal places.

test/wrappers/ortools_test.rb
braktar: https://github.com/Mapotempo/optimizer-api/pull/213#discussion_r626393005

As asked by @fonsecadeline whenever we change the instance, the test will become obsolete and nothing will break

fonsecadeline: Right, I was suggesting something more like 

```suggestion
    assert_equal 13, result[:routes].map{ |route| route[:activities].count{ |activity| activity[:service_id] } }.sum + result[:unassigned].size,
                 'The problem and result are expected to contain 13 services'

```

You can eventually add something like 

```
assert result[:routes].any?{ |r| r[:activities].any{ |a| a[:service_id] } }
```

to make sure we were able to assign services but we did not overpass limit.

fonsecadeline: Maybe we can just but a limit of 4 and avoid all these checks (because I do not think we would reduce instance with less than 4 services).

senhalil: I didn't think of the case of empty solution, so yeah we can have something like that (`assert result[:routes].any?{ |r| r[:activities].any?{ |a| a[:service_id] } }, 'All services are unplanned'`)

The first one doesn't work if everything is unassigned... and we have that exact check in the wrapper_vrp for everything we optimise.

(We should never change the instances exactly because of this reason, these are "fixed" instances, except for obvious/simple correction, we never touch the VRP.instance type instances)

senhalil: This (no change) is the rule on the mt-web side for their fixed instances as well; otherwise, you lose the validity of all the tests using the instance.

test/wrappers/ortools_test.rb
```diff
@@ -5770,4 +5770,27 @@ def test_self_selection_should_not_change_vehicle_ids
...
+    problem[:vehicles].each{ |vehicle|
+      vehicle[:capacities] = [{ unit_id: 'kg', limit: 3 }]
+    }
+
+    result = OptimizerWrapper.wrapper_vrp('demo', { services: { vrp: [:ortools] }}, TestHelper.create(problem), nil)
+    assert_equal 2, result[:routes].first[:activities].count{ |activity| activity[:service_id] },
+                 'The route is expected to contain 2 services'
+    assert_equal 1, result[:unassigned].size, 'The result is expected to contain 1 unassigned'
+
+    assert_operator result[:routes].first[:activities].count{ |activity| activity[:service_id] }, :<=, 2,
```
fonsecadeline, line 5788: Isn't this equivalent to first assert you wrote in this test ?

test/wrappers/ortools_test.rb
```diff
@@ -5770,4 +5770,27 @@ def test_self_selection_should_not_change_vehicle_ids
...
+    problem[:services].each{ |service|
+      service[:quantities] = [{ unit_id: 'kg', value: 1.001 }]
+    }
+    problem[:vehicles].each{ |vehicle|
+      vehicle[:capacities] = [{ unit_id: 'kg', limit: 3 }]
+    }
+
+    result = OptimizerWrapper.wrapper_vrp('demo', { services: { vrp: [:ortools] }}, TestHelper.create(problem), nil)
+    assert_equal 2, result[:routes].first[:activities].count{ |activity| activity[:service_id] },
+                 'The route is expected to contain 2 services'
```
fonsecadeline, line 5785: I think there is an indentation problem here

test/wrappers/ortools_test.rb
fonsecadeline: In fact I think this only check is enough.

test/wrappers/ortools_test.rb
braktar: the second line is aligned with the first parameter which is `2`

test/wrappers/ortools_test.rb
```diff
@@ -5770,4 +5770,25 @@ def test_self_selection_should_not_change_vehicle_ids
...
+    problem = VRP.basic
+    problem[:services].each{ |service|
+      service[:quantities] = [{ unit_id: 'kg', value: 1.001 }]
+    }
+    problem[:vehicles].each{ |vehicle|
+      vehicle[:capacities] = [{ unit_id: 'kg', limit: 3 }]
+    }
+
+    result = OptimizerWrapper.wrapper_vrp('demo', { services: { vrp: [:ortools] }}, TestHelper.create(problem), nil)
+    assert_equal 1, result[:unassigned].size, 'The result is expected to contain 1 unassigned'
```
senhalil, line 5784: why are we failing the test if there is more than 1 unassinged. This makes the test "flaky".

All we know is that there should be at least one unassigned due to capacity constraint. 
```suggestion
    assert_operator  result[:unassigned].size, :>=, 1, 'The result is expected to contain at least 1 unassigned due to capacity'
```

senhalil (changes requested):

test/wrappers/ortools_test.rb
braktar: If we have more than one unassigned, we have probably applied a multiplier on the quantities and not on the capacities.

senhalil: I think we should have a proper unit_test (instead of a wrapper_vrp) for that -- for example "stub"ing the solve and checking if a value 1.003 becomes 1003 or not. This is not the right way, wrapper_vrp is a big black box and this test adds more false negatives than protection.

senhalil: not solve, run_solver..

senhalil (approved):
I think we should have a proper unit test that validates the functions doing the multiplication (i.e., `collect_jobs` etc) and verifies that fractional quantity values are correctly transformed. And in the functionality test, we should test only the necessary checks.

fonsecadeline (approved):

PR #214, opened 2021-05-05
Use Mapotempo balanced_vrp_clustering gem back

fonsecadeline:


Gemfile.lock
```diff
@@ -7,8 +7,8 @@ GIT
       activesupport (>= 5.0.0)
 
 GIT
-  remote: https://github.com/senhalil/balanced_vrp_clustering.git
-  revision: 657b71af8721cbbae44c5bba36790fb1c1d6842f
```
senhalil, line 11: hmmm if hash is changing that's not good. That means the code is not at the same state. Can you find the diff and make sure that it will not break anything? Actually that is a waste of time tbh, we already released v1.7 (and touching this dependency without testing is risky so we can't attach it to v1.7) plus I will work on clustering gem in the next version, so before this modification can be released we will modify it again. So we can just leave it as it is instead of taking a risk. What do you @fonsecadeline ?

fonsecadeline: If I run `git diff mapotempo/dev 657b71af8721cbbae44c5bba36790fb1c1d6842f` there is no difference. 

I guess hash changes because we changed repository.

I am okay to wait to make more tests before we go back to Mapotempo version, there is no hurry. I just noticed that condition to go back to this version was satisfied.

braktar: Can't follow a tag instead of a branch or a hash ?

senhalil: Fabien prefers that we use branch because it is faster and it gives the same protection as hash, but I agree when we reach a more stable version we should switch to tag/release logic.

senhalil (approved):

PR #217, opened 2021-05-07
Allow synchrone if threshold is zero

fonsecadeline:


optimizer_wrapper.rb
```diff
@@ -890,14 +890,14 @@ def self.apply_zones(vrp)
   end
 
   def self.clique_cluster(vrp, cluster_threshold, force_cluster)
-    if vrp.matrices.size.positive? && vrp.shipments.size.zero? && (cluster_threshold.to_f.positive? || force_cluster) && !vrp.scheduling?
+    if vrp.matrices.any? && vrp.shipments.empty? && (cluster_threshold || force_cluster) && !vrp.scheduling?
```
senhalil, line 893: I know it is not really in the socpe of this dev but now that 0 will call clique_cluster we need to fix the condition (vis-a-vis shipmets) because up to today mt-web was protected from clique_cluster (with or without shipment) because it always sends `0` but now `0` will call clique_cluster so we risk calling clique_cluster with shipments if we don't fix it in this PR because mt-web always sends shipments as relations. 

we need to add `vrp.relations.none?{ |r| r.type == :shipment }` (or something similar) to this condition

optimizer_wrapper.rb
```diff
@@ -890,14 +890,14 @@ def self.apply_zones(vrp)
...
       raise UnsupportedProblemError('Threshold is not supported yet if one service has serveral activies.') if vrp.services.any?{ |s| s.activities.size.positive? }
 
       original_services = Array.new(vrp.services.size){ |i| vrp.services[i].clone }
       zip_key = zip_cluster(vrp, cluster_threshold, force_cluster)
     end
     result = yield(vrp)
-    if !vrp.matrices.empty? && vrp.shipments.empty? && (cluster_threshold.to_f.positive? || force_cluster) && !vrp.scheduling?
+    if vrp.matrices.any? && vrp.shipments.empty? && (cluster_threshold || force_cluster) && !vrp.scheduling?
```
senhalil, line 900: and here

senhalil (changes requested):
except the mt-web shipment + clique_cluster issue, it is :rocket:

fonsecadeline:
@senhalil we used to forbid _evaluate_only_ option when we had shipments but I see that **test_simplified_pause_returns_the_same_cost** calls evaluate only with shipment relation. This does not seem to be a problem.

Do you agree that I can remove this assert ?

senhalil (changes requested):
This PR was about sync vs async originally but we added modifications related to shipments. Is there a specific reason to not to have a separate PR because they are not related at all and shipment modification needs more work but `sync vs async` dev looks finished.

wrappers/ortools.rb
```diff
@@ -50,7 +50,6 @@ def solver_constraints
         :assert_only_force_centroids_if_kmeans_method,
         :assert_no_scheduling_if_evaluation,
         :assert_route_if_evaluation,
-        :assert_no_shipments_if_evaluation,
```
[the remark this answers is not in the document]
fonsecadeline: I did test it and it works well.

fonsecadeline:
@senhalil the reason why I am touching sync vs async AND shipments is because I wanted to use pickup_delivery? in my condition https://github.com/Mapotempo/optimizer-api/pull/217/commits/6acf41e45255f9002e659e5e4388b7d2104ae082#diff-6b6fec5242129149dd15ae48b6f6e8f383e2b202c6e9bc35357552bd2d9b4eb2R893 .

I apparently changed my mind after introducing pickup_delivery? so I can definetely create another PR so we can make sure these changes are clean.

models/vrp.rb
```diff
@@ -777,5 +777,10 @@ def scheduling?
     def periodic_heuristic?
       self.preprocessing_first_solution_strategy.to_a.include?('periodic')
     end
+
+    def pickup_delivery?
```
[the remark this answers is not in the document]
fonsecadeline: rubocop suggests vrp.shipments?

fonsecadeline:
> This PR was about sync vs async originally but we added modifications related to shipments. Is there a specific reason to not to have a separate PR because they are not related at all and shipment modification needs more work but `sync vs async` dev looks finished.

I created a new PR : https://github.com/Mapotempo/optimizer-api/pull/222

fonsecadeline:
@fab-girard @senhalil @braktar 

- [ ] we need to discuss it better : _zip_cluster_ ignores relations. So in fact calling _zip_cluster_ when _threshold_ is zero can be dangerous because mapotempo-web is sending 0 and we do not want to ignore relations from mapotempo-web.

@fab-girard tryied sending nil insteaf of 0 from mapotempo-web.

fonsecadeline:
See https://gitlab.com/mapotempo/optimizer-api/-/issues/798

braktar (approved):

optimizer_wrapper.rb
```diff
@@ -101,7 +101,7 @@ def self.wrapper_vrp(api_key, services, vrp, checksum, job_id = nil)
       raise UnsupportedProblemError.new('Cannot apply any of the solver services', inapplicable_services)
     elsif config[:solve][:synchronously] || (
             services_vrps.size == 1 &&
-            !vrp.preprocessing_cluster_threshold &&
+            vrp.preprocessing_cluster_threshold.to_f == 0.0 &&
```
senhalil, line 104: I think the previous version is correct, no? 

It has been some time but now that zip_clustering will be active when threshold is 0, we don't want to activate it for 0 and we want async only for `nil`

senhalil (changes requested):
- [x] Also, we need to make sure that mt-web changed the default value of this parameter and it sends `nil` and not `0` by default

optimizer_wrapper.rb
```diff
@@ -777,7 +777,7 @@ def self.apply_zones(vrp)
   end
 
   def self.clique_cluster(vrp, cluster_threshold, force_cluster)
-    zip_condition = vrp.matrices.any? && (cluster_threshold.to_f.positive? || force_cluster) && !vrp.schedule?
+    zip_condition = vrp.matrices.any? && (cluster_threshold || force_cluster) && !vrp.schedule?
 
     if zip_condition
       if vrp.services.any?{ |s| s.activities.any? }
```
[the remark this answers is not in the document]
fonsecadeline: Right, we used to have something for shipments and we removed it in https://github.com/Mapotempo/optimizer-api/pull/240/files#diff-6b6fec5242129149dd15ae48b6f6e8f383e2b202c6e9bc35357552bd2d9b4eb2R780 so I guess it just disappeared when rebasing. 

I think I had made a similar comment by the time, but anyways. 

Just to make sure we are one the same page, you suggest that we raise as soon as one service needs to be merged with another one, and they are both included in a (different) relation which type is in : 
* same_vehicle, 
* same_route, 
* sequence,
* order
* shipment 
* meetup
Is that right ?

optimizer_wrapper.rb
fonsecadeline: I could not find the comment but I remember that we said that if cluster_threshold is zero then zip_cluster would be fast enough to be solve synchronously if other parameters let it.

Still, I agree it is easier for us to know that zip_clusters is always callend in async. Moreover, it is safer because we might add cost_full checks or operations in zip_clusters without noticing they will be done synchronously.

senhalil: Now that you mention I vaguely remember (we first had to wait for mt-web defaults to get corrected and then this PR fell in the cracks).

Yeah, I think your argument is even stronger so lets skip synch if it is zip_cluster

optimizer_wrapper.rb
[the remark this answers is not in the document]
senhalil: I skimmed the zip_cluster code and didn't see any logic to handle the relations of merged services (can you verify please in case I missed something).

Yeah similar to that and I think raise is more aptly (similar to how it is done vis-a-vis activities, just below the code you shared) because otherwise we silently ignore the threshold setting.

My suggestion was more rudimentary (and easy) -- like it is already done vis-a-vis service.activities, we don't verify if a services with activities is getting merged or not, we just refuse cluster_threshold option -- i.e., if zip_condition is true but we have `vrp.relations.any?{ |r| RELATIONS_ZIP_CLUSTER_CAN_HANDLE.exclude?(r.type) }` then we raise `clsuter threshold option is compatible only with the following relations: #{RELATIONS_ZIP_CLUSTER_CAN_HANDLE}` or something like that.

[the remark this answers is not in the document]
senhalil: And `RELATIONS_ZIP_CLUSTER_CAN_HANDLE` is all the relations that doesn't get affected by merging services (i.e., the ones that only effect vehicles). This way, when we add a new relation types, the refusal will be handled automatically.

config/environments/development.rb
```diff
@@ -40,6 +40,8 @@ module OptimizerWrapper
   WEEKDAYS = %i[mon tue wed thu fri sat sun].freeze
   DEMO = Wrappers::Demo.new(tmp_dir: TMP_DIR)
   VROOM = Wrappers::Vroom.new(tmp_dir: TMP_DIR, threads: 1)
+  RELATION_ZIP_CLUSTER_CAN_HANDLE = %i[minimum_day_lapse maximum_day_lapse minimum_duration_lapse maximum_duration_lapse force_first never_first force_end
```
braktar, line 43: zip cluster cannot ensure that force_first, never_first or force_end are respected if multiple services with different position preference are within the same zip cluster

braktar (changes requested):

senhalil:
- [ ] Needs https://github.com/Mapotempo/optimizer-api/pull/321 to be merged

senhalil (approved):

optimizer_wrapper.rb
```diff
@@ -20,6 +20,10 @@
 require_all 'lib'
 require_all 'util'
 
+RELATION_ZIP_CLUSTER_CAN_HANDLE = %i[minimum_day_lapse maximum_day_lapse minimum_duration_lapse maximum_duration_lapse
+                                     vehicle_trips vehicle_group_duration vehicle_group_duration_on_weeks
+                                     vehicle_group_duration_on_months vehicle_group_number].freeze
```
senhalil: Screenshot from 2022-03-11 16-00-00

Alphabetical order and new lines for easy maintenance otherwise it leads to :point_up: 
```suggestion
RELATION_ZIP_CLUSTER_CAN_HANDLE = %i[
  maximum_day_lapse
  maximum_duration_lapse
  minimum_day_lapse
  minimum_duration_lapse
  vehicle_group_duration
  vehicle_group_duration_on_months
  vehicle_group_duration_on_weeks
  vehicle_group_number
  vehicle_trips
].freeze
```

lib/tsp_helper.rb
```diff
@@ -36,7 +36,8 @@ def self.create_tsp(vrp, options = {})
       id: "#{tsp_suffix}_#{vehicle.id}",
       start_point_id: start_point && "#{tsp_suffix}_#{start_point.id}",
       end_point_id: end_point && "#{tsp_suffix}_#{end_point.id}",
-      matrix_id: vehicle.matrix_id
+      matrix_id: vehicle.matrix_id,
+      timewindow: { start: options[:begin_time] || 0 }
```
senhalil:
```suggestion
      timewindow: { start: options[:begin_time] || 0 },
```

senhalil (approved):

PR #218, opened 2021-05-17
Fix nil capacities when vroom is used

braktar:


test/wrappers/vroom_test.rb
```diff
@@ -791,4 +791,19 @@ def test_quantity_precision
...
+      service[:quantities] = [{ unit_id: 'kg', value: 1 }]
+    }
+    problem[:vehicles].first[:capacities] = [{ unit_id: 'kg', limit: 3 }]
+
+    result = OptimizerWrapper.wrapper_vrp('demo', { services: { vrp: [:vroom] }}, TestHelper.create(problem), nil)
+
+    assert_operator result[:routes].first[:activities].count{ |activity| activity[:service_id] }, :<=, 3,
+                    'The vehicle cannot load more than 3 services and 3 kg'
+
+    assert_empty result[:unassigned], 'The second vehicle has no capacity limit and should the remaining services'
```
fonsecadeline, line 807:
```suggestion
    assert_empty result[:unassigned], 'The second vehicle has no capacity limit and should be assigned to the remaining services'
```

fonsecadeline: There is only 1 vehicle in this test, and 3 services. There is no second vehicle. Shouldn't we have more than 3 services and one additional vehicle ?

fonsecadeline (changes requested):

test/wrappers/vroom_test.rb
braktar: Oops, I thought we were in the previous configuration of the above test (11 services and 2 vehicles)

test/wrappers/vroom_test.rb
```diff
@@ -791,4 +791,26 @@ def test_quantity_precision
...
+    vroom.stub(
+      :run_vroom, lambda{ |vroom_vrp, _job|
+        assert_equal [2 * Wrappers::Vroom::CUSTOM_QUANTITY_BIGNUM], vroom_vrp[:vehicles].first[:capacity]
+        assert_equal vroom_vrp[:jobs].flat_map{ |job| job[:pickup].first }.sum,
+                     vroom_vrp[:vehicles].last[:capacity].first
+        nil
+      }
+    ) do
+      vroom.solve(TestHelper.create(problem))
+    end
```
fonsecadeline, line 814: I am not sure about second assert. What are we checking ? I would say we need to check result.

```suggestion
   vroom_vrp = vroom.vroom_problem(vrp, [:time, :distance])
   expected_capacity = problem[:vehicles].first[:capacitities].first[:limit]
   assert_equal [expected_capacity * Wrappers::Vroom::CUSTOM_QUANTITY_BIGNUM], vroom_vrp[:vehicles].first[:capacity]
   
   result = vroom.run_vroom(vroom_vrp, nil)
   assert_operator result['routes'].first.sum{ |a| ['service', 'pickup', 'delivery'].include?('a[:type]) ? a[:detail][:quantities].first[:value] : 0 }, :<=, problem[:vehicles].first[:capacity].first[:limit]
```

braktar: We are checking that the capacity of the vehicle with a capacity set to nil is equal to the sum of all the demands

braktar: Vroom is not called here

fonsecadeline: Oh okay I read 'first' instead of 'last', sorry :)

fonsecadeline (approved):

fab-girard (approved):

PR #223, opened 2021-05-27
Fix: Vroom handles quantities correctly

senhalil:
Vroom expects all quantities to be positive (delivery quantities are multiplied with -1)
It cannot handle durations begin modified by vehicle (assert added)
It shouldn't silently ignore services (both pickup and delivery amounts are sent to the solver)

### DIFF
--- a/wrappers/vroom.rb
+++ b/wrappers/vroom.rb
@@ -274,7 +274,7 @@ module Wrappers
           delivery: vrp_units.map{ |unit|
             q = service.quantities.find{ |quantity| quantity.unit.id == unit.id && quantity.value.negative? }
             @total_quantities[unit.id] -= q&.value || 0
-            ((q&.value || 0) * CUSTOM_QUANTITY_BIGNUM).round
+            (-(q&.value || 0) * CUSTOM_QUANTITY_BIGNUM).round
           },
           pickup: vrp_units.map{ |unit|
             q = service.quantities.find{ |quantity| quantity.unit.id == unit.id && quantity.value.positive? }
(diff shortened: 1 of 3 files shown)


fonsecadeline (approved):

wrappers/vroom.rb
```diff
@@ -293,9 +291,9 @@ def collect_shipments(vrp, vrp_skills, vrp_units)
       vrp.shipments.map.with_index{ |shipment, index|
         {
           amount: vrp_units.map{ |unit|
-            q = shipment.quantities.find{ |quantity| quantity.unit.id == unit.id && quantity.value&.positive? }
-            @total_quantities[unit.id] += q&.value || 0
-            ((q&.value || 0) * CUSTOM_QUANTITY_BIGNUM).round
+            value = shipment.quantities.find{ |quantity| quantity.unit.id == unit.id }&.value.to_f.abs
```
braktar, line 294: if the result of find is nil .value won't be invoked neither as .to_f and .abs

braktar (approved):

senhalil:
@braktar share me the test you have on your local when you have time and I will add it the PR

braktar:
Here are the data i used. I didn't implement the asserts
```
    problem = VRP.basic
    problem[:units] << { id: 'l' }
    problem[:services].each{ |service|
      service[:quantities] = [{ unit_id: 'kg', value: 1 }, { unit_id: 'l', value: -1}]
    }
    problem[:vehicles].each{ |vehicle|
      vehicle[:capacities] = [{ unit_id: 'kg', limit: 3 }, { unit_id: 'l', limit: 2}]
    }
```

PR #224, opened 2021-06-01
Document output

braktar:


api/v01/vrp.rb
```diff
@@ -196,7 +210,9 @@ class Vrp < APIBase
 
           desc 'Delete vrp job', {
             nickname: 'deleteJob',
-            success: { code: 204 },
+            success: {
+              code: 202,
+              model: VrpResult},
```
fonsecadeline, line 215:
```suggestion
            success: {
              code: 202,
              model: VrpResult
            },
```

fonsecadeline (approved):

fonsecadeline (approved):

api/v01/vrp.rb
```diff
@@ -75,10 +75,32 @@ class Vrp < APIBase
...
+                code: 400,
+                message: 'Bad Request',
+                model: ::Api::V01::Status
+              }, {
+                code: 401,
+                message: 'Unauthorized',
+                model: ::Api::V01::Status
+              }, {
+                code: 402,
+                message: 'Subscription expired',
```
senhalil, line 97: We can add a message like 'Please contact support to renew your subscription.' , what do you think?

Also should we give an email <xxx@mapotempo.com> and/or phone <+33xxx> with these messages (in addition to or instead of just saying support/sales)?

api/v01/vrp.rb
```diff
@@ -75,10 +75,32 @@ class Vrp < APIBase
...
+                model: ::Api::V01::Status
+              }, {
+                code: 413,
+                message: 'Exceeded limit authorized for your account. Please contact support or sales to increase limits.',
```
senhalil, line 101: We do not need to do exactly the same but for user quotas google uses [402](https://developers.google.com/webmaster-tools/search-console-api-original/v3/errors#PAYMENT_REQUIRED) and [413](https://developers.google.com/webmaster-tools/search-console-api-original/v3/errors#REQUEST_ENTITY_TOO_LARGE) is literally related to size of the data.

senhalil: But again they differentiate between the http status code and a separate error code. We don't have an error code but we give the error message directly..

senhalil (commented):
This is good :+1: I have some suggestions but they are not

api/v01/vrp.rb
braktar: The current PR has only for purpose to document all the http error codes. We currently have no such shade on errors except for http code 500 such as ClusteringError, DiscordantProblemError, JobKilledError, SchedulingHeuristicError or UnsupportedRouterModeError. But error code 500 probably requires its own PR.

Moreover grape-swagger and swagger-ui doesn't look like to fulfill this need.

fonsecadeline (approved):

senhalil:
Failing test is not related. We can merge it wo re-test.
```
 FAIL["test_detecting_unfeasible_services_can_not_take_too_long", #<Minitest::Reporters::Suite:0x00005622133b5c50 @name="WrapperTest">, 30.50278225599999]
 test_detecting_unfeasible_services_can_not_take_too_long#WrapperTest (30.50s)
        check_distances function took longer than expected.
        Expected 3.0710660050000005 to be <= 3.0.
```

PR #226, opened 2021-06-03
ensure simplify pause works even without timewindow end

fonsecadeline:


senhalil (approved):

senhalil:
it looks like there still is an issue 
```
 ERROR["test_rest_with_exclusion_cost", #<Minitest::Reporters::Suite:0x000055feec0a3688 @name="Wrappers::OrtoolsTest">, 116.55539198600002]
 test_rest_with_exclusion_cost#Wrappers::OrtoolsTest (116.56s)
Minitest::UnexpectedError:         NoMethodError: undefined method `[]' for nil:NilClass
            /srv/app/wrappers/wrapper.rb:1026:in `block in simplify_vehicle_pause'
```

fonsecadeline:
> it looks like there still is an issue
> 
> ```
>  ERROR["test_rest_with_exclusion_cost", #<Minitest::Reporters::Suite:0x000055feec0a3688 @name="Wrappers::OrtoolsTest">, 116.55539198600002]
>  test_rest_with_exclusion_cost#Wrappers::OrtoolsTest (116.56s)
> Minitest::UnexpectedError:         NoMethodError: undefined method `[]' for nil:NilClass
>             /srv/app/wrappers/wrapper.rb:1026:in `block in simplify_vehicle_pause'
> ```

I saw that. This failed twice but it works in local. I will look at this deeper.

senhalil:
Oh yeah no worries, I realised there was a broken test after validating the PR and wanted to make a note of it. 

If it doesn't break on local that is pretty weird. Good luck :muscle: 

By the way, I just noticed that I completely forgot the fact that `pause` can have `exclusion_cost` during the `simply_pause` dev.  

If the pause has an `exclusion_cost`, we should not simplify it (or we need to improve the `patch_result` function) because the decision of inserting or not inserting it depends on the actual costs. 

I think it is better if we don't simplify the pauses with exclusion cost in the first place because they are not really "simple" :) what do you think?


line 945 has the following check
```
next if vehicle.rests.size > 1 || vehicle.rests.any?{ |r| r.timewindows&.size.to_i > 1 }
```
which needs to be modified to be
```
next if vehicle.rests.size > 1 || vehicle.rests.any?{ |r| r.timewindows&.size.to_i > 1 || r.exclusion_cost.to_f.positive? }
``` 
or something  like that (not tested)

braktar (approved):

PR #227, opened 2021-06-04
Scheduling move checks to find best index

fonsecadeline:
- [x] needs https://github.com/Mapotempo/optimizer-api/pull/194

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 ### Changed
 
+- Improve cases where a service has two visits in periodic heuristic: ensure that the second visit can be assigned to the right day [#227](https://github.com/Mapotempo/optimizer-api/pull/227)
+
 ### Removed
 
 - Field `trips` in vehicle model. Use `vehicle_trips` relation instead [#123](https://github.com/Mapotempo/optimizer-api/pull/123)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 ### Fixed
 
 - VROOM was used incorrectly in various cases: negative quantities, vehicle duration, activity position [#223](https://github.com/Mapotempo/optimizer-api/pull/223) [#242](https://github.com/Mapotempo/optimizer-api/pull/242)
+- Capacity violation in periodic heuristic algorithm (`first_solution_strategy='periodic'`) [#227](https://github.com/Mapotempo/optimizer-api/pull/227)
 
 ## [v1.7.1] - 2021-05-20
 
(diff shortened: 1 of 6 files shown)


braktar (approved):

CHANGELOG.md
```diff
@@ -12,12 +12,16 @@
 
 ### Changed
 
+- Improve case where service has two visits in periodic heuristic : ensure second visit can be assigned to right day [#227](https://github.com/Mapotempo/optimizer-api/pull/227)
```
senhalil, line 15:
```suggestion
- Improve cases where a service has two visits in periodic heuristic: ensure that the second visit can be assigned to the right day [#227](https://github.com/Mapotempo/optimizer-api/pull/227)
```

CHANGELOG.md
```diff
@@ -12,12 +12,16 @@
...
+
 ### Removed
 
 - Field `trips` in vehicle model. Use `vehicle_trips` relation instead [#123](https://github.com/Mapotempo/optimizer-api/pull/123)
 
 ### Fixed
 
+- Capacity violation from periodic heuristic [#227](https://github.com/Mapotempo/optimizer-api/pull/227)
```
senhalil, line 23:
```suggestion
- Capacity violation in periodic heuristic algorithm (`first_solution_strategy='periodic'`) [#227](https://github.com/Mapotempo/optimizer-api/pull/227)
```

lib/heuristics/concerns/scheduling_end_phase.rb
```diff
@@ -120,15 +120,14 @@ def add_missing_visits
   def update_costs(costs, best_cost)
     # update costs for inserted id, available_days changed
     uninserted_set = @uninserted.select{ |_key, info| info[:original_id] == best_cost[0] }.keys
-    @uninserted.delete(uninserted_set.first)
+    @uninserted.delete(uninserted_set.first) # faire en fct nb in seq
```
senhalil, line 123: I can't really understand the comment.

senhalil (changes requested):
Maybe I missed but I didn't see any unit or functionality non-regression tests against the capacity violation bug. Can we add a tests that breaks without this PR and gets fixed with it?

fonsecadeline:
@senhalil in fact it is not possible to create a code that would fail without these edits, because checks were correctly done (repeatedly) everywhere (like in find_best_day_cost for instance). 

Now this capacity check is called at only once place and we are sure we are going to call it everytime we compute cost of inserting a service in a route. It was not the case before and I generated capacity violation while working on same_point_day improve. That is why I generated this PR.

senhalil (approved):

PR #229, opened 2021-06-08
Keep matrix when dumping

fonsecadeline:


braktar (approved):

test/test_helper.rb
```diff
@@ -120,14 +120,31 @@ def self.create(problem)
...
+    end
+
     File.write("test/fixtures/#{vrp.name}.dump", Zlib.deflate(Marshal.dump(vrp)))
   end
 
   def self.multipe_matrices_required(vrps, filename)
     return unless ENV['TEST_DUMP_VRP'].to_s == 'true'
 
-    vrps.each(&:compute_matrix)
+    if ENV['KEEP_DUMPED_MATRIX'] == 'true'
```
braktar, line 138: We could maybe factorize single and multiple vrps managment ?

test/api/v01/output_test.rb
```diff
@@ -192,7 +192,7 @@ def test_files_generated
     OptimizerWrapper.config[:debug][:output_clusters] = true
     OptimizerWrapper.config[:debug][:output_schedule] = true
 
-    vrp = TestHelper.load_vrp(self, fixture_file: 'instance_clustered')
```
braktar, line 195: the method name `load_vrps` is implicitely expected to return an array. Assigning a simple `vrp` seems odd.

I would prefer to keep both `load_vrp` and `load_vrps` but to factorize their codes.

or

to always return an array of vrps

braktar (changes requested):

test/api/v01/output_test.rb
fonsecadeline: I could keep TestHelper.load_vrp function but simply call TestHelper.load_vrps in it and just keep the first one. What do you think ? Something like : 

```
  def self.load_vrp(test, options = {})
    load_vrp(test, options)[0]
  end
```

braktar: I agree !

test/test_helper.rb
```diff
@@ -117,61 +117,50 @@ def self.create(problem)
...
     else
       json_file = "test/fixtures/#{filename}.json"
-      vrps = options[:problem] || JSON.parse(File.read(json_file), symbolize_names: true).map{ |vrp| vrp[:vrp] }
-
-      vrps.map!{ |vrp|
-        create(vrp)
-      }
+      vrps = options[:problem] ||
+             JSON.parse(File.read(json_file), symbolize_names: true).map{ |vrp|
+              create(vrp[:vrp])
```
braktar, line 160: There is a missing space before create :
rubocop: warning
Layout/IndentationWidth - Use 2 (not 1) spaces for indentation.

braktar (changes requested):

braktar (approved):

test/test_helper.rb
```diff
@@ -131,66 +131,60 @@ def self.create(problem)
...
+      if dumped_data.nil? && ENV['TEST_DUMP_VRP'] != 'true'
+        # WebMock.enable_net_connect!
+        vrp.compute_matrix
+        # WebMock.disable_net_connect!
+      else
+        OptimizerWrapper.router.stub(:matrix, lambda { |url, mode, dimensions, row, column, options|
+          puts 'stubed fct'
+          if ENV['TEST_DUMP_VRP'] == 'true'
+            matrices =
+              OptimizerWrapper.router.send(:__minitest_stub__matrix, url, mode, dimensions, row, column, options) # call original method
```
braktar, line 151:
```suggestion
        Routers::RouterWrapper.stub_any_instance(:matrix, lambda { |url, mode, dimensions, row, column, options|
```
What I understand here, is that you overload the current instance of the router loaded by the during all the test suite. So, it cannot be unloaded at the end of the stub. While Routers::RouterWrapper.stub_any_instance is not related to the context, so it can be removed once the current context is finished

test/test_helper.rb
```diff
@@ -131,66 +131,60 @@ def self.create(problem)
...
+      next if vrp.matrices.any?
 
-  def self.multipe_matrices_required(vrps, filename)
-    return unless ENV['TEST_DUMP_VRP'].to_s == 'true'
+      if dumped_data.nil? && ENV['TEST_DUMP_VRP'] != 'true'
+        # WebMock.enable_net_connect!
+        vrp.compute_matrix
+        # WebMock.disable_net_connect!
+      else
+        OptimizerWrapper.router.stub(:matrix, lambda { |url, mode, dimensions, row, column, options|
```
braktar, line 147:
```suggestion
        Routers::RouterWrapper.stub_any_instance(:matrix, lambda { |url, mode, dimensions, row, column, options|
```
What I understand here, is that you overload the current instance of the router loaded by the during all the test suite. So, it cannot be unloaded at the end of the stub. While Routers::RouterWrapper.stub_any_instance is not related to the context, so it can be removed once the current context is finished

braktar (approved):

test/test_helper.rb
```diff
@@ -131,66 +131,61 @@ def self.create(problem)
...
+              { url: url, mode: mode, dimensions: dimensions, row: row, column: column, options: options, matrices: matrices }
+            matrices
+          else
+            corresponding_matrices = dumped_data.find{ |request|
+              # TODO : the day filter.rb (from models/vrp.rb) filters services then some test may not find
+              # their corresponding matrix anymore because row and/or column will be different
+              request[:mode] == mode && request[:dimensions] == dimensions &&
+                request[:row] == row && request[:column] == column && request[:options] == options
+            }[:matrices]
+            dimensions.collect.with_index{ |_dim, dim_i| corresponding_matrices[dim_i] }
```
senhalil, line 163:
```suggestion
              request[:mode] == mode && request[:dimensions] == dimensions &&
                request[:options] == options && request[:row] == row && request[:column] == column
            }[:matrices]
            # TODO: If services are filtered then some test may not find
            # their corresponding matrices because row and/or column will be different
            # In this case, we need to modify the above code and filter via lat/lon
            # to have the same order/content.
            raise 'Could not find matrix in the dump' if corresponding_matrices.nil?
            dimensions.collect.with_index{ |_dim, dim_i| corresponding_matrices[dim_i] }
```

senhalil: I added a raise, moved the find condition order and modified the comment

test/test_helper.rb
```diff
@@ -131,66 +131,61 @@ def self.create(problem)
...
+      next if vrp.matrices.any?
 
-  def self.multipe_matrices_required(vrps, filename)
-    return unless ENV['TEST_DUMP_VRP'].to_s == 'true'
+      if dumped_data.nil? && ENV['TEST_DUMP_VRP'] != 'true'
+        WebMock.enable_net_connect!
+        vrp.compute_matrix
+        WebMock.disable_net_connect!
+      else
+        Routers::RouterWrapper.stub_any_instance(:matrix, lambda { |url, mode, dimensions, row, column, options|
```
senhalil, line 149:
```suggestion
        OptimizerWrapper.router.stub(:matrix, lambda { |url, mode, dimensions, row, column, options|
```

test/lib/heuristics/scheduling_test.rb
```diff
@@ -515,13 +512,10 @@ def test_provide_initial_solution
 
     def test_fix_unfeasible_initial_solution
       vrp = TestHelper.load_vrp(self, fixture_file: 'instance_baleares2')
-      vrp.routes = Marshal.load(File.binread('test/fixtures/formatted_route.bindump')) # rubocop: disable Security/MarshalLoad
+      vrp.routes = [Models::Route.create(vehicle: vrp.vehicles.first, mission_ids: %w[5482 0833 8595 0352 0799 2047 5446 0726 0708], day_index: 0)]
```
senhalil, line 515: Thanks! This is a lot better than dumps, now we can see what the significance is :pray:

senhalil (approved):
perfect :ok_hand:

PR #230, opened 2021-06-09
Detect unfeasible timewindows

fonsecadeline:
Avoids case where we can not find an initial solution just because one vehicle has an invalid time window.


models/concerns/validate_timewindows.rb
```diff
@@ -1,43 +0,0 @@
...
-
-  private
-
-  def validate_timewindows
-    t = timewindows.collect{ |timewindow|
-      [timewindow.start, timewindow.end]
-    }
-
-    t.flatten!
-    if t.size > 2 && !t[1..-2].all?
```
braktar, line 35: How do we check now that all only first timewindow has a nil start and the last timewindow has a nil end ?

fonsecadeline: We do not check this now. Is this required ? This was never called before.

braktar: It is probably not necessary anymore as nil values are converted to integers and timewindows are merged if they overlap

braktar (approved):

senhalil (approved):

PR #231, opened 2021-06-09
Fix: missing route details during reinsertion of simplified pause

senhalil:
- In line empty_result function so that it has the same details of an empty route from a non-empty result
- Protect pause simplification routine against such missing details

### DIFF
--- a/wrappers/wrapper.rb
+++ b/wrappers/wrapper.rb
@@ -1188,7 +1188,21 @@ module Wrappers
         cost: nil,
         cost_details: Models::CostDetails.new({}),
         iterations: nil,
-        routes: vrp.vehicles.collect{ |vehicle| { vehicle_id: vehicle.id, activities: [] } },
+        routes: vrp.vehicles.collect{ |vehicle|
+          route_start_time = vehicle.timewindow&.start.to_i
+          route_end_time = route_start_time
+          {
+            vehicle_id: vehicle.id,
+            original_vehicle_id: vehicle.original_id,
+            cost_details: Models::CostDetails.new({}),
+            activities: [], # TODO: check if depot activities are needed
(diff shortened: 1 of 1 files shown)


wrappers/ortools.rb
```diff
@@ -465,19 +465,6 @@ def kill
 
     private
 
-    def build_cost_details(cost_details)
```
braktar, line 468: Here, the object cost_details is provided by the protobuf model returned by OR-Tools. Its parsing is specific to this solver

fonsecadeline: If we remove this function then we should remove it from vroom.rb too. But as @braktar said maybe it is not possible. 

I think we can have minimal build_cost_details in wrapper.rb and then overlad it and fill each key according to each solver specificities.

braktar: Wrapper.rb should implement the method and the solver wrappers should override it if necessary

senhalil: It turns out the solution was right in front of us... `Models::CostDetails.new({})` and we don't even need a `build_cost_details` function for `empty_result` generation.

By the way, I checked your suggestion and it doesn't work. Even if we have the function defined in `Wrappers::Wrapper` and `Wrappers::Ortools` since the empty_result is called via `config[:services][service].empty_result` -- i.e., from `Wrappers::Ortools` -- due to inheritance it behaves as a `Wrappers::Ortools` function (even though empty_result is only defined in `Wrappers::Wrapper`) and it calls the `build_cost_details` of `Wrappers::Ortools` in this case, instead of the vanilla version defined in `Wrappers::Wrapper`. Since all of our methods are instance methods I don't see a way around this without a design overhaul.

fonsecadeline (approved):

braktar (approved):

PR #234, opened 2021-06-11
Fix Vroom remove skills key if unused

braktar:


fonsecadeline (approved):

wrappers/vroom.rb
```diff
@@ -282,7 +282,7 @@ def collect_jobs(vrp, vrp_skills, vrp_units)
             ((q&.value || 0) * CUSTOM_QUANTITY_BIGNUM).round
           }
         }.delete_if{ |_k, v|
-          v.nil? || v.is_a?(Array) && v.empty?
+          !v || v.is_a?(Array) && v.empty?
```
senhalil, line 285: Does this related to the fix? It looks like a bug to me -- with `!v` if we ever add a field with a value `false` to this hash, it will be removed --  I find v.nil? more expressive and to the point.

senhalil (changes requested):

fonsecadeline (approved):

senhalil (approved):

PR #237, opened 2021-06-15
Fix travel times computation

braktar:
- [x] https://github.com/Mapotempo/optimizer-ortools/pull/21

### DIFF
--- a/lib/interpreters/split_clustering.rb
+++ b/lib/interpreters/split_clustering.rb
@@ -515,6 +515,9 @@ module Interpreters
 
         if load_flag && time_flag
           emptied_routes = true
+          %i[total_distance total_time total_travel_time total_value].each{ |key|
+            result[key] -= route[key] if route[key]
+          }
 
           number_of_services_in_the_route = route[:activities].map{ |a| a[:service_id] && a.slice(:service_id, :detail).compact }.compact.size
 
--- a/optimizer_wrapper.rb
+++ b/optimizer_wrapper.rb
@@ -156,7 +156,10 @@ module OptimizerWrapper
       }
     }
 
-    check_result_consistency(expected_activity_count, several_results) if services_vrps.collect{ |sv| sv[:service] } != [:demo] # demo solver returns a fixed solution
+    # demo solver returns a fixed solution
+    unless services_vrps.collect{ |sv| sv[:service] }.uniq == [:demo]
+      check_result_consistency(expected_activity_count, several_results)
+    end
 
     nb_routes = several_results.sum{ |result| result[:routes].count{ |r| r[:activities].any?{ |a| a[:service_id] } } }
     nb_unassigned = several_results.sum{ |result| result[:unassigned].size }
@@ -497,8 +500,38 @@ module OptimizerWrapper
 
   def self.check_result_consistency(expected_value, results)
     [results].flatten(1).each{ |result|
-      nb_assigned = result[:routes].sum{ |route| route[:activities].count{ |a| a[:service_id] } }
-      nb_unassigned = result[:unassigned].count{ |unassigned| unassigned[:service_id] }
+      if result[:routes].any?{ |route| route[:activities].any?{ |a| a[:waiting_time].to_i < 0 } }
+        log 'Computed waiting times are invalid', level: :warn
+        raise RuntimeError, 'Computed waiting times are invalid' if ENV['APP_ENV'] != 'production'
+      end
+
+      waiting_times = result[:routes].map{ |route| route[:total_waiting_time] }.compact
+      durations = result[:routes].map{ |route|
+        route[:activities].map{ |act| act[:departure_time] && (act[:departure_time] - act[:begin_time]) }.compact
+      }
+      setup_durations = result[:routes].map{ |route|
+        route[:activities].map{ |act|
+          next if act[:type] == 'rest'
+
+          (act[:travel_time].nil? || act[:travel_time]&.positive?) && act[:detail][:setup_duration] || 0
+        }.compact
+      }
+      total_time = result[:total_time] || 0
+      total_travel_time = result[:total_travel_time] || 0
+      if total_time != (total_travel_time || 0) +
+                       waiting_times.sum +
+                       (setup_durations.flatten.reduce(&:+) || 0) +
+                       (durations.flatten.reduce(&:+) || 0)
+
+        log_string = "Expected #{total_time} == #{total_travel_time} +"\
+                     " #{waiting_times.sum} + #{setup_durations.flatten.reduce(&:+)}"\
+                     " + #{durations.flatten.reduce(&:+)}"
+        log log_string, level: :warn
+        raise RuntimeError, 'Computed times are invalid' if ENV['APP_ENV'] != 'production'
+      end
+
+      nb_assigned = result[:routes].sum{ |route| route[:activities].count{ |a| a[:service_id] || a[:pickup_shipment_id] || a[:delivery_shipment_id] } }
+      nb_unassigned = result[:unassigned].count{ |unassigned| unassigned[:service_id] || unassigned[:pickup_shipment_id] || unassigned[:delivery_shipment_id] }
 
       if expected_value != nb_assigned + nb_unassigned # rubocop:disable Style/Next for error handling
         log "Expected: #{expected_value} Have: #{nb_assigned + nb_unassigned} activities"
@@ -550,6 +583,7 @@ module OptimizerWrapper
           activity[:current_distance] ||= total[dimension].round if dimension == :distance
         }
       end
+      next if point.nil?
 
       previous = point
     }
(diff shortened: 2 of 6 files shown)


optimizer_wrapper.rb
```diff
@@ -585,6 +585,20 @@ def self.build_csv(solutions)
   private
 
   def self.check_result_consistency(expected_value, results)
+    [results].flatten(1).each{ |result|
```
senhalil, line 588: `[results].flatten(1)` why do we need this `[].flatten` logic?

optimizer_wrapper.rb
```diff
@@ -585,6 +585,20 @@ def self.build_csv(solutions)
...
+      duration_map = result[:routes].map{ |route|
+        route[:activities].map{ |act| act[:departure_time] && (act[:departure_time] - act[:begin_time]) }.compact
+      }
+
+      next if result[:total_time] == result[:total_travel_time] + waiting_time_map.sum + duration_map.flatten.sum
+
+      log_string = "Expected #{result[:total_time]} == #{result[:total_travel_time]} +"\
+                   " #{waiting_time_map.sum} + #{duration_map.flatten.sum}"
+      log log_string, level: :warn
+      raise RuntimeError, 'Invalid computed times' if ENV['APP_ENV'] != 'production'
```
senhalil, line 599: Computed times are invalid

optimizer_wrapper.rb
```diff
@@ -585,6 +585,20 @@ def self.build_csv(solutions)
   private
 
   def self.check_result_consistency(expected_value, results)
+    [results].flatten(1).each{ |result|
+      waiting_time_map = result[:routes].map{ |route| route[:total_waiting_time] }.compact
+      duration_map = result[:routes].map{ |route|
```
senhalil, line 590: `x_map` has a specific meaning -- for example a mapping between variable ids and variables. here we have a simple array, I think we can just make do with pluralisation `waiting_times`, `durations`

optimizer_wrapper.rb
```diff
@@ -585,6 +585,20 @@ def self.build_csv(solutions)
...
+        route[:activities].map{ |act| act[:departure_time] && (act[:departure_time] - act[:begin_time]) }.compact
+      }
+
+      next if result[:total_time] == result[:total_travel_time] + waiting_time_map.sum + duration_map.flatten.sum
```
senhalil, line 594: Can you add another "non-production" before this `next` so that we don't have any negative weights in any routes?

Something similar to
```ruby
if result[:routes][:activities].any?{ |a| a[waiting_time].to_i < 0 }
  log 'Computed waiting times are invalid', level: :warn 
  # I wish the level warn would raise automatically on non-production environments 
  # so that we wouldn't need the next line
  raise 'Computed waiting times are invalid' if ENV['APP_ENV'] != 'production'
end
```

optimizer_wrapper.rb
```diff
@@ -648,6 +662,7 @@ def self.compute_route_total_dimensions(vrp, route, matrix)
           activity[:current_distance] ||= total[dimension].round if dimension == :distance
         }
       end
+      next if point.nil?
```
senhalil, line 665: Do we need this? It looks like to me that to skip one `previous = nil` assignment, we will do hundereds of extra `point.nil?` checks which is not necessary in the first place because `previous && point` handles the `nil` case correctly

optimizer_wrapper.rb
```diff
@@ -679,31 +694,38 @@ def self.compute_result_total_dimensions_and_round_route_stats(result)
...
-          route[:activities][seen][:detail][:setup_duration] || 0
-        else
-          0
+    loc_index = nil
+    consumed_travel_time = 0
+    route[:activities][1..-1].each.with_index{ |act, index|
+      used_travel_time = 0
+      if act[:type] == 'rest'
+        if loc_index.nil?
+          loc_index = 1 + index + route[:activities][1 + index..-1].index{ |a| a[:type] != 'rest' }
```
senhalil, line 708: Here, are we trying to handle the case where there are multiple pauses one after the other?

optimizer_wrapper.rb
```diff
@@ -679,31 +694,38 @@ def self.compute_result_total_dimensions_and_round_route_stats(result)
...
 
   def self.compute_route_waiting_times(route)
-    seen = 1
-    previous_end =
-      if route[:activities].first[:type] == 'depot'
-        route[:activities].first[:begin_time]
-      else
-        route[:activities].first[:end_time]
-      end
+    setup_types = %w[service pickup delivery]
```
senhalil, line 697: At the end of the day what we want to do is calculating the waiting_times given the arrival, departure, travel times of activities. So I read the function with that in mind but it is not easy to comprehend. Is there a way to simplify it?

senhalil (changes requested):

senhalil:
I saw too late that you put the PR in draft mode while I was reviewing :|

optimizer_wrapper.rb
braktar: We do need this line, otherwise, the previous_point is updated with a nil value even if it is a break

optimizer_wrapper.rb
braktar: Sure, the draft was created in a early stage. I was searching a way to launch the tests suite with optimizer-ortools dev branch

optimizer_wrapper.rb
senhalil: You say bug, I say feature :p

optimizer_wrapper.rb
senhalil: Well having a feedback at the early stage is always good :)

lib/interpreters/split_clustering.rb
```diff
@@ -481,6 +481,9 @@ def self.remove_poorly_populated_routes(vrp, result, limit)
 
         if load_flag && time_flag
           emptied_routes = true
+          %i[total_distance total_time total_travel_time total_value].each{ |key|
+            result[key] -= route[key] if route[key]
+          }
```
senhalil, line 486: Good catch :+1:  

This remainds me the internal mode::result class will need it own add/update/remove/empty functions to handle the logic inside the result class so that split_clustering (for example) does not have to have a result related update logic. It looks like it is going to be you whose going to implement the result model, can you make a note of this comment. You can just copy/paste the comment link if there is a ticket.

senhalil (commented):
This looks good to me. I saw that there are 3 failing real_cases tests. If the dev of this PR is correct that means we still have a bug somewhere which touches the time calculations. But we have no failing unit/functional tests. If this is the case can you please add a unit test based on an MWE?

braktar:
> This looks good to me. I saw that there are 3 failing real_cases tests. If the dev of this PR is correct that means we still have a bug somewhere which touches the time calculations. But we have no failing unit/functional tests. If this is the case can you please add a unit test based on an MWE?

The current failing are requiring this PR : https://github.com/Mapotempo/optimizer-ortools/pull/21

optimizer_wrapper.rb
```diff
@@ -501,6 +504,39 @@ def self.job_remove(api_key, id)
 
   def self.check_result_consistency(expected_value, results)
     [results].flatten(1).each{ |result|
+      if result[:routes].any?{ |route| route[:activities].any?{ |a| a[:waiting_time].to_i < 0 } }
+        log 'Computed waiting times are invalid', level: :warn
+        raise RuntimeError, 'Computed waiting times are invalid' if ENV['APP_ENV'] != 'production'
```
fonsecadeline, line 509: Do we want to raise an error ? This might return no result because of a wrong waiting time. Shouldn't we rather raise an error to debug internally instead of blocking optimization ? 

I can understand that we do not want to return a wrong value, but at the same time we would have no bypass for user to get its result. 

In order to minimize error in return value we could eventually replace negative values by zero and keep computing ? (while sending a warning to sentry)

optimizer_wrapper.rb
```diff
@@ -501,6 +504,39 @@ def self.job_remove(api_key, id)
...
+      if total_time != (total_travel_time || 0) +
+                       waiting_times.sum +
+                       (setup_durations.flatten.reduce(&:+) || 0) +
+                       (durations.flatten.reduce(&:+) || 0)
+
+        log_string = "Expected #{total_time} == #{total_travel_time} +"\
+                     " #{waiting_times.sum} + #{setup_durations.flatten.reduce(&:+)}"\
+                     " + #{durations.flatten.reduce(&:+)}"
+        log log_string, level: :warn
+        raise RuntimeError, 'Computed times are invalid' if ENV['APP_ENV'] != 'production'
```
fonsecadeline, line 537: Same here, could we use warn instead of raise ? We could return none of these errored fields but still return a result to customer ?

optimizer_wrapper.rb
```diff
@@ -585,31 +622,36 @@ def self.compute_result_total_dimensions_and_round_route_stats(result)
   end
 
   def self.compute_route_waiting_times(route)
-    seen = 1
-    previous_end =
-      if route[:activities].first[:type] == 'depot'
-        route[:activities].first[:begin_time]
+    previous_end = route[:activities].first[:begin_time]
```
fonsecadeline, line 625: Why is the check of first_route being a depot not needed anymore ? I feel route[:activities].first[:begin_time] is only valid if first activity is a depot. If it is more homogeneous now, shouldn't we now use route[:activities].first[:end_time] ?

optimizer_wrapper.rb
braktar: This raise won't happen in production environment. It will only raise during development, the behavior is similar to the raise performed for `total_time`

optimizer_wrapper.rb
braktar: A route always have a start_time if we have the time matrix.
Regardless, the type of the first activity, this first computation represent this start_time.

Here, it cannot be end_time to ensure the correct computation of the consumed_setup_time variable for the first activity, especially if the first activity is a rest and the next service has a setup_duration.

optimizer_wrapper.rb
braktar: We do log a warn, when in production environment, and raise the error while in test or in development environments ! :)

optimizer_wrapper.rb
fonsecadeline: Maybe there is something I do not understand. It seems to me that in the case first route activity is not a depot then we do not consider this first activity duration in l.651 (arrival_time computation).

braktar: If we have as first activity a rest and then a service with a setup_duration.

In some situation setup_duration may be applied before the rest. Then if i do not compute correctly this value from the beginning, we may have a shift in the time computed. Or we will have a wrongly computed waiting_time or we will consider twice the setup_duration

fonsecadeline: I agree with the fact that you recompute the value. I just do not see the case where first activity is a visit. (And I am wondering if part of your previous answer did not disappear ?)

fonsecadeline (approved):

fonsecadeline (changes requested):

fonsecadeline (approved):

senhalil (approved):

senhalil:
- [ ] @braktar I re-run all the tests because some were failing but now all are failing. I suspect it needs a manual rebase and coherence check because there is no conflict...

braktar:
It seems that the tests were all failing due to a CI issue. It was expecting an image which didn't exists. It seems to be solved after the rebase

PR #238, opened 2021-06-16
Forbids net connection during tests

braktar:


braktar:
It prevents any test from calling an external service on the internet

senhalil:
> It prevents any test from calling an external service on the internet

:+1:  Oh, I thought it works only on already mocked or stubbed requests.

senhalil:
Which also means that if we ever want to use a test to quickly launch a new instance or TDD we need to have `WebMock.allow_net_connect!` in the test.

test/test_helper.rb
```diff
@@ -120,7 +123,9 @@ def self.create(problem)
   def self.matrix_required(vrp)
     return unless ENV['TEST_DUMP_VRP'].to_s == 'true' && vrp.name
 
+    WebMock.enable_net_connect!
     vrp.compute_matrix
+    WebMock.disable_net_connect!
```
braktar, line 128: This should allow us to define new tests and avoid locals `WebMock.enable_net_connect!`

fonsecadeline (approved):

test/test_helper.rb
senhalil: You tried to dump a tests and it works right? Do you know by any chance what `allow_net_connection` does by the way?

senhalil (approved):

braktar:
`allow_net_connect` and `disable_net_connect` are aliases : https://github.com/bblimke/webmock/pull/789/files

PR #239, opened 2021-06-17
Model solution

braktar:
- [x] Requires https://github.com/Mapotempo/optimizer-api/pull/254.

This PR is divided into multiple part to ease the review : 
- [x] https://github.com/Mapotempo/optimizer-api/pull/308
- [ ] https://github.com/Mapotempo/optimizer-api/pull/309
- [ ] https://github.com/Mapotempo/optimizer-api/pull/310
- [ ] https://github.com/Mapotempo/optimizer-api/pull/311


models/service.rb
```diff
@@ -58,5 +58,14 @@ class Service < Base
     has_many :sticky_vehicles, class_name: 'Models::Vehicle'
     has_many :quantities, class_name: 'Models::Quantity'
     has_many :relations, class_name: 'Models::Relation'
+
+    def route_activity(index)
+      Models::RouteActivity.new(
+        service_id: self.original_id || self.id,
```
fonsecadeline, line 64: maybe this attribute could be only "id". Since you defined rests as an activity I think it is better to have the most neutral field.

models/solution/activity.rb
```diff
@@ -0,0 +1,36 @@
...
+#
+require './models/base'
+
+module Models
+  class RouteActivity < Base
+    # field :point_id
+    field :type
+    field :alternative
+    field :reason, default: nil
+    # TODO: The following fields should be merged in v2
```
fonsecadeline, line 26: What do you think about starting to return _id_ field already ?

braktar: As we are not in v2, we still need to use these multiple ids. But yes, at the end we definitely need to use a single id field

fonsecadeline: I meant, can't we return both _id_ and _service_id_/_pickup_id_/_shipment_id_ fields at the same time ? That would help anticipating transition to v2, because new customers could use _id_ field. We will have a transition period anyways.

lib/heuristics/periodic_heuristic.rb
```diff
@@ -1113,53 +1101,34 @@ def matrix(route_data, start, arrival, dimension = :time)
...
         number_in_sequence =
           @services_assignment[data[:id]][:assigned_indices][@services_assignment[data[:id]][:days].find_index(day)]
+        Models::Solution::Step.new(
+          service,
+          service_id: "#{data[:id]}_#{number_in_sequence}_#{service.visits_number}",
+          visit_index: number_in_sequence,
+          info: times,
+          index: data[:activity],
+          loads: loads
+        )
```
braktar, line 1130:
```suggestion
        Models::Solution::Step.new(
          service,
          service_id: "#{data[:id]}_#{number_in_sequence}_#{service.visits_number}",
          visit_index: number_in_sequence,
          info: times,
          index: data[:activity],
          loads: loads
        )kkkkk
```

braktar:
Replaced by https://github.com/Mapotempo/optimizer-api/pull/321

PR #240, opened 2021-06-18
Remove shipment model

braktar:
- [x] It should be rebased on https://github.com/Mapotempo/optimizer-api/pull/229 when merged
- [x] It has a commit from https://github.com/Mapotempo/optimizer-api/pull/241 so needs that to be merged


test/wrapper_test.rb
```diff
@@ -3040,8 +3046,28 @@ def test_assert_inapplicable_relations
...
+    problem[:vehicles].first[:capacities] = [{ unit_id: 'kg', limit: 2 }]
+    problem[:services][0][:quantities] = [{ unit_id: 'kg', value: 1 }]
+    problem[:services][1][:quantities] = [{ unit_id: 'kg', value: -1 }]
+    vrp = TestHelper.create(problem)
+    refute_includes OptimizerWrapper.config[:services][:vroom].inapplicable_solve?(vrp),
+                    :assert_no_relations_except_simple_shipments
+
+    problem[:services][1][:quantities] = [{ unit_id: 'kg', value: 2 }]
+    vrp = TestHelper.create(problem)
+    assert_includes OptimizerWrapper.config[:services][:vroom].inapplicable_solve?(vrp),
```
fonsecadeline, line 3068: Shouldn't this relation be rejected by vrp.create ?

test/wrappers/ortools_test.rb
```diff
@@ -2068,344 +2068,6 @@ def test_setup_duration
     assert_equal problem[:services].size, result[:routes][1][:activities].size - 1
   end
 
-  def test_pickup_delivery
```
fonsecadeline, line 2071: Why removing those tests ? We need to ensure this still works after our clean. These tests can be cleaned thought.

wrappers/ortools.rb
```diff
@@ -50,7 +50,6 @@ def solver_constraints
         :assert_only_force_centroids_if_kmeans_method,
         :assert_no_scheduling_if_evaluation,
         :assert_route_if_evaluation,
-        :assert_no_shipments_if_evaluation,
```
fonsecadeline, line 53: This should not be removed but assert should be edited with check on vrp.relations

wrappers/ortools.rb
```diff
@@ -594,44 +514,26 @@ def parse_output(vrp, _services, points, _matrix_indices, _cost, _iterations, ou
                 end
               elsif activity.type == 'service'
                 collected_indices << current_index
-                if current_index < vrp.services.size
                   service = vrp.services[current_index]
```
fonsecadeline, line 517: Isn't there an indentation problem ?

wrappers/wrapper.rb
```diff
@@ -23,9 +23,7 @@ def initialize(hash = {})
     end
 
     def solver_constraints
-      [
-       :assert_no_pickup_timewindows_after_delivery_timewindows,
```
fonsecadeline, line 27: I think this should still be checked.

wrappers/wrapper.rb
```diff
@@ -64,19 +62,6 @@ def assert_vehicles_no_alternative_skills(vrp)
       vrp.vehicles.none?{ |vehicle| vehicle.skills.size > 1 }
     end
 
-    def assert_no_direct_shipments(vrp)
```
fonsecadeline, line 67: I think this should still be checked.

wrappers/wrapper.rb
```diff
@@ -64,19 +62,6 @@ def assert_vehicles_no_alternative_skills(vrp)
...
-      vrp.shipments.none?{ |shipment| shipment.direct }
-    end
-
-    def assert_no_pickup_timewindows_after_delivery_timewindows(vrp)
```
fonsecadeline, line 71: I think this should still be checked.

wrappers/wrapper.rb
```diff
@@ -169,13 +144,22 @@ def assert_correctness_matrices_vehicles_and_points_definition(vrp)
...
+      vrp.relations.all?{ |r|
+        return true if r.linked_ids.empty? && r.linked_vehicle_ids.empty?
+        return false unless r.type == :shipment && r.linked_ids.size == 2
+
+        quantities = Hash.new {}
+        vrp.units.each{ |unit| quantities[unit.id] = [] }
+        r.linked_services.first.quantities.each{ |q| quantities[q.unit.id] << q.value }
+        r.linked_services.last.quantities.each{ |q| quantities[q.unit.id] << q.value }
+        quantities.all?{ |_unit, values|
+          [0, 2].include?(values.size) && values.first >= 0 && values.first == -values.last
```
fonsecadeline, line 160: This check should be done in vrp.create

braktar: OR-Tools can handle a shipment with a pickup of -1 and a delivery of -2. There no reason to refuse a problem with such quantities

wrappers/wrapper.rb
braktar: A direct shipment is converted into a sequence. Vroom is only candidate if the relation is a shipment. Then this check is no more needed

wrappers/wrapper.rb
braktar: As the consistency of timewindows within relations is checked within validate_data in a more generic way : `check_relation_consistent_timewindows` we are sure that this situation cannot happen here

wrappers/ortools.rb
braktar: I see no reason to keep the assert in the first place. So I didn't port it to the more generic way

test/wrappers/ortools_test.rb
braktar: Theses tests were verifying an unreachable part of the code

test/wrapper_test.rb
braktar: OR-Tools can handle a shipment with a pickup of -1 and a delivery of -2. There no reason to refuse a problem with such quantities

models/vrp.rb
```diff
@@ -393,7 +468,7 @@ def self.generate_schedule_indices_from_date(hash)
 
       # remove unavailable dates and ranges :
       hash[:vehicles].each{ |v| deduce_unavailable_days(hash, v, start_index, end_index, :vehicle) }
-      [hash[:services].to_a + hash[:shipments].to_a].flatten.each{ |element|
+      hash[:services].each{ |element|
```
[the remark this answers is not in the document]
braktar: One of the first thing done when arriving in the create method is `hash[:services] ||= []`

Then adding a `&` is not required

models/vrp.rb
```diff
@@ -354,7 +429,7 @@ def self.deduce_unavailable_days(hash, element, start_index, end_index, type)
     end
 
     def self.detect_date_indices_inconsistency(hash)
-      missions_and_vehicles = hash[:services].to_a + hash[:shipments].to_a + hash[:vehicles].to_a
```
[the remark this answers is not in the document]
braktar: hash[:services] cannot be nil here, as the first thing we done when arriving in the create is to assign `hash[:services] ||= []`
vehicles are required by grape we cannot receive a vrp without a vehicle

models/vrp.rb
```diff
@@ -138,10 +138,88 @@ def self.create(hash, delete = true)
...
+      hash[:services] ||= []
+      hash[:relations] ||= []
+      service_ids = hash[:services].map{ |service| service[:id] }
+      hash[:shipments]&.each{ |shipment|
+        linked_ids = []
+        %i[pickup delivery].each{ |part|
+          service = Oj.load(Oj.dump(shipment))
+          service[:original_id] = shipment[:id]
+          service[:id] += "_#{part}"
+          service[:id] += '_conv' while service_ids.any?{ |id| id == service[:id] } # protect against id clash
```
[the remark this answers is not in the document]
braktar: It comes with the commit of @senhalil 

I assume it stands for 'converted'

lib/interpreters/split_clustering.rb
```diff
@@ -146,8 +146,7 @@ def self.split_solve_candidate?(service_vrp)
           vrp.preprocessing_max_split_size &&
           vrp.vehicles.size > 1 &&
           (vrp.resolution_vehicle_limit.nil? || vrp.resolution_vehicle_limit > 1) &&
-          (vrp.services.size - empties_or_fills.size) > vrp.preprocessing_max_split_size &&
-          vrp.shipments.empty? # Clustering supports Shipment only as Relation  TODO: delete this check when Model::Shipment is removed
+          (vrp.services.size - empties_or_fills.size) > vrp.preprocessing_max_split_size
```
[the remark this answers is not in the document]
braktar: Services of the same relation can be correctly clustered together thanks to recent edits from @senhalil (https://github.com/Mapotempo/optimizer-api/pull/145)

This condition was a relicate which has no more reason to exist

lib/interpreters/periodic_visits.rb
```diff
@@ -92,30 +90,15 @@ def generate_timewindows(timewindows_set)
...
-            additional_tag = if mission_id == "#{related_missions[index].id}pickup"
-              'pickup'
-            elsif mission_id == "#{related_missions[index].id}delivery"
-              'delivery'
-            else
-              ''
-            end
-            "#{related_missions[index].id}_#{relation_index}_#{first_service.visits_number || 1}#{additional_tag}"
+          new_relation.linked_ids = related_missions.collect{ |mission|
+            "#{mission.id}_#{relation_index}_#{mission.visits_number}"
```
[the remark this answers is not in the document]
braktar: The logic of the method remains exactly the same. We first check if all services of the relation have the same number of visits. And then we duplicate and update the linked_ids. minus, the additional tag from the shipment.

An eventual rework of this method is unrelated to the removal of the shipments.

lib/interpreters/periodic_visits.rb
```diff
@@ -92,30 +90,15 @@ def generate_timewindows(timewindows_set)
...
-        related_missions = relation_linked_ids.collect{ |mission_id|
-          vrp.services.find{ |service| service.id == mission_id } ||
-            vrp.shipments.find{ |shipment|  ["#{shipment.id}pickup", "#{shipment.id}delivery"].include?(mission_id) }
-        }
-        next unless first_service && first_service.visits_number &&
-                    related_missions.all?{ |mission| mission.visits_number == first_service.visits_number }
+        related_missions = relation.linked_services
 
-        (1..(first_service.visits_number || 1)).collect{ |relation_index|
+        visits_number = related_missions.first&.visits_number
```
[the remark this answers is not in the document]
braktar: The purpose of this PR is to remove the Shipment model. It is far beyond the scope to expect change of behaviors

lib/interpreters/compute_several_solutions.rb
```diff
@@ -264,9 +263,9 @@ def self.select_best_heuristic(vrp)
       elsif vehicles.any?{ |vehicle| vehicle[:force_start] || vehicle[:shift_preference] && vehicle[:shift_preference] == 'force_start' }
         verified('path_cheapest_arc')
       elsif loop_route && unique_configuration &&
-            (vehicles.any?{ |vehicle| vehicle[:duration] } && vehicles.size == 1 || size_mtws.to_f / (services.collect{ |service| service[:visits_number] }.sum + shipments.size * 2) > 0.2 && size_rest.zero?)
+            (vehicles.any?{ |vehicle| vehicle[:duration] } && vehicles.size == 1 || size_mtws.to_f / (services.collect{ |service| service[:visits_number] }.sum) > 0.2 && size_rest.zero?)
         verified('global_cheapest_arc')
-      elsif vehicles.size == 1 && size_rest.positive? || !shipments.empty? || size_mtws > 0
+      elsif vehicles.size == 1 && size_rest.positive? || size_mtws > 0
```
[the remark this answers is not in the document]
braktar: I assume that the assignment of heuristic is quite obsolete and should be updated any way. I am not sure it is a good thing to keep old, cryptic and unverified conditions.

lib/heuristics/dichotomious_approach.rb
```diff
@@ -31,7 +31,6 @@ def self.dichotomious_candidate?(service_vrp)
           (service_vrp[:vrp].resolution_vehicle_limit.nil? || service_vrp[:vrp].resolution_vehicle_limit > service_vrp[:vrp].resolution_dicho_algorithm_vehicle_limit) &&
           service_vrp[:vrp].services.size - service_vrp[:vrp].routes.map{ |r| r[:mission_ids].size }.sum > service_vrp[:vrp].resolution_dicho_algorithm_service_limit &&
           !service_vrp[:vrp].scheduling? &&
-          service_vrp[:vrp].shipments.empty? && # TODO: check dicho with a P&D instance to remove this condition
```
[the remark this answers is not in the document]
braktar: Relations are now compatible with clustering approaches

models/vrp.rb
```diff
@@ -138,10 +138,88 @@ def self.create(hash, delete = true)
...
+          relation[:linked_ids].each_with_index{ |id, id_i|
+            next unless id == shipment_id
+
+            relation[:linked_ids][id_i] = pickup_service_id # which will be in same_route as delivery
+          }
+        when :sequence, :order
+          if relation[:linked_ids].any?{ |id| id == shipment_id }
+            msg = 'Relation between shipment pickup and delivery should be explicitly specified for `:sequence` and `:order` relations.'
+            raise OptimizerWrapper::DiscordantProblemError.new(msg)
+          end
```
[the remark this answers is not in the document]
braktar: @senhalil implemented it in the first place, he will probably be more precise than me on the expected behalf. But, day_lapse or meetup are inter-route relations. While shipment, order, etc are related to intra-route relations. Both cannot be applied on the same services. Shall we maybe forbid these combinations ?

wrappers/wrapper.rb
braktar: This assert has been moved to validate_data.

models/vrp.rb
```diff
@@ -138,10 +138,88 @@ def self.create(hash, delete = true)
...
+    def self.convert_shipment_within_routes(hash)
+      return unless hash[:shipments]
+      shipment_ids = Hash.new
+      hash[:shipments].each{ |shipment| shipment_ids[shipment[:id]] = 0 }
+      hash[:routes]&.each{ |route|
+        route[:mission_ids].map!{ |mission_id|
+          if shipment_ids.key?(mission_id)
+            if shipment_ids[mission_id].zero?
+              shipment_ids[mission_id] += 1
+              "#{mission_id}_pickup"
```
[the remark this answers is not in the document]
braktar: We add '_conv' to the shipment id until we find no match in the service ids

models/vrp.rb
```diff
@@ -152,23 +230,23 @@ def self.convert_position_relations(hash)
         case r[:type]
         when :force_first
           r[:linked_ids].each{ |id|
-            to_modify = [hash[:services], hash[:shipments]].flatten.find{ |s| s[:id] == id }
+            to_modify = [hash[:services]].flatten.find{ |s| s[:id] == id }
```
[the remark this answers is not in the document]
braktar: It is not better without your suggestion. I include this in my edits. Nevertheless, I wont push them until #229 and #241 are integrated

models/concerns/validate_data.rb
```diff
@@ -39,9 +39,9 @@ def check_consistency(hash)
     check_relations(periodic_heuristic)
     # TODO : this should be replaced by schedule when max_split does not use visits_number > 1 without schedule anymore
     # indeed, no configuration implies no schedule and there should be no visits_number > 1 in this case
-    # check_services_and_shipments(schedule)
-    check_services_and_shipments(configuration, schedule)
-    check_shipments_specificities
+    # check_services(schedule)
+    check_services(configuration, schedule)
+    check_position_relation_specificities
```
fonsecadeline, line 44: This too should be called in check_relations

models/vrp.rb
```diff
@@ -140,10 +140,94 @@ def self.create(hash, delete = true)
...
+            next unless id == shipment_id
+
+            relation[:linked_ids][id_i] = pickup_service_id # which will be in same_route as delivery
+          }
+        when :sequence, :order
+          if relation[:linked_ids].any?{ |id| id == shipment_id }
+            msg = 'Relation between shipment pickup and delivery should be explicitly specified for `:sequence` and `:order` relations.'
+            raise OptimizerWrapper::DiscordantProblemError.new(msg)
+          end
+        else
```
fonsecadeline, line 195: Could we use elsif to avoid a new indentation level ?

models/vrp.rb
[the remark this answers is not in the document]
fonsecadeline: Nice :) It was confusing to me to see this as 'Resolved' while nothing had changed.

fonsecadeline (approved):

models/vrp.rb
```diff
@@ -618,5 +699,19 @@ def scheduling?
...
+
+    def legacy_shipments?
+      self.relations.any?{ |r|
+        next unless r.type == :shipment && r.linked_ids.size == 2
+
+        quantities = Hash.new {}
+        self.units.each{ |unit| quantities[unit.id] = [] }
+        r.linked_services.first.quantities.each{ |q| quantities[q.unit.id] << q.value }
+        r.linked_services.last.quantities.each{ |q| quantities[q.unit.id] << q.value }
+        quantities.all?{ |_unit, values|
```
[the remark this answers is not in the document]
braktar: I see no reason to refuse such combination, OR-Tools has no restriction on this point.

models/vrp.rb
braktar: The `else` comes from the `case` not from a `if`. And this generates a syntax error :/

models/vrp.rb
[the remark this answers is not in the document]
fonsecadeline: I think I had not understand something : legacy_shipments checks if anything was provided through vrp[:shipments], right ? 

What I had suggested was to check that any shipment was provided, either with vrp[:shipments] or with shipment relations. For me, best heuristic when vrp[:shipments].any? should be the same as if vrp[:relations].any?{ |r| r[:type] == :shipment }. Reworking of these conditions should be done in another PR.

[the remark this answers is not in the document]
braktar: I strictly stick to the previous behavior, as the intention in this PR is only to convert the shipment models into services + relations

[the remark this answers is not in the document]
fonsecadeline: Now we totally ignore the case of new possible version of shipments while it could be safer to "at least" do as we used to do with shipments before.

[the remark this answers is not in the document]
fonsecadeline: I understand your point of view. Maybe @senhalil can help us find the best way :)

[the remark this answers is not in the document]
fonsecadeline: As discussed with @braktar in fact we already used to ignore cases with shipment relation instead of vrp[:shipments] so this is okay for me now.

fonsecadeline (approved):

models/concerns/expand_data.rb
```diff
@@ -80,14 +56,14 @@ def expand_unavailable_days
...
+    self.services.each{ |mission|
       mission.unavailable_days |= unavailable_days
     }
   end
 
   def provide_original_info
-    (self.services + self.shipments + self.vehicles).each{ |element|
-      element.original_id = element.id
+    (self.services + self.vehicles).each{ |element|
+      element.original_id ||= element.id
```
fonsecadeline, line 66: Failing scheduling tests come from this edit. Why should we keep original_id ? This field is not provided by users (not in grape model).

braktar: When converting shipment into services we edit the id of the pickup and the delivery are edited, we assign the original id at this moment. Then, at this point, the original id may already be defined

lib/heuristics/dichotomious_approach.rb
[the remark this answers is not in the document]
senhalil: We haven't tested `dicho` for shipments and if I am not mistaken dicho uses a different (customised) split method. It is better if we keep this check (as a relation condition).

lib/heuristics/scheduling_heuristic.rb
```diff
@@ -397,7 +397,7 @@ def reorder_stops(vrp)
         all_routes.each{ |day, route_data|
           next if route_data[:stops].uniq{ |s| s[:point_id] }.size <= 1
 
-          corresponding_vehicle = vrp.vehicles.find{ |v| v.id == "#{vehicle_id}_#{day}" }
+          corresponding_vehicle = vrp.vehicles.find{ |v| v.id.match?(/#{vehicle_id}.*_#{day}/) }
```
senhalil, line 400: Can't see why we need such a modification in this PR :thinking: 

Expansion of shipments into services shouldn't require a modification of a logic dependent on vehicle_ids, no?

lib/interpreters/compute_several_solutions.rb
```diff
@@ -238,37 +238,47 @@ def self.find_best_heuristic(service_vrp)
...
         end
       }
-      size_mtws = services.select{ |service| service[:timewindows].to_a.size > 1 }.size
-      size_rest = vehicles.collect{ |vehicle| vehicle[:rests].to_a.size }.sum
-      unique_configuration = vehicles.collect{ |vehicle| vehicle[:router_mode] }.uniq.size == 1 && vehicles.collect{ |vehicle| vehicle[:router_dimension] }.uniq.size == 1 &&
-                             vehicles.collect{ |vehicle| vehicle[:start_point_id] }.uniq.size == 1 && vehicles.collect{ |vehicle| vehicle[:end_point_id] }.uniq.size == 1
-
-      if vehicles.any?{ |vehicle| vehicle[:overall_duration] }
+      size_mtws = services.count{ |service| service.activity.timewindows.size > 1 }
+      size_rest = vehicles.map{ |vehicle| vehicle.rests.size }.sum
```
senhalil, line 261: I am not 100% sure but the following should work
```suggestion
      size_rest = vehicles.sum{ |vehicle| vehicle.rests.size }
```

lib/interpreters/compute_several_solutions.rb
```diff
@@ -238,37 +238,47 @@ def self.find_best_heuristic(service_vrp)
...
+      unique_configuration = vehicles.map(&:router_mode).uniq.size == 1 &&
+                             vehicles.map(&:router_dimension).uniq.size == 1 &&
+                             vehicles.map{ |vehicle| vehicle.start_point.id }.uniq.size == 1 &&
+                             vehicles.map{ |vehicle| vehicle.end_point.id }.uniq.size == 1
```
senhalil, line 265: no need for collect `map().uniq.size` -> `uniq().size`

@braktar you replaced `vehicle[:start_point_id]` with `vehicle.start_point.id` but the latter would fail if the vehicle have no start point, right? I suggest just replacing it with `vehicle.start_point_id` as follows:

```suggestion
      unique_configuration = vehicles.uniq(&:router_mode).size == 1 &&
                             vehicles.uniq(&:router_dimension).size == 1 &&
                             vehicles.uniq(&:start_point_id).size == 1 &&
                             vehicles.uniq(&:end_point_id).size == 1
```

models/concerns/validate_data.rb
```diff
@@ -298,10 +345,33 @@ def check_relations(periodic_heuristic)
     )
   end
 
+  def check_sticky_relation_consistency
+    unconsistent_stickies = []
+    @hash[:relations].none?{ |relation|
```
senhalil, line 350: If I am not mistaken in the VRP hash any key can be `nil`, right? Either we should initialise `hash[:relations] ||= []` as we do for vehicles and services at the beginning of `check_consistency` or we need to do as follows...
```suggestion
    @hash[:relations]&.none?{ |relation|
```

I prefer the first solution (i.e., initialisation), in that case it is better to clean the unnecessary `&.`, what do you think @braktar ?

models/vrp.rb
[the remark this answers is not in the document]
senhalil: Yeah I found it that way and only implemented the existing relations. You did well by adding the `raiswe`

optimizer_wrapper.rb
```diff
@@ -174,7 +174,7 @@ def self.define_process(service_vrp, job = nil, &block)
     vrp = service_vrp[:vrp]
     dicho_level = service_vrp[:dicho_level].to_i
     split_level = service_vrp[:split_level].to_i
-    log "--> define_process VRP (service: #{vrp.services.size}, shipment: #{vrp.shipments.size}, vehicle: #{vrp.vehicles.size}, v_limit: #{vrp.resolution_vehicle_limit}) with levels (dicho: #{dicho_level}, split: #{split_level})", level: :info
+    log "--> define_process VRP (service: #{vrp.services.size}, vehicle: #{vrp.vehicles.size}, v_limit: #{vrp.resolution_vehicle_limit}) with levels (dicho: #{dicho_level}, split: #{split_level})", level: :info
```
senhalil, line 177: Maybe we can keep the count of shipments in these logs, what do you think? It kinda is a valuable info when we need to check the logs (since shipment relations translate into different type of "service").

optimizer_wrapper.rb
```diff
@@ -197,7 +197,7 @@ def self.solve(service_vrp, job = nil, block = nil)
     vrp = service_vrp[:vrp]
     service = service_vrp[:service]
     dicho_level = service_vrp[:dicho_level]
-    log "--> optim_wrap::solve VRP (service: #{vrp.services.size}, shipment: #{vrp.shipments.size} vehicle: #{vrp.vehicles.size} v_limit: #{vrp.resolution_vehicle_limit}) with levels (dicho: #{service_vrp[:dicho_level]}, split: #{service_vrp[:split_level].to_i})", level: :debug
+    log "--> optim_wrap::solve VRP (service: #{vrp.services.size}, vehicle: #{vrp.vehicles.size} v_limit: #{vrp.resolution_vehicle_limit}) with levels (dicho: #{service_vrp[:dicho_level]}, split: #{service_vrp[:split_level].to_i})", level: :debug
```
senhalil, line 200: this too

wrappers/wrapper.rb
senhalil: I was thinking of. If the user uses the shorcut `api:shipment` and send us a `direct` shipment we will refuse the vroom solver with `sequence` relation explanation :/ Not ideal but at the same time not so important either.

senhalil (commented):
I think it is almost good to go, I have some remarks before we merge it.

lib/heuristics/scheduling_heuristic.rb
fonsecadeline: Indeed, this should not be needded now that we removed original_id in test_helper -> coerce

lib/interpreters/compute_several_solutions.rb
fonsecadeline: I think we should rather use `vehicle.start_point&.id` because every field with form `vehicle[:start_point_id]` corresponds to raw data provided by user. If `vrp.create` modifies this field in any way we will miss it. We should avoid using this form.

optimizer_wrapper.rb
fonsecadeline: This might output conflicting values I think : vrp.services will include vrp[:shipments]. Vrp.shipments does not exist anymore.

senhalil: Do you find the following confusing? 

`services: 10, shipment relations: 5` which means there are 5 shipments and 0 normal services
`services: 10, shipment relations: 1` which means there is only one P&D pair and remaining 8 services are normal

Now that it is possible to send P&D from mt-web, it is useful to have shipment count in the logs for quick debugging. I used `service: x, shipment: y` logs before and I don't think it can output conflicting values.

fonsecadeline: Oh, right, if we write shipment **relations** then it is clear :) friday afternoon...

lib/heuristics/dichotomious_approach.rb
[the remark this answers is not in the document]
braktar: Shouldn't we check at least order and sequence relations as well ?

lib/interpreters/compute_several_solutions.rb
braktar: I agree with @fonsecadeline `start_point_id` and `end_point_id` did not belongs to the model, they are co-products of our model creation from JSON or CSV. Any internal vehicle creation may not guarantee the presence of these fields.

lib/interpreters/compute_several_solutions.rb
```diff
@@ -238,37 +238,47 @@ def self.find_best_heuristic(service_vrp)
...
     def self.select_best_heuristic(vrp)
-      vehicles = vrp[:vehicles] || []
-      services = vrp[:services] || []
-      shipments = vrp[:shipments] || []
+      vehicles = vrp.vehicles
+      services = vrp.services
 
       loop_route = vehicles.any?{ |vehicle|
-        if (vehicle.start_point_id.nil? || vehicle.end_point_id.nil?) && (vehicle.start_point.nil? || vehicle.end_point.nil?)
+        if (vehicle.start_point_id.nil? || vehicle.end_point_id.nil?) &&
```
fonsecadeline, line 245: Regarding what we discussed this morning, I guess we can simply use one oh these lines ?

senhalil (approved):
:rocket:

PR #241, opened 2021-06-18
collect_items and Vroom can handle shipment relations

senhalil:
- collect_data_items respects the minimum vehicle limits
- linking_relations can be grouped in collect_items
- split_solve represents services w multiple points
- Vroom can handle simplified setup durations
- Vroom can handle shipments as relation
- max_split_size prevents only optim-sync and Vroom can be called for sub-rpoblems
- small quality and linter fixes

### DIFF
--- a/lib/interpreters/split_clustering.rb
+++ b/lib/interpreters/split_clustering.rb
@@ -233,7 +233,7 @@ module Interpreters
         ).sort_by!{ |side|
           [side.size, side.sum(&:visits_number)] # [number_of_vehicles, number_of_visits]
         }.reverse!.collect!{ |side|
-          enum_current_vehicles.select{ |v| side.any?{ |s| s.id == v.id } }
+          enum_current_vehicles.select{ |v| side.any?{ |s| s.id == "0_representative_vrp_s_#{v.id}" } }
         }
 
         unless sides.size == 2 && sides.none?(&:empty?)
@@ -242,6 +242,10 @@ module Interpreters
           log 'There should be exactly two clusters in split_solve_core!', level: :warn
           ss_data[:cannot_split_further] = true
         end
+
+        # TODO: In some case one side might be significantly smaller than the other side.
+        # This is not necessarily a bad thing but we might want to re-groupe such "tiny" VRPs
+        # to prevent inefficient use of vehicles.
 
         split_service_counts = sides.collect{ |current_vehicles|
           current_vehicles.sum{ |v| ss_data[:service_vehicle_assignments][v.id].size }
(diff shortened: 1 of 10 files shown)


wrappers/wrapper.rb
```diff
@@ -1111,6 +1113,139 @@ def simplify_vehicle_pause(vrp, result = nil, options = { mode: :simplify })
...
+          vehicle[:simplified_coef_setup] = nil
+          vehicle[:simplified_additional_setup] = nil
+        }
+      when :patch_result
+        # patches the result
+        # the travel_times need to be decreased and setup_duration need to be increased by
+        # (coef_setup * setup_duration + additional_setup) if setup_duration > 0 and travel_time > 0
+        return nil unless vrp.services.any?{ |s| s[:simplified_setup_duration] }
+
+        vehicles_grouped_by_vehicle_id = vrp.vehicles.group_by(&:id)
```
braktar, line 1209: can we group by id ? I assume it is supposed to be unique. original_id may not be unique

I you are expecting to have exactly one vehicle by group, I would be expecting a hash id=>vehicle.

braktar (approved):

lib/interpreters/split_clustering.rb
```diff
@@ -1162,26 +1279,27 @@ def remove_used_empties_and_refills(vrp, result)
...
-        returned = if node.nil?
-          []
         elsif (graph[node][:level]).zero?
           [node]
         else
           [tree_leafs(graph, graph[node][:left]), tree_leafs(graph, graph[node][:right])]
         end
+      end
+
+      def tree_leafs_delete(graph, node)
```
fonsecadeline, line 1289: _tree_leafs_delete_ and _tree_leafs_ seem very similar. Could we consider having only _tree_leafs_ function but with an option to delete ?

senhalil: I haven't modified `tree_leafs_delete` in purpose, it shouldn't be in my dev at all :confused: I am confused!

senhalil: Oh it is the syntax correction. I apparently corrected the `if` conditional assignment but the dev has nothing to do with `tree_leafs` functions. They are used in `split_hierarchical`. So I would rather not make such a change in this PR.

fonsecadeline: In fact I could not see the difference :p maybe some spaces where removed ?

senhalil: If you check with "Hide whitespace changes" option of the review page, then the diff becomes more obvious and minimal.

fonsecadeline (approved):

PR #242, opened 2021-06-21
Fix: Vroom cannot handle activity positions

senhalil:
Added an assert to refuse Vroom if positions
Added a test for the new assert

### DIFF
--- a/wrappers/wrapper.rb
+++ b/wrappers/wrapper.rb
@@ -274,6 +274,12 @@ module Wrappers
       (vrp.vehicles.map(&:shift_preference).uniq - [:minimize_span] - ['minimize_span']).empty? || vrp.preprocessing_first_solution_strategy.to_a.first != 'periodic'
     end
 
+    def assert_no_activity_with_position(vrp)
+      vrp.services.none?{ |service|
+        (service.activities.to_a + [service.activity]).compact.any?{ |a| a.position != :neutral }
+      }
+    end
+
     def assert_no_vehicle_overall_duration_if_heuristic(vrp)
       vrp.vehicles.none?(&:overall_duration) || vrp.preprocessing_first_solution_strategy.to_a.first != 'periodic'
     end
--- a/test/wrapper_test.rb
+++ b/test/wrapper_test.rb
@@ -2129,6 +2129,18 @@ class WrapperTest < Minitest::Test
     assert_includes OptimizerWrapper.config[:services][:vroom].inapplicable_solve?(vrp), :assert_correctness_matrices_vehicles_and_points_definition
   end
 
+  def test_activity_position_presence
+    problem = VRP.basic
+    vrp = TestHelper.create(problem)
+    refute_includes OptimizerWrapper.config[:services][:ortools].inapplicable_solve?(vrp), :assert_no_activity_with_position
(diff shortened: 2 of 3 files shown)


senhalil:
Failing `test_deleted_job` is not related to the dev, it randomly fails.

wrappers/wrapper.rb
```diff
@@ -275,6 +275,12 @@ def assert_wrong_vehicle_shift_preference_with_heuristic(vrp)
       (vrp.vehicles.map(&:shift_preference).uniq - [:minimize_span] - ['minimize_span']).empty? || vrp.preprocessing_first_solution_strategy.to_a.first != 'periodic'
     end
 
+    def assert_no_activity_with_position(vrp)
```
fab-girard, line 278: since this `assert` is only for one solver, I suggest you to add it in `vroom.rb` instead of `wrapper.rb`

senhalil: All of our asserts are implemented in one place `wrapper.rb` and used by the needing solvers.

If an assert is used by both solvers, then it becomes a vrp.create.check (except certain cases) because it is about invalidity or disagreement of parameters.

fab-girard: yes I already explained it's not a best practice: code needs to be defined as near as possible it is used (when it's possible). Somebody has changed this by the past. At the birth of the `vroom.rb` file, existing `assert_***` methods could be for some of them in this file (and were sometimes named `assert_vroom_***`: this was more common)

senhalil: Why don't we open a ticket and move all the asserts at the same time, this PR uses the existing code design and changing this right here doesn't sound a good idea -- one single assert in `vroom.rb` all others in `wrapper.rb` ?

Both vroom and ortools are based on the wrapper class and `wrapper.rb` implements the `inapplicable_solve?` logic which is the place the actual place where these asserts are called and having the asserts defined in here makes sense because asserts are general statement (for exeample `if the vrp has a service with position or not` is a generic question) and some solvers have certain limitations (in this case vroom has this limitation, so it adds this assert `assert_no_activity_with_position` to its list).

braktar: An assert describes a property of a vrp. Each solver say if he can manage it or not. It is not a property of VROOM. Nevertheless, as we want to try other solvers. Implementing an assert directly within a solver class will just increase the maintenance.

I don't remember specific asserts with a solver name in it. If so, it was a misinterpretation of the assert feature.

braktar (approved):

fonsecadeline (approved):

wrappers/wrapper.rb
fab-girard: open a ticket: we will rollback to the initial way to define assert in solvers class with following rule (as we started):
- if assert is used by only one solver, define it in this solver
- if assert is used by more than one solver, define it in wrapper base class
- if assert is used by all solvers: maybe, it should not be an assert but a Exception instead if vrp is not relevant (exemple: assert_no_pickup_timewindows_after_delivery_timewindows ?)

senhalil: Done https://gitlab.com/mapotempo/optimizer-api/-/issues/821

PR #244, opened 2021-06-25
vehicle can have no timewindow when calculating service exclusion cost

fonsecadeline:


fonsecadeline:
We can not use 2**56.

PR #245, opened 2021-06-25
[OO] Add maximum initial quantity to ortools

braktar:
Requirement
- [x] https://github.com/Mapotempo/optimizer-ortools/pull/22

### DIFF
--- a/wrappers/ortools.rb
+++ b/wrappers/ortools.rb
@@ -36,7 +36,6 @@ module Wrappers
       super + [
         :assert_end_optimization,
         :assert_vehicles_objective,
-        :assert_vehicles_no_capacity_initial,
         :assert_vehicles_no_alternative_skills,
         :assert_zones_only_size_one_alternative,
         :assert_only_empty_or_fill_quantities,
@@ -97,6 +96,31 @@ module Wrappers
           unit_id: unit.id,
           fill: false,
           empty: false
+        }
+      }
+      total_quantities = vrp.units.map{ |unit| [unit.id, 0] }.to_h
+
+      vrp.relations.select{ |r| r.type == :shipment }.each{ |r|
+        vrp.units.each{ |unit|
+          total_negative = 0
+          total_positive = 0
+          r.linked_services.each{ |s|
+            quantity = s.quantities.find{ |q| q.unit_id == unit.id }&.value || 0
+            if quantity > 0
+              total_positive += quantity
+            else
+              total_negative -= quantity
+            end
+          }
+          total_quantities[unit.id] += [total_positive, total_negative].max
+        }
+      }
+
+      vrp.services.each{ |service|
+        next if service.relations.any?{ |r| r.type == :shipment }
+
+        service.quantities.each{ |q|
+          total_quantities[q.unit.id] += (q.value || 0).abs
         }
       }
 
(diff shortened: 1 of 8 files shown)


wrappers/ortools.rb
```diff
@@ -100,6 +99,17 @@ def solve(vrp, job, thread_proc = nil, &block)
           empty: false
         }
       }
+      sum_quantities = vrp.units.map{ |unit| [unit.id, 0] }.to_h
+      vrp.services.each{ |service|
+        service.quantities.each{ |q|
+          sum_quantities[q.unit.id] += (q.value || 0).abs
+        }
+      }
+      vrp.shipments.each{ |shipment|
```
fonsecadeline, line 108: I guess we can remove this part now ?

wrappers/ortools.rb
```diff
@@ -366,7 +376,8 @@ def solve(vrp, job, thread_proc = nil, &block)
             OrtoolsVrp::Capacity.new(
               limit: (q&.limit && q.limit < 1e+22) ? q.limit : -1,
               overload_multiplier: q&.overload_multiplier || 0,
-              counting: unit&.counting || false
+              counting: unit&.counting || false,
+              initial: q.initial || [sum_quantities[unit.id], q.limit].compact.max
```
senhalil, line 380: I suspect this `max` should be `min`, right?


Can there be a reason to take the max -- i.e., why would we load the vehicle more than the total_demand if the capacity.limit > total_demand?

Similarly if the the vehicle capacity is less than total_demand, then initial load is set to total_demand > capacity.limit.

wrappers/ortools.rb
```diff
@@ -100,6 +99,17 @@ def solve(vrp, job, thread_proc = nil, &block)
           empty: false
         }
       }
+      sum_quantities = vrp.units.map{ |unit| [unit.id, 0] }.to_h
+      vrp.services.each{ |service|
+        service.quantities.each{ |q|
+          sum_quantities[q.unit.id] += (q.value || 0).abs
+        }
+      }
```
senhalil, line 107: This double counts the demand of the services in shipment relation. 

I think you can first do a loop on `vrp.relations.select{:shipment}.each.linked_services.last.quantities.each`
then in the `vrp.services.each` loop skip the `service.relations.any?{:shipment}`

senhalil: I think `total_quantities` or `quantity_totals` is easier to understand than `sum_quantities`.

wrappers/wrapper.rb
```diff
@@ -56,7 +56,7 @@ def assert_vehicles_start_or_end(vrp)
 
     def assert_vehicles_no_capacity_initial(vrp)
       vrp.vehicles.none?{ |vehicle|
-        vehicle.capacities.find{ |c| c.initial&.positive? }
+        vehicle.capacities.any?{ |c| c.initial && c.initial < c.limit }
```
senhalil, line 59: I don't quite get this condition.

senhalil: Does `initial_load` signify the existing load in the vehicle? If so, how come it can be greater than the vehicle capacity?

senhalil (changes requested):

wrappers/ortools.rb
braktar: if we have an overload_multiplier, we can allow the sum of the quantities loaded to exceed the capacity, the `max` is used at this purpose

wrappers/wrapper.rb
braktar: If the initial capacity is greater than the capacity, it supposes or that we have a overload multiplier, or the initial capacity is not restricted

senhalil: My question was due to the definition miss-match :) initial-load vs initial-capacity. it makes sense now

fonsecadeline (approved):

PR #248, opened 2021-06-29
Fix: relation validations are skipped because type is a symbol

senhalil:


models/concerns/validate_data.rb
```diff
@@ -229,11 +229,11 @@ def check_trip_timewindows_consistency(relation_vehicles)
   end
 
   def check_relations(periodic_heuristic)
-    return unless @hash[:relations].to_a.any?
+    return unless @hash[:relations]&.any?
 
     @hash[:relations].group_by{ |relation| relation[:type] }.each{ |type, relations|
-      case type
-      when 'vehicle_trips'
+      case type.to_sym
```
fonsecadeline, line 235: For your informaiton, this is already done in test_helper -> coerce . But it can be safer to keep those both.

fonsecadeline (approved):

braktar (approved):

models/concerns/validate_data.rb
senhalil: I still wanted to protect us since there is no type safety :/ Plus, sym.to_sym is a constant time O(1) operation and this way we know that inside the `case ... when` there can't ever be a silly `string is not a symbol` type of mistake in the future.

PR #251, opened 2021-07-06
Fix: dont simplify pause if v.cost_late_multiplier

senhalil:
Resolves api-optimizer-prod/issues/16755 and api-optimizer-beta/issues/16779
Closes optimizer-api/-/issues/828

### DIFF
--- a/wrappers/wrapper.rb
+++ b/wrappers/wrapper.rb
@@ -937,8 +937,11 @@ module Wrappers
           # TODO: (r.timewindows&.size.to_i > 1) having multiple TWs for pauses is possible but even harder.
           # If necessary, this implementation can be extended to handle this case by re-optimizing with or-tools
           # in a generic way
-          next if vehicle.rests.size > 1 ||
-                  vehicle.rests.any?{ |r| r.timewindows&.size.to_i > 1 || r.exclusion_cost.to_f.positive? }
+          next if (vehicle.timewindow&.end && vehicle.cost_late_multiplier) ||
+                  vehicle.rests.size > 1 ||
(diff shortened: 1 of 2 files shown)


wrappers/wrapper.rb
```diff
@@ -937,8 +937,11 @@ def simplify_vehicle_pause(vrp, result = nil, options = { mode: :simplify })
           # TODO: (r.timewindows&.size.to_i > 1) having multiple TWs for pauses is possible but even harder.
           # If necessary, this implementation can be extended to handle this case by re-optimizing with or-tools
           # in a generic way
-          next if vehicle.rests.size > 1 ||
-                  vehicle.rests.any?{ |r| r.timewindows&.size.to_i > 1 || r.exclusion_cost.to_f.positive? }
+          next if (vehicle.timewindow&.end && vehicle.cost_late_multiplier) ||
+                  vehicle.rests.size > 1 ||
+                  vehicle.rests.any?{ |r|
+                    r.timewindows&.size.to_i > 1 || r.late_multiplier.to_f.positive? || r.exclusion_cost.to_f.positive?
```
braktar, line 943: rest late_multiplier and exclusion_cost are currently not used. We should consider to remove them from both the input and the model.

braktar (approved):

wrappers/wrapper.rb
senhalil: Good idea! Can you create a ticket for removing these params, so that we will deprecate them properly? 

I had to add the checks since these fields exists on the model and API currently. They will be removed with the dev deprecating these fields.

fonsecadeline (approved):

PR #254, opened 2021-07-08
Improve base.rb & update active_hash

senhalil:
For `has_many` and `belongs_to`

Closes `optimizer-api/-/issues/446`


braktar (commented):
Regarding this test error : https://github.com/Mapotempo/optimizer-api/pull/254/checks?check_run_id=3021525858#step:5:1659
It seems like the getter for end_point_id may not work as expected

senhalil:
It turns out this is due to a weird behaviour in active_hash. It is possible to create two Point objects as `{id: 1}` and `{id: "1"}` but it is not possible to to get `{id: "1"}` because there is a `to_s` in the `insert` and `find` methods. 

This means if the user defines two points -- one with integer, the other with string -- there is no warning (which is cool) but then `start_point_id: "1"` and `start_point_id: 1` will both point to the Point with `id: 1`. This is just calling for trouble because `"1"` and `1` are not the same thing and `to_s` is not necessary because Hash can handle integer keys. I will open a PR in `active_hash` repo

senhalil:
- [ ] Needs https://github.com/Mapotempo/active_hash/pull/4 to be merged and gemfile corrected and gemfile.lock updated

braktar (approved):

braktar (approved):

fonsecadeline (approved):

lib/tsp_helper.rb
```diff
@@ -19,14 +19,17 @@
...
+    tsp_suffix = Digest::MD5.hexdigest(
+      (vrp.points.map(&:id) + vrp.services.map(&:id) + vrp.vehicles.map(&:id) + [Time.now.to_f, rand]).join(' ')
+    )
     vrp.points.each{ |pt|
       service = vrp.services.find{ |service_| service_.activity.point_id == pt.id }
       next if !service
 
       services << {
-        id: service[:id],
+        id: "#{tsp_suffix}_#{service[:id]}",
```
fab-girard, line 30: prefer `prefix` instead

senhalil: fixed

fab-girard (approved):

PR #256, opened 2021-07-19
Introduce same_cluster relation

fonsecadeline:
Requires : 
- [x] merge of and edit gemfile.lock accordingly  https://github.com/Mapotempo/balanced_vrp_clustering/pull/24

### DIFF
--- a/lib/interpreters/split_clustering.rb
+++ b/lib/interpreters/split_clustering.rb
@@ -33,6 +33,7 @@ module Interpreters
       same_route
       sequence
       shipment
+      same_vehicle
     ].freeze
     # Relations that force multiple services/vehicles to stay in the same VRP
     FORCING_RELATIONS = %i[
@@ -575,6 +576,10 @@ module Interpreters
       points_ids = sub_vrp.services.map{ |s| s.activity.point.id }.uniq.compact
       sub_vrp.rests = sub_vrp.rests.select{ |r| sub_vrp.vehicles.flat_map{ |v| v.rests.map(&:id) }.include? r.id }
       sub_vrp.relations = sub_vrp.relations.select{ |r| r.linked_ids.all? { |id| sub_vrp.services.any? { |s| s.id == id } } }
+      if entity == :vehicle
+        sub_vrp.relations.delete_if{ |r| r.type == :same_vehicle }
+        sub_vrp.services.each{ |s| s.relations.delete_if{ |r| r.type == :same_vehicle } }
+      end
       sub_vrp.points = sub_vrp.points.select{ |p| points_ids.include? p.id }.compact
       sub_vrp.points += sub_vrp.vehicles.flat_map{ |vehicle| [vehicle.start_point, vehicle.end_point] }.compact.uniq
       sub_vrp = add_corresponding_entity_skills(entity, sub_vrp)
--- a/models/concerns/expand_data.rb
+++ b/models/concerns/expand_data.rb
@@ -40,14 +40,6 @@ module ExpandData
     }
   end
 
-  def clean_according_to(unfeasible_services)
-    unfeasible_services.each{ |unfeasible_service|
-      self.routes.each{ |route|
-        route.mission_ids.delete_if{ |id| id == unfeasible_service[:original_service_id] }
-      }
-    }
-  end
-
   def expand_unavailable_days
     unavailable_days = self.schedule_unavailable_days.select{ |unavailable_index|
       unavailable_index >= self.schedule_range_indices[:start] && unavailable_index <= self.schedule_range_indices[:end]
--- a/models/vrp.rb
+++ b/models/vrp.rb
@@ -187,7 +187,7 @@ module Models
           relation[:linked_ids][0] = delivery_service_id if relation[:linked_ids][0] == shipment_id
           relation[:linked_ids][1] = pickup_service_id if relation[:linked_ids][1] == shipment_id
           relation[:lapse] ||= 0
-        when :same_route
+        when :same_route, :same_vehicle
           relation[:linked_ids].each_with_index{ |id, id_i|
             next unless id == shipment_id
 
@@ -392,7 +392,7 @@ module Models
       }
       hash[:relations] = uniq_relations
 
-      return if hash[:configuration][:preprocessing].to_h[:partitions].to_a.any?{ |p| p[:entity] == :vehicle } ||
+      return if hash[:configuration].to_h[:preprocessing].to_h[:partitions].to_a.any?{ |p| p[:entity] == :vehicle } ||
                 hash[:relations].none?{ |r| r[:type] == :same_vehicle }
 
       hash[:relations].each{ |relation|
--- a/models/vrp.rb
+++ b/models/vrp.rb
@@ -375,7 +375,7 @@ module Models
     end
 
     def self.remove_unnecessary_relations(hash)
-      return hash unless hash[:relations]&.any?
+      return unless hash[:relations]&.any?
 
       types_with_duration =
         %i[minimum_day_lapse maximum_day_lapse
@@ -391,6 +391,16 @@ module Models
         uniq_relations += relations_set.uniq
       }
       hash[:relations] = uniq_relations
+
+      return if hash[:configuration][:preprocessing].to_h[:partitions].to_a.any?{ |p| p[:entity] == :vehicle } ||
+                hash[:relations].none?{ |r| r[:type] == :same_vehicle }
+
+      hash[:relations].each{ |relation|
+        next unless relation[:type] == :same_vehicle
+
+        # when no partition with vehicle entity is provided then :same_vehicle corresponds to same_route
+        relation[:type] = :same_route
+      }
     end
 
     def self.convert_availability_dates_into_indices(element, hash, start_index, end_index, type)
--- a/wrappers/wrapper.rb
+++ b/wrappers/wrapper.rb
@@ -697,7 +697,6 @@ module Wrappers
       unfeasible = []
       vehicle_max_shift = compute_vehicles_shift(vrp.vehicles)
       vehicle_max_capacities = compute_vehicles_capacity(vrp)
-      available_vehicle_skillsets = vrp.vehicles.flat_map(&:skills).uniq
 
       vrp.services.each{ |service|
         service.quantities.each{ |qty|
@@ -707,18 +706,7 @@ module Wrappers
           end
         }
 
-        if !service.skills.empty?
-          if service.sticky_vehicles.empty?
-            if available_vehicle_skillsets.none?{ |skillset| (service.skills - skillset).empty? }
-              add_unassigned(unfeasible, vrp, service, 'Service skill combination is not available on any vehicle')
-            end
-          elsif service.sticky_vehicles.all?{ |vehicle| vehicle.skills.none?{ |skillset| (service.skills - skillset).empty? } }
-            add_unassigned(unfeasible, vrp, service, 'Incompatibility between service skills and sticky vehicles')
-          end
-        end
-
         activities = [service.activity, service.activities].compact.flatten
-
         if activities.any?{ |a| a.timewindows.any?{ |tw| tw.start && tw.end && tw.start > tw.end } }
           add_unassigned(unfeasible, vrp, service, 'Service timewindows are infeasible')
         end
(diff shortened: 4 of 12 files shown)


braktar (approved):

fonsecadeline:
This is ready but it does not exactly correspond to our user need : we need to do this only for vehicles partitions. I suggest that we discuss this on corresponding gitlab ticket.

models/vrp.rb
```diff
@@ -391,6 +391,16 @@ def self.remove_unnecessary_relations(hash)
         uniq_relations += relations_set.uniq
       }
       hash[:relations] = uniq_relations
+
+      return unless hash[:configuration][:preprocessing].to_h[:partitions].to_a.none?{ |p| p[:entity] == :vehicle } &&
```
braktar, line 395: if+any will avoid the double negation here
```suggestion
      return if hash[:configuration][:preprocessing].to_h[:partitions].to_a.any?{ |p| p[:entity] == :vehicle } &&
```

models/vrp.rb
```diff
@@ -694,6 +704,88 @@ def calculate_service_exclusion_costs(type = :time, force_recalc = false)
       nil
     end
 
+    def service_reachable_by_any_vehicle(service)
```
braktar, line 707: Are these change required by the introduction of same_vehicle ?

If so, i you prefer to put them in a brand new class or module instead of just moving them into vrp.rb.

Any idea of the design pattern we could apply here ?

braktar (changes requested):

models/vrp.rb
fonsecadeline: In fact this needs to be improved with : https://gitlab.com/mapotempo/optimizer-api/-/issues/817

fonsecadeline: Right, I could first move `service_reachable_by_vehicle` and `service_uncompatible_with_vehicle_timewindows` to service.rb.

Do you have any further suggestions ? @senhalil ?

fonsecadeline: Discussed during daily : we move those to wrapper.rb. In the later this will edit service.compatible_vehicles

wrappers/wrapper.rb
```diff
@@ -693,6 +693,34 @@ def compute_vehicles_capacity(vrp)
...
+          }
+
+        common_vehicle = compatible_vehicles[0].any?{ |vehicle|
+          compatible_vehicles[1..-1].all?{ |set| set.include?(vehicle) }
+        }
+
+        next if common_vehicle
+
+        relation.linked_services.each{ |service|
+          add_unassigned(unfeasible, vrp, service, 'Service in same_route or same_vehicle relation with uncompatible services')
```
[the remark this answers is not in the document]
senhalil:
```suggestion
          add_unassigned(unfeasible, vrp, service, 'Service is in a same_route or same_vehicle relation with services that cannot be serviced by the same vehicle')
```

wrappers/wrapper.rb
```diff
@@ -693,6 +693,34 @@ def compute_vehicles_capacity(vrp)
       capacities.reject{ |_k, v| v.nil? || v.negative? }
     end
 
+    def reject_services_from_inconsistent_relations(vrp, unfeasible)
```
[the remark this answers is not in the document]
senhalil: I am still not sure if we need this check or not. Imagine 5 services marked as `same_route`; 3 of them has a common vehicle and 2 of them has another common vehicle -- i.e., we cannot serve all with the same vehicle. 

If we do nothing or-tools will serve either 2 or 3 of these services respecting  `same_route` and unassign/skip the remaining part (3 or 2, respectively). Since or-tools will take into account all the constraints (capacity, duration) and objective function (distance, duration) to select which "group" to plan and which "group" to unassign; we shouldn't do this elimination beforehand.  

With the suggested code, we unassign all services of the same `same_route` relation because of a single incompatible service; but this would make it really hard to use `same_route` and `same_vehicle` relations. It would force the user to make sure all services of a `same_route/vehicle` relation can be "actually" done on the same vehicle. But instead of forcing them we can just do "what we can" and the unplanned part will stay unplanned. (*)

I think we should let or-tools decide who to assign, who to unassign.

* : We might want to do this check as a post-process ! over unassigned services (and check if it was ever possible to process all of the in the same vehicle/route and if that is not the case, add this "reasoning" only to the unassigned ones). This is kind a out-of-scope of this PR (and since we don't have a proper post-processing mechanism yet, it would take too long), we can create a ticket for this -- by copying this code to that ticket (with the suggestions I just made) or giving the link of this comment so that we don't rediscover the wheel.

wrappers/wrapper.rb
```diff
@@ -739,87 +767,108 @@ def detect_unfeasible_services(vrp)
...
+          service_uncompatible_with_vehicle_timewindows(vrp, activity, vehicle)
         }
+      }
+    end
 
-        if !reachable_by_a_vehicle
-          add_unassigned(unfeasible, vrp, service, 'Service cannot be served due to vehicle parameters -- e.g., timewindow, distance limit, etc.')
-          next
-        end
+    def service_uncompatible_with_vehicle_timewindows(vrp, activity, vehicle)
```
senhalil, line 833: service_incompatible_with_vehicle_timewindows

wrappers/wrapper.rb
```diff
@@ -739,87 +767,108 @@ def detect_unfeasible_services(vrp)
...
+
+      vrp.services.each{ |service|
+        if !service_reachable_by_any_vehicle(vrp, service)
+          add_unassigned(
+            unfeasible, vrp, service, 'Service cannot be served due to vehicle parameters -- e.g., timewindow, distance limit, etc.')
+          next
         end
 
-        add_unassigned(unfeasible, vrp, service, 'Service cannot be reached within its timewindows') if service_unreachable_within_its_tw
+        next unless service_unreachable_within_its_tw(vrp, service)
```
senhalil, line 869: Why don't we reorder this logic like you did in `reject_services_from_inconsistent_relations` code so that timewindow logic is called only for "compatible" vehicles? (This is less ideal than the [:compatible_vehicles] idea that we first tried but it doesn't need more dev, even if deletes extra code.)

``` ruby
              (service.activity ? [service.activity] : service.activites).any?{ |activity|
                service_reachable_by_vehicle(vrp, activity, vehicle) &&
                  service_uncompatible_with_vehicle_timewindows(vrp, activity, vehicle) 
```

Then we can use a more encompassing reason like `There is no vehicle (with compatible skills, timewindows, duration, depot distance, etc.) that can serve this service`.

senhalil (changes requested):

lib/interpreters/split_clustering.rb
```diff
@@ -575,6 +576,10 @@ def self.build_partial_service_vrp(service_vrp, partial_service_ids, available_v
       points_ids = sub_vrp.services.map{ |s| s.activity.point.id }.uniq.compact
       sub_vrp.rests = sub_vrp.rests.select{ |r| sub_vrp.vehicles.flat_map{ |v| v.rests.map(&:id) }.include? r.id }
       sub_vrp.relations = sub_vrp.relations.select{ |r| r.linked_ids.all? { |id| sub_vrp.services.any? { |s| s.id == id } } }
+      if entity == :vehicle
+        sub_vrp.relations.delete_if{ |r| r.type == :same_vehicle }
+        sub_vrp.services.each{ |s| s.relations.delete_if{ |r| r.type == :same_vehicle } }
+      end
```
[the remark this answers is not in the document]
fonsecadeline: I dealt with this relation type like this : if no vehicle partition is in current vrp, there should be no same_vehicle relation : 

1. if there is a vehicle partition, then we should respect same_vehicle constraints in this partition. Once vehicle cluster is done then there is no vehicle partition in vrp anymore so we can ignore this relation because it is already respected by default.
2. if no vehicle partition is provided then this should be replaced by same_route relation

I think it is a good way to do this instead of editing a global variable LINKING_RELATIONS according to some parameters.

models/concerns/expand_data.rb
```diff
@@ -41,10 +41,13 @@ def add_sticky_vehicle_if_routes_and_partitions
...
-      self.routes.each{ |route|
-        route.mission_ids.delete_if{ |id| id == unfeasible_service[:original_service_id] }
-      }
+    unfeasible_services_ids = unfeasible_services.collect{ |un| un[:original_service_id] }
+    self.routes.each{ |route|
+      route.mission_ids.delete_if{ |id| unfeasible_services_ids.include?(id) }
+    }
+
+    self.relations.each{ |relation|
+      relation.linked_services.delete_if{ |service| unfeasible_services_ids.include?(service.id) }
```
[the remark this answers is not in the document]
fonsecadeline: Right, I remember having a bug because of the fact that some relations where pointing at services that do not exist anymore in vrp (because they were rejected by detect_unfeasible) and I did not realize this is already taken into account. 

I can indeed skip these lines.

models/concerns/expand_data.rb
```diff
@@ -41,10 +41,13 @@ def add_sticky_vehicle_if_routes_and_partitions
...
 
   def clean_according_to(unfeasible_services)
-    unfeasible_services.each{ |unfeasible_service|
-      self.routes.each{ |route|
-        route.mission_ids.delete_if{ |id| id == unfeasible_service[:original_service_id] }
-      }
+    unfeasible_services_ids = unfeasible_services.collect{ |un| un[:original_service_id] }
+    self.routes.each{ |route|
+      route.mission_ids.delete_if{ |id| unfeasible_services_ids.include?(id) }
+    }
```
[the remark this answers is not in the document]
fonsecadeline: I did not see this comment and said almost the same thing in my previous answer :)

lib/interpreters/split_clustering.rb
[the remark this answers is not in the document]
senhalil: I just check the code and it is `split_balanced_kmeans` who is telling `build_partial_vrp` what the `entity` is; so I am okay. 

Apparently I hadn't check where the `entity` was coming from when I asked my question. Because I though it was the entity coming from the duplicated original vrp.preprocessing_partitions. 

Otherwise, we are on the same page  ;) 

I guess I also imagined it a bit differently -- a logic closer to a place where we handle the `vrp.preprocessing_partitions` separation but I guess this a place as good as any.

models/concerns/expand_data.rb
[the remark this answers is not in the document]
senhalil: In theory, we should clean all non existent objects but I don't think `clean_according_to` is the right way to do it. So yes you can remove the route cleaning if ortools has the same protections wrt invalid ids.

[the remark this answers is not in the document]
senhalil: I think the right way to do all this cleaning is "add_unassigned" (even the `vrp.services -= unasinged` logic should be there) -- that is, the meaning of "filter infeasible" should include all the related logic of removing a service...

wrappers/wrapper.rb
```diff
@@ -739,87 +767,108 @@ def detect_unfeasible_services(vrp)
...
+        feasible &&= (vehicle_available_time >= min_time)
+      end
 
-              feasible &&= (vehicle_available_time >= min_time)
-            end
+      if feasible &&
+         (start_index || end_index) &&
+         vehicle.cost_distance_multiplier&.positive? &&
+         vehicle.distance
+        min_dist = (start_index && matrix[:distance][start_index][service_index]).to_f +
```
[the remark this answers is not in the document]
senhalil: Ah, I must have missed the if condition :facepalm: sorry ^o^

wrappers/wrapper.rb
```diff
@@ -742,89 +748,69 @@ def detect_unfeasible_services(vrp)
...
             }
           }
-        end
 
-        add_unassigned(unfeasible, vrp, service, 'Service cannot be reached within its timewindows') if service_unreachable_within_its_tw
-      }
+        next unless no_vehicle_compatible
 
-      log "Following services marked as infeasible:\n#{unfeasible.group_by{ |u| u[:reason] }.collect{ |g, set| "#{(set.size < 20) ? set.collect{ |s| s[:service_id] }.join(', ') : "#{set.size} services"}\n with reason '#{g}'" }.join("\n")}", level: :debug unless unfeasible.empty?
+        add_unassigned(unfeasible, vrp, service, 'No compatible vehicle can reach this service while respecting all constraints')
```
senhalil, line 807:
```suggestion
        add_unassigned(unfeasible, vrp, service, 'No compatible vehicle can reach this service within its time window while respecting all other constraints')
```

wrappers/wrapper.rb
```diff
@@ -742,89 +748,69 @@ def detect_unfeasible_services(vrp)
...
+      if !activity.late_multiplier&.positive? && activity.timewindows.any?(&:end)
+        return false if activity.timewindows.all?{ |tw| tw.end && earliest_arrival > tw.end }
+      end
 
-            if feasible &&
-               (start_index || end_index) &&
-               vehicle.cost_distance_multiplier&.positive? &&
-               vehicle.distance
-              min_dist = (start_index && matrix[:distance][start_index][index]).to_f + (end_index && matrix[:distance][index][end_index]).to_f
+      if vehicle_available_time && vehicle.cost_time_multiplier&.positive? && !vehicle.cost_late_multiplier&.positive?
```
senhalil, line 767: `vehicle_available_time` always exists what we want is `vehicle_end != 2**56`, right ?

wrappers/wrapper.rb
```diff
@@ -742,89 +748,69 @@ def detect_unfeasible_services(vrp)
...
+      end
 
-              feasible &&= (vehicle.distance >= min_dist)
-            end
+      if !vehicle.cost_late_multiplier&.positive? && activity.timewindows.any?{ |tw| tw.start.positive? } &&
+         (vehicle.timewindow&.end || vehicle.sequence_timewindows.any?(&:end))
+        latest_arrival = vehicle_end - activity.duration
+        latest_arrival -= matrix[:time][service_index][vehicle.end_point.matrix_index] if vehicle.end_point_id
+        return false if activity.timewindows.all?{ |tw| latest_arrival < tw.start }
+      end
```
senhalil, line 778: I suspect the first two checks above capture this case, don't they?

wrappers/wrapper.rb
senhalil: If the next check was in fact unnecessary, and if we are going to delete it then we can 
let `vehicle_end = vehicle.timewindow&.end || vehicle.sequence_timewindows.collect(&:end).max`
and do the check with `vehicle_end.nil?` here
and generate the `vehicle_available_time = vehicle_end - vehicle_start` inside this if condition 
instead of calculating `2**56` multiple times

wrappers/wrapper.rb
```diff
@@ -742,89 +748,69 @@ def detect_unfeasible_services(vrp)
...
 
-            feasible
-          }
-        }
+      if v_start_index && v_end_index && vehicle.distance && vehicle.cost_distance_multiplier&.positive?
+        min_dist = matrix[:distance][v_start_index][service_index].to_f +
+                   matrix[:distance][service_index][v_end_index].to_f
+        return false if min_dist > vehicle.distance
+      end
```
senhalil, line 784: We can just do as follows because if both `v_start_index` and `v_end_index` are `nil` then the `min_distance` will be `0` and `vehicle.distance` cannot be negative but if one of them us non-nil then we want to check the distance!
```suggestion
      if vehicle.distance && vehicle.cost_distance_multiplier&.positive?
        min_dist = (v_start_index && matrix[:distance][v_start_index][service_index]).to_f +
                   (v_end_index && matrix[:distance][service_index][v_end_index]).to_f
        return false if min_dist > vehicle.distance
      end
```
which translates into to correct if condition `(v_start_index || v_end_index)` and not `v_start_index && v_end_index`

senhalil (changes requested):
Almost there but I think I have found some bugs. (didn't check the tests yet, will do once the rest is done)

wrappers/wrapper.rb
```diff
@@ -659,6 +659,12 @@ def add_unassigned_internal(unfeasible, vrp, service, reason)
         }
       end
 
+      vrp.routes.each{ |route| route.mission_ids.delete(service.id) }
+      vrp.relations.each{ |relation|
+        relation.linked_ids.delete(service.id)
+        relation.linked_services.delete(service)
+      }
+
```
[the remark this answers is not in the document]
fonsecadeline: Periodic heuristic already ensures that we do not initialize routes with services that do not exist in current VRP so I will just remove these delete.

Periodic does not do anything on relations because current version of periodic heuristic does not support relations.

wrappers/wrapper.rb
```diff
@@ -742,89 +748,69 @@ def detect_unfeasible_services(vrp)
...
-              end
-              timely_arrival_not_possible || timely_return_not_possible
+      vrp.services.each{ |service|
+        no_vehicle_compatible =
+          vrp.vehicles.none?{ |vehicle|
+            (service.activity ? [service.activity] : service.activities).any?{ |activity|
+              (unfeasible.any?{ |s| s[:service_id] == service.id } ||
+                (service.skills.empty? || vehicle.skills.any?{ |skill_set| (service.skills - skill_set).empty? }) &&
+                (service.sticky_vehicles.empty? || service.sticky_vehicles.include?(vehicle))) &&
+                service_reachable_by_vehicle_within_timewindows(vrp, activity, vehicle)
```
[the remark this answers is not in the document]
fonsecadeline: I agree with what you say and I initially did it like you suggest. However in the case where skills are incompatible this 'repeats' the reason of unassignment. 

In the general case I would suggest that we only do the skills check in this new version of the code. However, in the case where services have sticky vehicles then we would not specify that vehicles in sticky_vehicles list do not have consistent skills.

Maybe this is not a big deal. I can use your suggestion.

wrappers/wrapper.rb
fonsecadeline: Yes I will do something like that. We still need || 2**56 because of next check. 

Instead of checking vehicle_available_time I will just check vehicle_end < 2**56

wrappers/wrapper.rb
fonsecadeline: I do not think it is the case for now but we can edit second check in order to remove this one.

fonsecadeline: I may have mixed some things because we check for vehicle.cost_late_multiplier and then check with service.timewindows. This is not very consistent.

fonsecadeline: In fact timewindow only regards "arrival time" so it is much easier than I thought.

senhalil: oh maybe one of them is for service tw and the other for vehicle tw :/ I suspect they can be cleanly handled ytogether but if not we can keep them seperate. you decide

wrappers/wrapper.rb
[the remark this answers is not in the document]
senhalil: I agree :point_right:  "In the general case I would suggest that we only do the skills check in this new version of the code."

and when we are moving the infeasibility checks to the "simplification -- i.e., pre/post processing"  structure, we will overhaul the reasons and make them uniform and coherent.

wrappers/wrapper.rb
```diff
@@ -742,89 +730,62 @@ def detect_unfeasible_services(vrp)
...
-               !vehicle.cost_late_multiplier&.positive? &&
-               ((vehicle.timewindow&.start && vehicle.timewindow&.end) || vehicle.sequence_timewindows&.size&.positive?)
-              vehicle_available_time = vehicle.timewindow.end - vehicle.timewindow.start if vehicle.timewindow
-              vehicle_available_time = vehicle.sequence_timewindows.collect{ |tw| tw.end - tw.start }.max if vehicle.sequence_timewindows&.size&.positive?
+      earliest_arrival = vehicle_start + time_to_go
+      earliest_back = earliest_arrival + activity.duration + time_back
+      latest_arrival = vehicle_end - time_back - activity.duration if vehicle_end
 
-              min_time = (start_index && matrix[:time][start_index][index]).to_f + (end_index && matrix[:time][index][end_index]).to_f + activity.duration
+      unless activity.late_multiplier && activity.timewindows.none?(&:end)
```
senhalil, line 745: I suspect there is a bug here but I can't put my head around... :thinking:  `unless` with multiple conditions is just turning my head (and apparently I am not the only person, I suggest -> http://www.railstips.org/blog/archives/2008/12/01/unless-the-abused-ruby-conditional/ )

If I didn't make a mistake, your `unless` statement translates into the following
`if !activity.late_multiplier || activity.timewindows.any?(&:end)`
but I suspect you wanted something like this
`if !activity.late_multiplier && activity.timewindows.any?(&:end)`
which can also be written
`if activity.late_multiplier.nil? && activity.timewindows.any?(&:end)`
but this has two problems:

First, `activity.late_multiplier = 0` also means no lateness allowed (I know it is not straight forward, and it should have mean "being late is free" but it is not :( )

However, even with the correct condition, we need to separate the conditions because existence of `late_multiplier` and lack of `tw.end` shouldn't effect the `latest_arrival < tw.start` check.

I will make a suggestion :point_down:

wrappers/wrapper.rb
```diff
@@ -742,89 +730,62 @@ def detect_unfeasible_services(vrp)
...
+        return false if activity.timewindows.all?{ |tw| tw.end && earliest_arrival > tw.end }
+        return false if latest_arrival && activity.timewindows.all?{ |tw| latest_arrival < tw.start }
+      end
```
senhalil, line 748:
```suggestion
      if !activity.late_multiplier&.positive? # if service lateness is not allowed
        return false if activity.timewindows.none?{ |tw| tw.end.nil? || earliest_arrival <= tw.end }
      end
      
      return false if latest_arrival && activity.timewindows.all?{ |tw| latest_arrival < tw.start }
```

senhalil: I paused here will continue after the break

senhalil (changes requested):
This is much much cleaner and easier to follow than the original convoluted check_distance logic! 

there still are some bugs to fix though

wrappers/wrapper.rb
```diff
@@ -742,89 +730,63 @@ def detect_unfeasible_services(vrp)
...
+      
+      return false if latest_arrival && activity.timewindows.all?{ |tw| latest_arrival < tw.start }
 
-              feasible &&= (vehicle_available_time >= min_time)
-            end
+      unless vehicle.cost_late_multiplier&.positive? && vehicle.cost_time_multiplier&.positive? &&
+             (vehicle_end || vehicle.duration)
+        return false if vehicle_end && earliest_back > vehicle_end
+        return false if vehicle.duration && earliest_back - earliest_arrival > vehicle.duration
+      end
```
senhalil, line 757: This `unless` statement looks incorrect to me, it should be something like this
```suggestion
      if vehicle.cost_time_multiplier&.positive? && !vehicle.cost_late_multiplier&.positive? # lateness is not allowed
        return false if vehicle_end && earliest_back > vehicle_end
        return false if vehicle.duration && earliest_back - earliest_arrival > vehicle.duration
      end
```

wrappers/wrapper.rb
```diff
@@ -742,89 +730,65 @@ def detect_unfeasible_services(vrp)
...
-            feasible
-          }
-        }
-
-        if !reachable_by_a_vehicle
-          add_unassigned(unfeasible, vrp, service, 'Service cannot be served due to vehicle parameters -- e.g., timewindow, distance limit, etc.')
-          next
-        end
+      dist_to_go + dist_back < vehicle.distance
+    end
```
senhalil, line 766: It is more neat and in phase with the remaining check as follows

```ruby
      if vehicle.distance && vehicle.cost_distance_multiplier&.positive?
        # check distances constraints
        dist_to_go = vehicle.start_point&.matrix_index ? matrix.distance[vehicle.start_point&.matrix_index][activity.point.matrix_index] : 0
        dist_back = vehicle.end_point&.matrix_index ? matrix.distance[activity.point.matrix_index][vehicle.end_point&.matrix_index] : 0

        return false if dist_to_go + dist_back < vehicle.distance
      end

      return true
    end
```

senhalil (changes requested):
I think I have found one more bug

senhalil (approved):
It looks ready to me. 

I checked the performance test and the new code is 50% faster :rocket:

PR #257, opened 2021-07-19
Move ci to one build

alain-andre:
Since the docker build can take a lot of time because of writing on the disk, Halil proposed to build once and use artifact to get the image and launch tests. As the new *docker/build-push-action@v2* allows to store image in artifact, this PR uses it.

To limit the number of pipelines, I changed the workflow. It won't launch pipelines on Draft PR. This PR also uses layer caching.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -3,30 +3,66 @@ name: Docker Image CI
 on:
   push:
   pull_request:
-    types: [opened, synchronize, reopened]
+    types: [opened, synchronize, reopened, ready_for_review]
 
 env:
   NB_SERVICES: ${{ secrets.NB_SERVICES }}
-  OPTIMIZER_ORTOOLS_VERSION: ${{ secrets.OPTIMIZER_ORTOOLS_VERSION }}
   PROJECT: optimizer
-  REGISTRY: ${{ secrets.REGISTRY }}
-  VROOM_VERSION: ${{ secrets.VROOM_VERSION }}
 
 jobs:
-  test_basis:
+  build:
+    if: github.event.pull_request.draft == false
     runs-on: ubuntu-latest
-    env:
-      OPTIONS: "COV=false LOG_LEVEL=info SKIP_DICHO=true SKIP_REAL_CASES=true SKIP_PERIODIC=true SKIP_SPLIT_CLUSTERING=true"
     steps:
-      - uses: actions/checkout@v2
+      - name: Checkout
+        uses: actions/checkout@v2
       - name: Cache
         id: cache
         uses: actions/cache@v2
         with:
           path: vendor/bundle
           key: ${{ runner.os }}
-      - name: Build image
-        run: ./.github/actions/build.sh
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+      - name: Build
+        uses: docker/build-push-action@v2
+        with:
+          context: .
+          file: docker/Dockerfile
+          build-args: |
+            BUNDLE_WITHOUT=production
+            OPTIMIZER_ORTOOLS_VERSION=${{ secrets.OPTIMIZER_ORTOOLS_VERSION }}
+            VROOM_VERSION=${{ secrets.VROOM_VERSION }}
+          tags: registry.test.com/mapotempo/optimizer-api:latest
+          outputs: type=docker,dest=/tmp/optimizer-api.tar
+      - name: Extract gems from image to cache folder
+        run: |
+          mkdir -p vendor/bundle
+          docker load --input /tmp/optimizer-api.tar
+          docker cp "$(docker create --rm registry.test.com/mapotempo/optimizer-api:latest)":/srv/app/vendor/bundle/. vendor/bundle
+      - name: Upload artifact
+        uses: actions/upload-artifact@v2
+        with:
+          path: /tmp/optimizer-api.tar
+
+  test_basis:
+    needs: build
+    runs-on: ubuntu-latest
+    env:
+      OPTIONS: "COV=false LOG_LEVEL=info SKIP_DICHO=true SKIP_REAL_CASES=true SKIP_PERIODIC=true SKIP_SPLIT_CLUSTERING=true"
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Set up Docker Buildx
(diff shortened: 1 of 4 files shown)


senhalil:
@alain-andre while you are at it, can you also extract the image creation (and upload) logic to a separate step that executes at the very end if and only if all previous test steps pass correctly?

Can we also start using the (ruby/setup-ruby@v1) action so that the gem caching is handled by github-action automatically and we don't have to do manual cache operations*? That way docker can point to the install directory of this action (I believe by default it installs to `$PWD/vendor/bundle`) to use the auto installed/cached gems. Here is the action https://github.com/ruby/setup-ruby#caching-bundle-install-automatically

*: Ruby specifically warns against manual gem caching (which is what we currently do if I am not mistaken) https://github.com/ruby/setup-ruby#caching-bundle-install-manually

senhalil:
@alain-andre can you correct the CI setting so that tests are launched with modification as well? 

Without `synchronise`, each time a correction/push is made to a PR, we will need to close/re-open the PR or draft/un-draft it. This is quite tedious and will create unnecessary e-mail noise. Since there is no limit on github action for open-source projects, this means we would lose time and gain nothing. We already decrease the registry operations 7x by moving the builds and uploads out of the tests, we don't really need to change the CI settings.

alain-andre:
> @alain-andre can you correct the CI setting so that tests are launched with modification as well?
> 
> Without `synchronise`, each time a correction/push is made to a PR, we will need to close/re-open the PR or draft/un-draft it. This is quite tedious and will create unnecessary e-mail noise. Since there is no limit on github action for open-source projects, this means we would lose time and gain nothing. We already decrease the registry operations 7x by moving the builds and uploads out of the tests, we don't really need to change the CI settings.

Ok, I changed this, now it won't launch CI on Drafted PR only.

alain-andre:
> @alain-andre while you are at it, can you also extract the image creation (and upload) logic to a separate step that executes at the very end if and only if all previous test steps pass correctly?
> 
> Can we also start using the (ruby/setup-ruby@v1) action so that the gem caching is handled by github-action automatically and we don't have to do manual cache operations*? That way docker can point to the install directory of this action (I believe by default it installs to `$PWD/vendor/bundle`) to use the auto installed/cached gems. Here is the action https://github.com/ruby/setup-ruby#caching-bundle-install-automatically
> 
> *: Ruby specifically warns against manual gem caching (which is what we currently do if I am not mistaken) https://github.com/ruby/setup-ruby#caching-bundle-install-manually

I tried, but it doesn't work. We are working inside a container with a different system and libraries. The generation of the gem on the system used to build the image can handle in the gem that will not be usable inside the container. That's why I copy the gems generated from the image after it's been built and let GitHub cache them.

Caching gems can be dangerous because of what they specify in the readme and especially on Github Action because we can't clear the cache like on Gitlab. But we can't do it otherwise.

senhalil:
> I tried, but it doesn't work. We are working inside a container with a different system and libraries. The generation of the gem on the system used to build the image can handle in the gem that will not be usable inside the container. That's why I copy the gems generated from the image after it's been built and let GitHub cache them.

Ah :/ thanks for trying @alain-andre! I think I understood why it doesn't work. I think the following lik might interest you. It explains how buildx can cache docker layers, I suspect this would speed up the whole image creation process (and it might be even better than caching bundle install) since we do not change much in docker layers (of course the docker layers might need reordering to get the best of this caching mechanism -- i.e., least changing comes/installs first, etc)  https://evilmartians.com/chronicles/build-images-on-github-actions-with-docker-layer-caching

---

By the way the following question was seperate from the other one. We still can do the image upload at the end, right? If the tests are not successful, we are currently bothering the registry to upload a broken code for nothing.
> @alain-andre while you are at it, can you also extract the image creation (and upload) logic to a separate step that executes at the very end if and only if all previous test steps pass correctly?

.github/workflows/main.yml
```diff
@@ -7,148 +7,208 @@ on:
...
   NB_SERVICES: ${{ secrets.NB_SERVICES }}
-  OPTIMIZER_ORTOOLS_VERSION: ${{ secrets.OPTIMIZER_ORTOOLS_VERSION }}
   PROJECT: optimizer
-  REGISTRY: ${{ secrets.REGISTRY }}
-  VROOM_VERSION: ${{ secrets.VROOM_VERSION }}
 
 jobs:
-  test_basis:
+  build:
+    if: github.event.pull_request.draft == false
```
senhalil, line 14: Is there a variable that distinguishes manually launched workflow? If so, we can skip drafts unless the workflow is launched manually. This would prevent an unnecessary un-draft/draft cycle.

https://github.com/asyncapi/.github/issues/13#issue-748881144 if this link is correct, we need to add `ready_for_review` to the `types` parameter; otherwise, even marking the PR as ready is not enough to launch the tests and one needs to do un-draft+closePR+openPR to launch the tests.

alain-andre: Indeed adding *ready_for_review* is a plus as it will automatically trigger the pipeline when un-drafting the PR. 

There is no variable that distinguishes manually launched workflow as triggering the pipeline manually means you want to run it.

alain-andre:
> Ah :/ thanks for trying @alain-andre! I think I understood why it doesn't work. I think the following lik might interest you. It explains how buildx can cache docker layers, I suspect this would speed up the whole image creation process (and it might be even better than caching bundle install) since we do not change much in docker layers (of course the docker layers might need reordering to get the best of this caching mechanism -- i.e., least changing comes/installs first, etc) https://evilmartians.com/chronicles/build-images-on-github-actions-with-docker-layer-caching

I had seen this, I keep it in mind for the moment for a next PR.

> By the way the following question was seperate from the other one. We still can do the image upload at the end, right? If the tests are not successful, we are currently bothering the registry to upload a broken code for nothing.
> 
> > @alain-andre while you are at it, can you also extract the image creation (and upload) logic to a separate step that executes at the very end if and only if all previous test steps pass correctly?

I did not understand the question the first time. In fact we do not upload to the registry at any time but we need to build the image at first step. This said, I'm wondering why we do not build on tag push; as we should.

My next PR will handle both things : push to register on tag and caching layouts.

braktar (approved):

fonsecadeline (approved):

senhalil (approved):
I think this is ready to merge :rocket:

PR #260, opened 2021-08-02
Clean double IDs in relation

fonsecadeline:


test/models/vrp_consistency_test.rb
```diff
@@ -571,5 +571,24 @@ def test_pickup_timewindow_after_delivery_timewindow
 
       OptimizerWrapper.wrapper_vrp('demo', { services: { vrp: [:demo] }}, TestHelper.create(problem), nil)
     end
+
+    def test_unicity_of_provided_services_or_vehicles_in_relation
```
senhalil, line 575: uniqueness

@fonsecadeline can you activate  `Code Spell Checker`?

models/concerns/validate_data.rb
```diff
@@ -139,6 +139,14 @@ def check_services(schedule)
   def check_relation_consistent_timewindows
     unconsistent_relation_timewindows = []
     @hash[:relations].each{ |relation|
+      unless relation[:linked_ids].uniq == relation[:linked_ids]
```
senhalil, line 142: relation[:linked_ids]&.uniq

models/concerns/validate_data.rb
```diff
@@ -139,6 +139,14 @@ def check_services(schedule)
...
+        raise OptimizerWrapper::DiscordantProblemError.new('Same service ID was provided several times in one relation.')
```
senhalil, line 143: it can be service or shipment so, it is better as follows:

Same ID is provided multiple times in the same relation

models/concerns/validate_data.rb
```diff
@@ -139,6 +139,14 @@ def check_services(schedule)
...
+      end
+
+      unless relation[:linked_vehicle_ids].uniq == relation[:linked_vehicle_ids]
+        raise OptimizerWrapper::DiscordantProblemError.new('Same vehicle ID was provided several times in one relation.')
```
senhalil, line 147: There is a subtle difference between several and multiple. We want "multiple" because it means "more than once" -- or we can directly use 'more than once' in both sentences.

Same vehicle ID is provided multiple times in the same relation.
Same vehicle ID is provided in the same relation more than once.

senhalil (changes requested):

test/models/vrp_consistency_test.rb
fonsecadeline: It is activated but does not work with ruby. I did not find an alternative yet. 
However, unicity would not have been shown because the word does exist according to google translate + reverso.
Still, I will edit this.

senhalil: I don't know why it doesn't work for you, it works for me and it does underline that word. 

Linguistically speaking yes, that word exists; however, we are not linguists, google search `define unicity` specifically mentions that this is a "rare" word and the corresponding french search `define unicité` only gives "uniqueness" as translation.

fonsecadeline: Bizarre... The documentation page specifically lists available languages and ruby is not in it. Are you sure you have no other extension for this ?

senhalil (changes requested):

senhalil (approved):

PR #265, opened 2021-08-11
Relation[:lapse] become relation[:lapses]

fonsecadeline:


models/concerns/validate_data.rb
```diff
@@ -128,6 +128,12 @@ def check_services(schedule)
   def check_relation_consistent_timewindows
     unconsistent_relation_timewindows = []
     @hash[:relations].each{ |relation|
+      if Models::Relation::ON_VEHICLES_TYPES.include?(relation[:type]) && relation[:linked_ids].to_a.any?
+        raise OptimizerWrapper::DiscordantProblemError.new("#{relation[:type]} does not expect any service ids")
+      elsif Models::Relation::ON_SERVICES_TYPES.include?(relation[:type]) && relation[:linked_vehicle_ids].to_a.any?
+        raise OptimizerWrapper::DiscordantProblemError.new("#{relation[:type]} does not expect any vehicle ids")
```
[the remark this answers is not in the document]
senhalil: this check is not really about `relation_consistent_timewindows` maybe we can merge lapse_need checks and this vehicle/service_need checks into `check_relation_needs` or `check_relation_parameter_need`, `check_relation_argument_need` or something like that. Vehicle/service check should come first because we use linked_ids/linked_vehicle_ids in the lapse checks.

models/concerns/validate_data.rb
```diff
@@ -128,6 +128,12 @@ def check_services(schedule)
...
+      end
```
senhalil, line 135: in fact we can do this more cleanly with `case .. when`
```suggestion
      case relation[:type]
      when *Models::Relation::ON_VEHICLES_TYPES
        if relation[:linked_ids].to_a.any? || relation[:linked_vehicle_ids].to_a.empty? 
          raise OptimizerWrapper::DiscordantProblemError.new("#{relation[:type]} relations do not support linked_ids and expect linked_vehicle_ids")
        end
      when *Models::Relation::ON_SERVICES_TYPES
         # opposite of the above
      when  *Models::Relation::ON_NOTHING_TYPES # only if necessary 
         # then we raise if there is any linked_ids or linked_vehicle_ids
      else
        raise 'Unknown relation type' # in case we ever forget to handle a new relation type
      end
```

senhalil (changes requested):

senhalil (changes requested):

models/concerns/validate_data.rb
```diff
@@ -264,30 +270,62 @@ def check_trip_timewindows_consistency(relation_vehicles)
...
+    elsif Models::Relation::ONE_LAPSE_TYPES.include?(relation[:type])
+      if relation[:lapse].nil? && relation[:lapses].to_a.size != 1
+        raise OptimizerWrapper::DiscordantProblemError.new(
+          "#{relation[:type]} relations expect exactly one lapse"
+        )
+      end
+    elsif relation[:type] == :vehicle_trips
+      if relation[:lapses].to_a.size > relation[:linked_vehicle_ids].size - 1
+        raise OptimizerWrapper::DiscordantProblemError.new('Too many lapses provided for at least one relation')
+      end
```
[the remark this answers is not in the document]
senhalil: That is exactly what I mean, so we should raise but `relation[:lapses].to_a.size > relation[:linked_vehicle_ids].size - 1` check fails to handle that case. My suggestion [here](https://github.com/Mapotempo/optimizer-api/pull/265#discussion_r686970499)  should handle that case correctly if I am not mistaken.

models/concerns/validate_data.rb
```diff
@@ -264,30 +264,76 @@ def check_trip_timewindows_consistency(relation_vehicles)
...
+          "#{relation[:type]} relations expect exactly one lapse"
+        )
+      end
+    when *Models::Relation::SEVERAL_LAPSE_TYPES
+      if relation[:lapse].nil? && relation[:lapses].to_a.size != 1
+        exp_lapse_count = (relation[:linked_ids]&.size || relation[:linked_vehicle_ids]&.size) - 1
+        if relation[:lapses].to_a.size != exp_lapse_count
+          raise OptimizerWrapper::DiscordantProblemError.new(
+            "#{relation[:type]} relations expect at least one lapse"
+          ) unless relation[:type] == :vehicle_trips && relation[:lapses].to_a.empty?
```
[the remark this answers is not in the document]
senhalil: If I am not mistaken we already had such a logic `relation.lapse ||= 0` for other relations. I don't see why it should be a problem for `:vehicle_trips`..

This can be easily handled in the code I suggested [here](https://github.com/Mapotempo/optimizer-api/pull/265#discussion_r686970499) by adding a single line to the last `when` case

```ruby
relation[:lapses] = [0] if relation[:type] == :vehicle_trips && relation[:lapse].nil? && relation[:lapses].nil?
```
that way we don't need to care about this case rest of the code.

test/test_helper.rb
```diff
@@ -129,7 +129,17 @@ def self.coerce(vrp)
...
+      r[:type] = r[:type]&.to_sym
+      next if [Models::Relation::NO_LAPSE_TYPES,
+               Models::Relation::ONE_LAPSE_TYPES,
+               Models::Relation::SEVERAL_LAPSE_TYPES
+              ].any?{ |set|
+                set.include?(r[:type])
+              }
+
+      raise 'This relation does not exit in any of NO_LAPSE_RELATIONS ONE_LAPSE_RELATIONS SEVERAL_LAPSE_RELATIONS, there is a risk of incorrect management'
+    }
```
senhalil, line 142: Maybe we can do the same thing for `ON_VEHICLES_TYPES` `ON_SERVICE_TYPES` as well.

fonsecadeline: I will write a new loop over ON_VEHICLES_TYPES and ON_SERVICE_TYPES because having a type in ON_SERVICE_TYPES but in none of [NO_LAPSE_TYPES, ONE_LAPSE_TYPES, SEVERAL_LAPSE_TYPES] still is a problem I think.

senhalil: yeah yeah, exactly, I meant a separate verification. You can repeat the same logic in a separate loop or you can raise `if [lapse_types].none?{ include} || [id_types].none?{ inlcude}`.

senhalil:
@fonsecadeline I just realised we forgot to mention this change in the CHANGELOG.md, can you mention under "changed" section that lapse became lapses

senhalil (approved):

PR #268, opened 2021-08-20
Remove number in sequence logic + refacto

fonsecadeline:


lib/heuristics/periodic_heuristic.rb
```diff
@@ -1362,26 +1408,23 @@ def collect_generated_routes(vehicle, services)
 
     def save_status
       @previous_candidate_routes = Marshal.load(Marshal.dump(@candidate_routes))
```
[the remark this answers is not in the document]
fonsecadeline: You can see it on https://github.com/fonsecadeline/optimizer-api/commit/42d0fd2bb65c71d6e1d1c44b01196339934aafb6 commit.

test/real_cases_periodic_test.rb
```diff
@@ -164,10 +165,28 @@ def test_fill_days_and_post_processing
       # checks performance on instance calling post_processing
       vrp = TestHelper.load_vrp(self, fixture_file: 'periodic_with_post_process')
       vrp.resolution_minimize_days_worked = true
-      result = OptimizerWrapper.wrapper_vrp('ortools', { services: { vrp: [:ortools] }}, vrp, nil)
+      found_uninserted_visits = 1000 # we start with 1000 unassigned visits
```
fab-girard, line 168: why this initialization?

fonsecadeline: This line should in fact be after line 173 : we are simulating phase one of periodic heuristic by setting @candidate_routes value.

The aim of this test is to test end_phase (refine_solution). 

To sum up : this instance has 3733 visits and after first stage we have 1000 still unassigned. We expect end_phase to plan some of those. 

In fact you are right, we do not need to initialize this variable, a simple comment would be enough.

fonsecadeline: However, if I do not initialize this value here, I can not use it as I do in this test.

braktar: Is it possible to deduce the value 1000 from the data ?

fonsecadeline: We can deduce it from fill_days_and_post_processing_services_assignment.dump file but, still, that would be in stub function so not accepted.

fonsecadeline: I can move data initialization before stub, and compute this variable from it. It makes less sense for me.

lib/heuristics/concerns/periodic_end_phase.rb
```diff
@@ -106,52 +65,48 @@ def add_missing_visits
...
+      costs.delete(inserted_id)
     end
 
-    # update costs for all ids that should take place at this day. Route changed so cost can change too.
+    # Update costs on this route
     costs.each{ |id, info|
-      next if info[:day] != best_cost[1][:day] && info[:vehicle] != best_cost[1][:vehicle]
-
-      cost = find_best_day_cost(@candidate_routes[info[:vehicle]], id)
+      next unless info[:day] == inserted_cost[:day] &&
```
braktar, line 91: the logical equivalence of !a && !b is !(a || b)
```suggestion
      next unless info[:day] == inserted_cost[:day] ||
```

braktar (changes requested):

lib/heuristics/concerns/periodic_end_phase.rb
fonsecadeline: There was a mistake before : it should avec been 

```
info[:day] != best_cost[1][:day] || info[:vehicle] != best_cost[1][:vehicle]
```

We do want to have both same day and vehicle, otherwise it is a different route and it will not be impacted.

fab-girard (approved):

lib/heuristics/concerns/periodic_end_phase.rb
```diff
@@ -106,52 +65,49 @@ def add_missing_visits
...
-      available_days = days_respecting_lapse(best_cost[0], @candidate_routes[best_cost[1][:vehicle]])
-
-      cost = find_best_day_cost(@candidate_routes[best_cost[1][:vehicle]], best_cost[0], available_days)
-
-      if cost
-        costs[best_cost[0]] = cost
-      else
-        costs.delete(best_cost[0])
-      end
+  # Updates costs after insertion of inserted_cost
```
braktar, line 75: Do you mean `inserted_id` instead of `inserted_cost` ?
```suggestion
  # Updates costs after insertion of inserted_id
```

lib/heuristics/periodic_heuristic.rb
```diff
@@ -464,83 +476,83 @@ def reorder_stops(vrp)
...
+        @services_assignment[service_id][:days].each{ |day|
+          route_data = @candidate_routes[vehicle_id][day]
+          removed_index = route_data[:stops].find_index{ |stop| stop[:id] == service_id }
+          next unless removed_index
+
+          service_point = route_data[:stops][removed_index][:point_id]
+          route_data[:stops].slice!(removed_index)
+          @services_data[service_id][:capacity].each{ |need, qty| route_data[:capacity_left][need] += qty }
+
+          unless route_data[:stops].any?{ |s| s[:point_id] == service_point }
```
braktar, line 522: A formulation with `if` is easier to understand
```suggestion
          if route_data[:stops].none?{ |s| s[:point_id] == service_point }
```

braktar (approved):

PR #269, opened 2021-08-30
Split independent by vehicle relation and sticky

braktar:


fonsecadeline (changes requested):

optimizer_wrapper.rb
```diff
@@ -427,7 +466,7 @@ def self.split_independent_vrp(vrp)
     return [vrp] if (vrp.vehicles.size <= 1) || vrp.services.empty? # there might be zero services
 
     if vrp.services.all?{ |s| s.sticky_vehicles.size == 1 }
-      return split_independent_vrp_by_sticky_vehicle(vrp)
+      return split_independent_vrp_by_relation_and_sticky(vrp)
```
fonsecadeline, line 469: I think we should do something similar with _split_independent_vrp_by_skills_ and convert it into something like _split_independent_vrp_by_relation_and_skills_. Because now  _split_independent_vrp_by_skills_ might also split vehicles from a common relation.

fonsecadeline (changes requested):

optimizer_wrapper.rb
```diff
@@ -350,9 +339,31 @@ def self.split_independent_vrp_by_skills(vrp)
...
     }
 
+    vrp.relations.select{ |relation|
+      Models::Relation::ON_VEHICLES_TYPES.include?(relation.type)
+    }.each{ |relation|
+      v_skills_indices = relation.linked_vehicle_ids.map{ |v_id| correspondance_hash[v_id] }
+      mission_skill_indices = []
+      next unless v_skills_indices.any?{ |v_index|
+        mission_skills.each_index.any?{ |m_index|
+          mission_skill_indices << m_index if compatibility_table[m_index][v_index]
```
fonsecadeline, line 349: It seems dangerous to me to insert elements in this array in a "next unless" condition. In fact I do not quite understand what we want to check here.

optimizer_wrapper.rb
```diff
@@ -363,78 +374,84 @@ def self.split_independent_vrp_by_skills(vrp)
...
+    return [vrp] if (vrp.vehicles.size <= 1) || vrp.services.empty? || vrp.subtours&.any? # there might be zero services
 
-    if vrp.services.all?{ |s| s.sticky_vehicles.size == 1 }
-      return split_independent_vrp_by_sticky_vehicle(vrp)
-    end
-    # Cannot split if there is multimodal subtours
-    if !vrp.subtours&.any? && vrp.services.all?{ |s| s.sticky_vehicles.empty? }
-      return split_independent_vrp_by_skills(vrp)
-    end
+    # - a service with no service (or sticky vehicle see: expand_data.rb:sticky_as_skills)
```
fonsecadeline, line 433: This is not very important but I do not understand this comment, could you eventually rephrase so it is more explicit for all ?

optimizer_wrapper.rb
```diff
@@ -363,78 +374,84 @@ def self.split_independent_vrp_by_skills(vrp)
...
-    independant_skill_sets = independent_skills.map{ |index_set|
+    # Independent skill sets : Original skills are retrieved
+    independent_skills.map{ |index_set|
       index_set.collect{ |index| mission_skills[index] }
     }
+  end
 
-    unused_vehicles_indices = (0..vrp.vehicles.size - 1).to_a
-    independent_vrps = independant_skill_sets.collect{ |skills_set|
+  def self.build_independent_vrps(vrp, skill_sets, unused_vehicle_indices, skill_vehicle_ids, skill_service_ids)
```
fonsecadeline, line 383: Maybe _unused_vehicle_indices_ could be initialized here to reduce this function number of parameters.

optimizer_wrapper.rb
```diff
@@ -363,78 +374,84 @@ def self.split_independent_vrp_by_skills(vrp)
...
+    independent_vrps = skill_sets.collect{ |skills_set|
       # Compatible problem ids are retrieved
-      vehicles_indices = skills_set.flat_map{ |skills| skill_vehicle_ids.select{ |k, _v| (k & skills) == skills }.flat_map{ |_k, v| v } }.uniq
-      vehicles_indices.each{ |index| unused_vehicles_indices.delete(index) }
+      vehicle_indices = skills_set.flat_map{ |skills|
+        skill_vehicle_ids.select{ |k, _v| (k & skills) == skills }.flat_map{ |_k, v| v }
+      }.uniq
+      vehicle_indices.each{ |index| unused_vehicle_indices.delete(index) }
       service_ids = skills_set.flat_map{ |skills| skill_service_ids[skills] }
```
fonsecadeline, line 390: I am not sure _skill_service_ids_ keys correspond to _skill_sets_ after calling _compute_independent_skills_sets_. Are they?

fonsecadeline: Could you maybe add a test with trip relation and different skills on each vehicle_trip to check this ?

fonsecadeline (changes requested):

optimizer_wrapper.rb
braktar: As it is updated during through the generation of the multiple independent vrps, we must initialize it before the call this method.

optimizer_wrapper.rb
braktar: _skill_service_ids_ is untouched by _compute_independent_skills_sets_. It only give the service ids associated to a given skill set.

fonsecadeline (approved):

senhalil (commented):
I have a few questions

models/vrp.rb
```diff
@@ -237,6 +237,7 @@ def self.expand_data(vrp)
       vrp.add_sticky_vehicle_if_routes_and_partitions
       vrp.expand_unavailable_days
       vrp.provide_original_info
+      vrp.sticky_as_skills
```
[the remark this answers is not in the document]
braktar: To do so, I should move both  this method and the method filter_skills which is a prerequisite in order to have consistent skills. Nevertheless, I agree that we should only consider skills internally. 

Getting rid of sticky vehicles requires a deeper change of the logic (asserts, checks, partition, dicho, periodic, split...etc) which requires a more deeper refactoring. The current change is a first step towards this goal.

Considering the current time constraint, could we postpone this change ?

optimizer_wrapper.rb
```diff
@@ -363,78 +378,84 @@ def self.split_independent_vrp_by_skills(vrp)
...
+    skill_vehicle_ids = Hash.new{ [] }
+    grouped_vehicles.each{ |skills, vehicles|
+      skill_vehicle_ids[skills] += vehicles.map{ |vehicle| vrp.vehicles.find_index(vehicle) }
+    }
+
+    independant_skill_sets = compute_independent_skills_sets(vrp, mission_skills, vehicle_skills)
 
-    [vrp]
+    unused_vehicle_indices = (0..vrp.vehicles.size - 1).to_a
+    build_independent_vrps(vrp, independant_skill_sets, unused_vehicle_indices, skill_vehicle_ids, skill_service_ids)
```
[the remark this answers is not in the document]
braktar: Sorry @fonsecadeline pointed this and I sweep with the hand her remark. I was assuming incorrectly that build_independent_vrps was called multiple time with the same unused_vehicle_indices.

unused_vehicle_indices is a local variable

optimizer_wrapper.rb
```diff
@@ -363,78 +378,84 @@ def self.split_independent_vrp_by_skills(vrp)
...
-    # Cannot split if there is multimodal subtours
-    if !vrp.subtours&.any? && vrp.services.all?{ |s| s.sticky_vehicles.empty? }
-      return split_independent_vrp_by_skills(vrp)
-    end
+    # - there is a service with no skills (or sticky vehicle see: expand_data.rb:sticky_as_skills)
+    mission_skills = vrp.services.map(&:skills).uniq
+    return [vrp] if mission_skills.include?([])
+
+    # Generate Services data
+    grouped_services = vrp.services.group_by(&:skills)
```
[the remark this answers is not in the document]
braktar: services and vehicles skills are sorted within `filter_skills` method : https://github.com/Mapotempo/optimizer-api/blob/eef0e623275dd818f0e2fd33155c8e343eb1733a/lib/filters.rb#L220

models/vrp.rb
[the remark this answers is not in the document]
senhalil: Gwen, do you remember whether we have a ticket for this? Apparently, I put myself a reminder not to forget.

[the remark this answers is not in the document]
braktar: No, we have no ticket on this subject. Only TODO comments

PR #271, opened 2021-09-09
Improve performance of first/last_possible_days check

fonsecadeline:
Apparently there was a better way to do this. 

Corrects the performance problem we found : 
- for a given instance we used ~3 to 4 seconds to solve before implementing first/last_possible_days check
- we needed ~10 seconds after implementing it (which correspond to 16 seconds on beta server)

This edit allows to solve this VRP in ~ 5 seconds. We do increase computation time little but we are make lot of additional checks in order to respect new provided constraints so this is consistent for me.

### DIFF
--- a/lib/heuristics/periodic_heuristic.rb
+++ b/lib/heuristics/periodic_heuristic.rb
@@ -208,51 +208,25 @@ module Wrappers
 
     private
 
-    # Returns true if day can be used for service with corresponding `id`
-    # This is, regarding first and last_possible_days available combinations
-    def compatible_ranges_for_day(id, day)
-      (0..@services_data[id][:raw].visits_number - 1).collect{ |v_i|
-        next unless day.between?(@services_data[id][:raw].first_possible_days[v_i],
-                                 @services_data[id][:raw].last_possible_days[v_i])
-
-        [@services_data[id][:raw].first_possible_days[v_i], @services_data[id][:raw].last_possible_days[v_i]]
-      }.compact
-    end
-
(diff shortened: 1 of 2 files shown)


braktar (approved):

lib/heuristics/periodic_heuristic.rb
```diff
@@ -208,51 +208,28 @@ def check_solution_validity
...
+    # regarding first/last_possible_days available ranges : it is possible to find a
+    # pair of first/last_possible_day for each and every day used for this service_id
+    def day_in_possible_interval(service_id, day)
+      all_days = (@services_data[service_id][:used_days] + [day]).sort
+      seen_visits = 0
+      tried_indices = 0
+
+      return false if all_days.size > @services_data[service_id][:raw].visits_number
+
+      while tried_indices < @services_data[service_id][:raw].visits_number &&
```
senhalil, line 221: I suggest `times` instead of `while` then we can get rid of the `tried_indices` variable declaration, increment and bound check.
```ruby
@services_data[service_id][:raw].visits_number.times{ |tried_indices|
  next unless all_days[seen_visits].between?(@services_data[service_id][:raw].first_possible_days[tried_indices],
                                             @services_data[service_id][:raw].last_possible_days[tried_indices])

  seen_visits += 1 # This combination can accept this day

  break if seen_visits == all_days.size
}
```

senhalil (approved):
I have a small suggestion otherwise looks good to me

PR #272, opened 2021-09-10
Massive test fixes

braktar:


fonsecadeline (commented):
Since there is a pending to merge refactoring + edits on documentation (https://github.com/Mapotempo/optimizer-api/pull/268) there might be too many changes in periodic (arrays converted to sets). We should probably do this step by step. Are these minimal changes to make massive_tests work ? 

If we keep all those changes then you need to ensure there is no impact on initial data (for instance, through my _VALIDATE_WITH_INITIAL_ gitlab branch)

braktar:
Converting Array to Set doesn't seems to be really effective observation, I may remove it from this PR

fonsecadeline:
> Converting Array to Set doesn't seems to be really effective observation, I may remove it from this PR

Thanks! If you can do so that would help for refactoring + documentation integration :)

fonsecadeline (approved):

models/matrix.rb
```diff
@@ -19,6 +19,11 @@
 
 module Models
   class Matrix < Base
+    field :original_id
     fields :time, :distance, :value
+
+    def self.array_field_names
```
senhalil, line 25: Why "array", can't we do just `field_names`? Or `types` (like the type of the matrix -- `Models::Matrix.types`)?

test/wrapper_test.rb
```diff
@@ -2879,14 +2879,14 @@ def test_number_of_service_vrps_generated_in_split_independent
 
   def test_split_independent_vrps_with_useless_vehicle
     vrp = TestHelper.create(VRP.independent_skills)
-    vrp.vehicles << Models::Vehicle.create(id: 'useless_vehicle')
+    vrp.vehicles << Models::Vehicle.create(id: 'useless_vehicle', matrix_id: 'matrix_0')
```
senhalil, line 2882: How did this and the next test pass without this correction O_O 

Without the correction the new vehicle has no matrix, so it should lead to a matrix/router call but tests were passing. I am confused, any idea?

senhalil (commented):
I only reviewed the last commit, going to review all together now..

wrappers/ortools.rb
```diff
@@ -590,8 +590,11 @@ def parse_output(vrp, output)
           service = vrp.services.find{ |s| s.id == service_id }
           {
             original_service_id: service.original_id,
+            pickup_shipment_id: service.type == :pickup && service.original_id,
+            delivery_shipment_id: service.type == :delivery && service.original_id,
```
senhalil, line 594: Don't we need such a precision in vroom.rb?

senhalil (commented):
Decided to go commit par commit

util/job_manager.rb
```diff
@@ -83,6 +84,7 @@ def perform
         p[:result].first[:elapsed] = p[:graph].last[:time]
       end
       Result.set(self.uuid, p)
+      Models.delete_all
```
senhalil, line 87: Do we need this delete? Worker dies after completing the process, right?

lib/helper.rb
```diff
@@ -139,6 +139,25 @@ def self.deep_copy(original, override: {}, shallow_copy: [])
     end
   end
 
+  def self.populate_data(services_vrps)
```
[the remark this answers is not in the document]
braktar: The `Models::Relation` is never retrieved using its `id`, so it is not mandatory to load it back in memory.

The `split_independent_vrp` is applied on server side, so the worker receive an array of `service_vrp` which indeed share at least their `points`. So it cannot be performed from a single vrp at the moment

models/matrix.rb
braktar: `field_names` is a method provided by active_hash which returns all the fields of the Model. Here, I only expect the proper matrices fields without the new `original_id` field. `types` is not expressive to me, do you have some other suggestions ?

test/wrapper_test.rb
braktar: This vehicle was put in a sub problem without services :/ So it never tried to load the matrix. But now, as there is a matrix is the problem and we build the sub_problems with an update of the matrix links, it fails.

wrappers/ortools.rb
braktar: It already exists in vroom wrapper, it was only missing here in the unassigned jobs of ortools

util/job_manager.rb
braktar: Thanks for pointing it out, it is a residual line. I will remove it

util/job_manager.rb
```diff
@@ -39,7 +39,7 @@ def perform
       tick('Starting job') # Important to kill job before any code
 
       services_vrps = Marshal.load(Base64.decode64(self.options['services_vrps'])) # Get the vrp
-      populate_data(services_vrps)
```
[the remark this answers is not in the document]
braktar: In synchronous resolution, the vrp is loaded during the creation. The first steps of wrapper_vrp may use it.
Once the worker load the problem to solve it should load the Models.

The load performed in define process is only for repetitions

models/matrix.rb
senhalil: I can't think of a more expressive name than matrix "type" (like in time matrix, value matrix, etc, here time, distance, value is the type of the matrix). What about `matrix_field_names` -- i.e., the names of the fields which actually hold the matrices?

lib/tsp_helper.rb
```diff
@@ -36,7 +36,7 @@ def self.create_tsp(vrp, vehicle)
     }
 
     problem = {
-      matrices: vrp.matrices,
+      matrices: [Models::Matrix.find_by_id(vehicle.matrix_id)],
```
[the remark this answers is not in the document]
fab-girard: by using, `Models::X.find_methods` we could find/retrieve objects in memory which are not from "the VRP" used by the current request/worker. To avoid this problem, I suggest to always retrieve sub-models from parent model

lib/helper.rb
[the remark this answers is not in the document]
senhalil: Relations can be retrieved by their `id` (for example creating or soft-copying a service with an existing relation by id). This has never been done since it wasn't possible since now (and we were used the objects themselves) due to workers' active_hash was missing the id's. But now, that you are adding this functionality in the current PR, we can use `id`s for such task and it would be better to add it for all necessary models, right? Otherwise, we literally don't realise that there is such issue until the code is deployed (or launched via server/worker).

Ideally this load operation would be in base.rb and it would be recursive (i.e., we wouldn't need to load everything manually by hand), I guess, this is out of scope but if you think it is easy to do so, you can even do it more generally like that.

I forgot about the server side :facepalm: cheers. So effectively, doing it *per-VRP* is impossible due to shared objects.

lib/heuristics/dichotomious_approach.rb
```diff
@@ -149,9 +144,16 @@ def self.transfer_unused_vehicles(result, sub_service_vrps)
...
-        sv_one.points += sv_zero.points.select{ |p| p.id == vehicle.start_point_id || p.id == vehicle.end_point_id }
+        sv_one.points += [vehicle.start_point, vehicle.end_point].compact
       }
 
+      # Transfered vehicles requires to update the current matrix
+      matrix_indices = sv_one.points.map{ |point|
+        Models::Point.find_by_id(point.id).matrix_index
+      }.sort
+      SplitClustering.update_matrix_index(sv_one)
+      SplitClustering.update_matrix(sv_one.matrices.map(&:original_id).uniq, sv_one, matrix_indices)
```
senhalil, line 155: Can we skip this if there is no "transferred" vehicle?

test/real_cases_periodic_solver_test.rb
```diff
@@ -23,12 +23,15 @@ def test_periodic_and_ortools
       vrps = TestHelper.load_vrps(self)
 
       vrps.each{ |vrp|
+        Models.delete_all # needed to prevent duplicate ids
+        Helper.load_models([{vrp: vrp}])
```
senhalil, line 27: Brainstorming: I don't know if it is a good idea but what if we had a `delete_all` at the beginning of `load_models`? It looks like we always call them together. What do you think?

If we try to load a model over another one, we risk generating duplicate id error. I don't know if we would ever need to load two completely separate VRPs (i.e., no overlapping ids) one day but even so we can do so by giving both VRPs in an array like you did for job_manager.

models/matrix.rb
braktar: Thanks for the precision on why you find "type" more expressive, It seems good to me now :)

lib/tsp_helper.rb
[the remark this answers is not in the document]
senhalil: As far as I know, the workers do not share the active_hash objects (and its database) and each worker has a unique active_hash object (and its database). I suspect this might not be the case for server threads. We now make sure that there is only the current VRPs loaded in the worker memory by loading the object into the active_hash database as well. 

The risque I see is that if a function that uses a `find_by_id` logic is called via server, it might be problematic (for example, if someday a refacto moves some parse_output logic to server side but I don't know if we should worry about that).

lib/helper.rb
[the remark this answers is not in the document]
braktar: Do you expect to have loaded in memory all the sub objects (timewindows, capacities, quantities and locations) ?
By my side I only expect to have the top level objects. For the sub ones, I don't really see the goal of retrieving them from the database as we will always access them within the context of their parent.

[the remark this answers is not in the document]
senhalil: I think your separation top- vs sub-level is v good. We don't need the subs. My comment about recursive function was wrt ease of maintenance and automatism. 

If we, one day, change the internal model to the "csv" json type in the future (i.e., identical TW/capacity/quantity objects are not recreated to safe memory) then we might need the sub ones but I don't think we need to think about that anytime soon.

lib/tsp_helper.rb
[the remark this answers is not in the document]
fab-girard: Think about our API can be called without worker (synchronously: true), in this case memory is shared by every request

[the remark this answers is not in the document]
braktar: Currently each server and worker thread have its own active_hash database. For server threads, it seems to be the default behavior of nginx/passenger, while locally it requires a config file for puma.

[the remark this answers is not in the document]
senhalil: Recap of Bdx discussion: this is not an issue at the moment because due to active_hash, we are forced to keep the nginx workers separate -- i.e., each nginx worker treats a single request so they don't share memory. We said we shall talk during daily point what to do. 

I suggest keeping @braktar modifications since it is less typing and if we ever could treat multiple requests per worker, this means we have fixed active_hash issue, which means find_by_id's are also fixed (since they come from active_hash).

lib/heuristics/dichotomious_approach.rb
```diff
@@ -149,13 +144,22 @@ def self.transfer_unused_vehicles(result, sub_service_vrps)
...
       # Transfer unsued vehicle limit to the other side as well
       sv_zero_used_vehicle_count = result[:routes].count{ |r| r[:activities].any?{ |a| a[:service_id] } }
       sv_zero_unused_vehicle_limit = sv_zero.resolution_vehicle_limit - sv_zero_used_vehicle_count
       sv_one.resolution_vehicle_limit += sv_zero_unused_vehicle_limit
+
+      return if sv_zero_used_vehicle_count == 0
+
+      # Transfered vehicles requires to update the current matrix
+      matrix_indices = sv_one.points.map{ |point|
+        Models::Point.find_by_id(point.id).matrix_index
```
senhalil, line 159: this should work, right? Is there a specific reason to go via  Models::Point?
```suggestion
        point.matrix_index
```

optimizer_wrapper.rb
```diff
@@ -762,7 +763,7 @@ def self.parse_result(vrp, result)
         route = empty_route(vrp, vehicle)
         result[:routes] << route
       end
-      matrix = vrp.matrices.find{ |mat| mat.id == vehicle.matrix_id }
+      matrix = Models::Matrix.find_by_id(vehicle.matrix_id)
```
senhalil, line 766: In fact since we plan to have a proper "printer" for models, why don't we just add the matrix to the vehicles as a Models::Matrix so that we completely prevent these find_by logics (This might be out of scope of this dev, though, what do you think?)

braktar: My expectation is that this change has to be shipped with the introduction of the models printer

util/job_manager.rb
```diff
@@ -96,6 +98,19 @@ def on_killed
...
+      points.each{ |point| Models::Point.insert(point) }
+
+      rests = services_vrps.flat_map{ |s_v| s_v[:vrp].rests }.uniq
+      rests.each{ |rest| Models::Rest.insert(rest) }
+
+      services_vrps.each{ |s_v|
+        s_v[:vrp].vehicles.each{ |vehicle| Models::Vehicle.insert(vehicle) }
+        s_v[:vrp].services.each{ |service| Models::Service.insert(service) }
+      }
+    end
```
senhalil, line 113: It looks like we lost all the modifications you made for this function. If my memory serves me well there was  Relations and Units and the name was something like `load_models`. I don't remember if there was anything else, can you check?

util/job_manager.rb
senhalil: This delete came back to life as well :zombie:

util/job_manager.rb
senhalil: Oh there was the strategic delete_all. 

Also I remember this function being called in job_manager,rb (which is still here) and from wrapper_vrp.rb (maybe?) which doesn't exist anymore. Is that voluntary?

models/vrp.rb
```diff
@@ -334,15 +334,14 @@ def self.convert_relation_lapse_into_lapses(hash)
         if Models::Relation::ONE_LAPSE_TYPES.include?(relation[:type])
           if relation[:lapse]
             relation[:lapses] = [relation[:lapse]]
-            relation.delete(:lapse)
           end
         elsif [:vehicle_trips, Models::Relation::SEVERAL_LAPSE_TYPES].flatten.include?(relation[:type])
```
senhalil, line 338: Not introduced in this PR but `Models::Relation::SEVERAL_LAPSE_TYPES` already includes `vehicle_trips` so the if condition can be simplified to be `Models::Relation::SEVERAL_LAPSE_TYPES.include?(relation[:type])`

senhalil (changes requested):
Lost some of the modifications

util/job_manager.rb
braktar: Oh, As all the find_by_id were in one commit, i was assuming these changes were wrap together, it should be removed as well.

models/vrp.rb
braktar: It seems it already belong to `Models::Relation::SEVERAL_LAPSE_TYPES` :open_mouth:

senhalil: Yeah, it was a change introduced with a PR comment, apparently we missed to correct this. I believe there is one more such location where we have `[:vehicle_trips, Models::Relation::SEVERAL_LAPSE_TYPES].flatten`

util/job_manager.rb
senhalil: I think the modification we did (which are not related to find_by_id) were legit and we can keep them.

braktar: The delete_all/populate logic was called both synchronously and asynchronously if we have multiple repetitions.

Populate the database without using it, may only bring performance issues :/

senhalil: Then I fail to see why we need this populate logic now at all. All find_by_id logic is removed.. If we don't use the database, lets get rid of populate_data all together.

My thinking is that, if we want to introduce such a function we should do as a whole. If we populate only some objects and populate the database only on job_manager then a soft/hard-copy by id (or create by id relation) operation would work for the first repeat and fail on the second repeat if there is repetition. I don't think this is acceptable.

senhalil: I just saw that this (removing populate data all together) is what you did ;)

models/vrp.rb
```diff
@@ -335,15 +335,14 @@ def self.convert_relation_lapse_into_lapses(hash)
         if Models::Relation::ONE_LAPSE_TYPES.include?(relation[:type])
           if relation[:lapse]
             relation[:lapses] = [relation[:lapse]]
-            relation.delete(:lapse)
           end
-        elsif [:vehicle_trips, Models::Relation::SEVERAL_LAPSE_TYPES].flatten.include?(relation[:type])
+        elsif [Models::Relation::SEVERAL_LAPSE_TYPES].flatten.include?(relation[:type])
```
senhalil, line 339: We don't need the `[].flatten` part, right ?

senhalil (approved):

PR #273, opened 2021-09-15
Provide rest timewindows compatible with vehicle

fonsecadeline:


wrappers/ortools.rb
```diff
@@ -317,8 +317,10 @@ def solve(vrp, job, thread_proc = nil, &block)
           rests: vehicle.rests.collect{ |rest|
             OrtoolsVrp::Rest.new(
               time_windows: rest.timewindows.collect{ |tw|
```
braktar, line 319:
```suggestion
              time_windows: rest.timewindows.select{ |tw| tw.compatible_with(vehicle.timewindow, false) }.collect{ |tw|
                OrtoolsVrp::TimeWindow.new(start: tw.start, end: tw.end || 2**56)
              }
```

braktar (changes requested):

senhalil (commented):
Does or-tools only have a problem when a single rest TW doesn't overlap with vehicle.tw or when there is no feasible TW for the pause that overlaps with the vehicle TW? Where the segfault/error comes from inside or-tools or from our code?

tsp_simple.cc has the following code which should prevent such issues inside or-tools normally. if the error is in our cpp code, we can correct it there.

```cpp
    for (const TSPTWDataDT::Rest& rest : vehicle.Rests()) {
      IntervalVar* const rest_interval = solver->MakeFixedDurationIntervalVar(
          std::max(rest.ready_time[0], vehicle.time_start),
          std::min(rest.due_time[0], vehicle.time_end - rest.service_time),
          rest.service_time, // Currently only one timewindow
          false, absl::StrCat("Rest/", rest.rest_id, "/", vehicle_index));
      rest_array.push_back(rest_interval);
``` 

optimiser-ortool model should be able to handle multiple TWs for the rests (even if some not overlapping with the vehicle's TW) because as long as there is an overlapping rest TW, the model/vehicle is feasible. However, if there is no overlapping rest TW then the model/vehicle itself is infeasible and for such instances, this code would ignore all the TWs for the rest, if I am not mistaken. (Is that it?) 

Instead of ignoring, I suggest we refuse incoherent vehicles/instances (because if you ask for a 10-minute pause between 12 and 15 but if the vehicle works between 16 to 20 then something is not right, right?)

fonsecadeline:
> Does or-tools only have a problem when a single rest TW doesn't overlap with vehicle.tw or when there is no feasible TW for the pause that overlaps with the vehicle TW? Where the segfault/error comes from inside or-tools or from our code?
> 
> tsp_simple.cc has the following code which should prevent such issues inside or-tools normally. if the error is in our cpp code, we can correct it there.
> 
> ```c++
>   for (const TSPTWDataDT::Rest& rest : vehicle.Rests()) {
>     interval_vars.emplace_back(solver.MakeNumVar(
>         std::max(rest.ready_time[0], vehicle.time_start),
>         std::min(rest.due_time[0], vehicle.time_end - rest.service_time),
>         absl::StrCat("Rest/", rest.rest_id, "/", vehicle_index)));
>   }
> ```
> 
> optimiser-ortool model should be able to handle multiple TWs for the rests (even if some not overlapping with the vehicle's TW) because as long as there is an overlapping rest TW, the model/vehicle is feasible. However, if there is no overlapping rest TW then the model/vehicle itself is infeasible and for such instances, this code would ignore all the TWs for the rest, if I am not mistaken. (Is that it?)
> 
> Instead of ignoring, I suggest we refuse incoherent vehicles/instances (because if you ask for a 10-minute pause between 12 and 15 but if the vehicle works between 16 to 20 then something is not right, right?)

In this particular case, only 1/7 of rest's timewindows overlapped vehicle timewindows.  What we are trying to fix here is the fact that we return a solution with 1/1 unassigned if we provide all rests timewindows to ortools (segfault happened in the case where no rest timewindow was provided).

Your comment pointed out that in the case were none of rest's timewindows are compatible with vehicle ones, then  we might violated rest timewindows.

senhalil:
> In this particular case, only 1/7 of rest's timewindows overlapped vehicle timewindows. What we are trying to fix here is the fact that we return a solution with 1/1 unassigned if we provide all rests timewindows to ortools

I don't understand why this is happening. Or-tools should simply ignore the pause "intervals" (timewindows) that doesn't coincide with the vehicle, intriguing..

Hmm I just found it during digging further, I think. Optimizer-ortools implements the pause for one and only one TW, when there are multiple TW's all except the first are ignored. `// Currently only one timewindow` Because of this when the first TW is the "incoherent" TW, pause is irrealisable, which leads to all services to be unplanned.

So long story short, we don't actually support multiple TWs for pauses. Apparently, it is already not possible send multiple-TWs for pauses via mt-web, so we can add a validator which verifies that there is only one TW for pauses (and we can add a note on vrp_input.rb that currently only one TW is supported for pause).  

>  (segfault happened in the case where no rest timewindow was provided)

I think the segfault happens (when there is no tw for pause) because even if there is no TW, in the above code snippet I shared we call  `rest.ready_time[0]` and `rest.due_time[0]` which is not initialized (in https://github.com/Mapotempo/optimizer-ortools/blob/f79803e51ca31a65fbc2e73720b0f10879d00252/tsptw_data_dt.h#L951-L965 ) so when or-tools tries to use the reference, it crashes. We need something as follows before the `v->rests.emplace_back` call:

```cpp
      if (timewindows.empty()) {
        ready_time.push_back(0);
        due_time.push_back(CUSTOM_MAX_INT);
        tws_counter_ += 1;
      }
```

> Your comment pointed out that in the case were none of rest's timewindows are compatible with vehicle ones, then we might violated rest timewindows.

Lets add a check for this case at the validation stage so that we don't accept incoherent pause/vehicle definitions. What do you think?

fonsecadeline:
> > In this particular case, only 1/7 of rest's timewindows overlapped vehicle timewindows. What we are trying to fix here is the fact that we return a solution with 1/1 unassigned if we provide all rests timewindows to ortools
> 
> I don't understand why this is happening. Or-tools should simply ignore the pause "intervals" (timewindows) that doesn't coincide with the vehicle, intriguing..
> 
> Hmm I just found it during digging further, I think. Optimizer-ortools implements the pause for one and only one TW, when there are multiple TW's all except the first are ignored. `// Currently only one timewindow` Because of this when the first TW is the "incoherent" TW, pause is irrealisable, which leads to all services to be unplanned.
> 
> So long story short, we don't actually support multiple TWs for pauses. Apparently, it is already not possible send multiple-TWs for pauses via mt-web, so we can add a validator which verifies that there is only one TW for pauses (and we can add a note on vrp_input.rb that currently only one TW is supported for pause).
> 
> > (segfault happened in the case where no rest timewindow was provided)
> 
> I think the segfault happens (when there is no tw for pause) because even if there is no TW, in the above code snippet I shared we call `rest.ready_time[0]` and `rest.due_time[0]` which is not initialized (in https://github.com/Mapotempo/optimizer-ortools/blob/f79803e51ca31a65fbc2e73720b0f10879d00252/tsptw_data_dt.h#L951-L965 ) so when or-tools tries to use the reference, it crashes. We need something as follows before the `v->rests.emplace_back` call:
> 
> ```c++
>       if (timewindows.empty()) {
>         ready_time.push_back(0);
>         due_time.push_back(CUSTOM_MAX_INT);
>         tws_counter_ += 1;
>       }
> ```
> 
> > Your comment pointed out that in the case were none of rest's timewindows are compatible with vehicle ones, then we might violated rest timewindows.
> 
> Lets add a check for this case at the validation stage so that we don't accept incoherent pause/vehicle definitions. What do you think?

I will rework this PR with these comments. We do need a validation, but we should also edit periodic_visits.rb in order to only provide one timewindow for every generated rest. In other words, my edit in ortools.rb should rather be in periodic_visits.rb.

test/models/timewindow_test.rb
```diff
@@ -0,0 +1,46 @@
...
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with Mapotempo. If not, see:
+# <http://www.gnu.org/licenses/agpl.html>
+#
+require './test/test_helper'
+
+module Models
+  class TimewindowTest < Minitest::Test
+    include Rack::Test::Methods
```
braktar, line 22: Is this include mandatory ?

wrappers/ortools.rb
```diff
@@ -315,9 +315,11 @@ def solve(vrp, job, thread_proc = nil, &block)
...
           rests: vehicle.rests.collect{ |rest|
             OrtoolsVrp::Rest.new(
-              time_windows: rest.timewindows.collect{ |tw|
-                OrtoolsVrp::TimeWindow.new(start: tw.start, end: tw.end || 2**56)
-              },
+              # Ortools can only deal with one and exactly one timewindow per rest :
+              time_windows:
+                [(rest.timewindows.collect{ |tw|
+                  OrtoolsVrp::TimeWindow.new(start: tw.start, end: tw.end || 2**56)
+                } + [OrtoolsVrp::TimeWindow.new(start: 0, end: 2**56)])[0]], # write test: avec, sans pause ça doit reprendre bon truc dans run_ortools
```
braktar, line 322:
```suggestion
                rest.timewindows.any? ?
                  (rest.timewindows.collect{ |tw|
                    OrtoolsVrp::TimeWindow.new(start: tw.start, end: tw.end || 2**56)
                   } : [OrtoolsVrp::TimeWindow.new(start: 0, end: 2**56)], # Rests should always have a timewindow
```

lib/interpreters/periodic_visits.rb
```diff
@@ -64,11 +64,15 @@ def expand(vrp, job, &block)
       vrp
     end
 
-    def generate_timewindows(timewindows_set)
+    def generate_timewindows(timewindows_set, only_day_index = nil)
```
braktar, line 67: From my point of view the prefix `only_` assumes that the variable is a boolean
```suggestion
    def generate_timewindows(timewindows_set, day_index = nil)
```

braktar (changes requested):

test/models/timewindow_test.rb
fonsecadeline: It is in most of our test files but it does not seem important in this one. I will remove it.

wrappers/ortools.rb
```diff
@@ -315,9 +315,12 @@ def solve(vrp, job, thread_proc = nil, &block)
...
             OrtoolsVrp::Rest.new(
-              time_windows: rest.timewindows.collect{ |tw|
-                OrtoolsVrp::TimeWindow.new(start: tw.start, end: tw.end || 2**56)
-              },
+              # Ortools can only deal with one and exactly one timewindow per rest :
+              time_windows:
+                rest.timewindows.any? ?
+                  (rest.timewindows.collect{ |tw|
+                    OrtoolsVrp::TimeWindow.new(start: tw.start, end: tw.end || 2**56)
+                   } : [OrtoolsVrp::TimeWindow.new(start: 0, end: 2**56)], # Rests should always have a timewindow
```
braktar, line 323: I didn't removed the first parenthesis in my previous suggestion 
```suggestion
                  rest.timewindows.collect{ |tw|
                    OrtoolsVrp::TimeWindow.new(start: tw.start, end: tw.end || 2**56)
                  } : [OrtoolsVrp::TimeWindow.new(start: 0, end: 2**56)], # Rests should always have a timewindow
```

braktar (changes requested):

braktar (approved):

fonsecadeline:
Test cassé est un des tests aléatoires corrigé par https://github.com/Mapotempo/optimizer-ortools/pull/27 : SplitClusteringTest#test_avoid_capacities_overlap

senhalil:
> Test cassé est un des tests aléatoires corrigé par [Mapotempo/optimizer-ortools#27](https://github.com/Mapotempo/optimizer-ortools/pull/27) SplitClusteringTest#test_avoid_capacities_overlap

I don't see how [Mapotempo/optimizer-ortools#27](https://github.com/Mapotempo/optimizer-ortools/pull/27) can fix a flaky optim-api test. Copy/paste typo? Did you mean https://github.com/Mapotempo/optimizer-api/pull/282?

fonsecadeline:
> > Test cassé est un des tests aléatoires corrigé par [Mapotempo/optimizer-ortools#27](https://github.com/Mapotempo/optimizer-ortools/pull/27) SplitClusteringTest#test_avoid_capacities_overlap
> 
> I don't see how [Mapotempo/optimizer-ortools#27](https://github.com/Mapotempo/optimizer-ortools/pull/27) can fix a flaky optim-api test. Copy/paste typo? Did you mean #282?

Maybe I mixed PRs, you pushed a list of tests that break randomly. I figured that was the PR that fixed it.

senhalil:
> Maybe I mixed PRs, you pushed a list of tests that break randomly. I figured that was the PR that fixed it.

Yeah, the list is introduced in https://github.com/Mapotempo/optimizer-api/pull/282/commits/3941e44b5299a28a96433fec4c0e2ff628bb8a29. optim-ortools#27 is not in this repo

models/concerns/validate_data.rb
```diff
@@ -107,6 +108,24 @@ def check_matrices
...
+
+    used_rest_ids.uniq.each{ |rest_id|
+      corresponding = @hash[:rests].find{ |r| r[:id] == rest_id }
+      unless corresponding
+        raise OptimizerWrapper::DiscordantProblemError.new(
+          "Unknown rest provided to at least one vehicle : #{rest_id}"
+        )
+      end
+
+      next unless corresponding[:timewindows].to_a.size > 1
```
senhalil, line 123: This is handled by active_hash (AH) during object creation automatically and AH sends the indices for finding the offending parameter easily in a uniform manner (with an IDerror if I remember correctly). 
```suggestion
      next unless corresponding && corresponding[:timewindows].to_a.size > 1
```

models/concerns/validate_data.rb
```diff
@@ -107,6 +108,24 @@ def check_matrices
...
+
+      raise OptimizerWrapper::UnsupportedProblemError.new('Rests can only have one timewindow')
```
senhalil, line 125: Can you add a similar note to the definition of `timewindows` for `:vrp_request_rest` in vrp_input.rb?

'Time slot while the rest may begin. At most one timewindow per rest is supported.` or something similar

wrappers/ortools.rb
```diff
@@ -315,9 +315,12 @@ def solve(vrp, job, thread_proc = nil, &block)
...
           rests: vehicle.rests.collect{ |rest|
             OrtoolsVrp::Rest.new(
-              time_windows: rest.timewindows.collect{ |tw|
-                OrtoolsVrp::TimeWindow.new(start: tw.start, end: tw.end || 2**56)
-              },
+              # Ortools can only deal with one and exactly one timewindow per rest :
+              time_windows:
+                rest.timewindows.any? ?
+                  rest.timewindows.collect{ |tw|
+                    OrtoolsVrp::TimeWindow.new(start: tw.start, end: tw.end || 2**56)
```
senhalil, line 322: If the API lets in only one timewindow for rests, can we have a check here (to make sure that we didn't do anything incoherent in the code) and verify that `rest.timewindows.size <= 1` and raise otherwise. If not, the collect will happily send optim-ortools multiple timewindows and optim-ortools will just ignore them.

wrappers/ortools.rb
```diff
@@ -315,9 +315,12 @@ def solve(vrp, job, thread_proc = nil, &block)
...
+                  } : [OrtoolsVrp::TimeWindow.new(start: 0, end: 2**56)], # Rests should always have a timewindow
```
senhalil, line 323: I still think that since the segfault is happening because of an error in optim-ortools ("de-referencing an uninitialized variable") and that there is an easy fix, we should fix it in optim-ortools. But I note this for myself, we can have this in the mean time.

senhalil (changes requested):
I think doc needs clarification and a check (and its test) need modification.

test/wrappers/ortools_test.rb
```diff
@@ -5368,4 +5368,28 @@ def test_relations_sent_to_ortools_when_different_lapses
       end
     }
   end
+
+  def test_always_one_time_window_provided_for_rests
+    problem = VRP.basic
+    problem[:rests] = [{ id: 'rest', duration: 10 }]
+    problem[:vehicles][0][:rest_ids] = ['rest']
```
senhalil, line 5375: This instance doesn't lead to a segfault (even without the fix introduced in the PR) because the pause is eliminated by the simplification. 

Adding a TW to one services leads to the segfault (the rest of the test would need a correction accordingly): 
```suggestion
    problem = VRP.basic
    problem[:rests] = [{ id: 'rest', duration: 10 }]
    problem[:vehicles][0][:rest_ids] = ['rest']
    problem[:services][0][:activity][:timewindows] = [{ start: 0, end: 20 }] # so that rest is not simplified
```

test/wrappers/ortools_test.rb
```diff
@@ -5368,4 +5368,28 @@ def test_relations_sent_to_ortools_when_different_lapses
...
+
+    [[], [{ start: 2, end: 3 }]].each{ |rest_tws|
+      OptimizerWrapper.config[:services][:ortools].stub(
+        :run_ortools,
+        lambda { |pb, _, _|
+          assert_equal 1, pb.vehicles.first.rests.first.time_windows.size
```
senhalil, line 5381: I suspect that there is a bug in the new TW collection logic. 

Because for the first `[]` case, the vehicle should have no rests since the `simplify_vehicle_pause` simplification is active (that's what I see when I test on local with this instance (without the modification I suggested above), the pause is really eliminated). So I don't quite understand how this test passes, any idea?

senhalil (commented):
There might be an issue with the

test/wrappers/ortools_test.rb
senhalil: Oh I just found it, I think, I called `wrapper_vrp` and this test calls `solve` so it skips the eliminations. However, because of that we basically test with a state that cannot happen in reality

lib/interpreters/periodic_visits.rb
```diff
@@ -212,12 +215,18 @@ def generate_vehicles(vrp)
...
-            new_vehicle
+            build_vehicle(vrp, vehicle, vehicle_day_index, nil, rests_durations)
           else
-            timewindows.select{ |timewindow| timewindow.day_index.nil? || timewindow.day_index == vehicle_day_index % 7 }.collect{ |associated_timewindow|
-              new_vehicle = build_vehicle(vrp, vehicle, vehicle_day_index, rests_durations)
-              new_vehicle.timewindow = Models::Timewindow.create(start: associated_timewindow.start || 0, end: associated_timewindow.end || 86400)
+            timewindows.select{ |timewindow|
+              timewindow.day_index.nil? || timewindow.day_index == vehicle_day_index % 7
+            }.collect{ |associated_timewindow|
+              # FIXME : this function assumes there are no two timewindows with same day_index.
```
senhalil, line 223: @braktar I thought you already fixed this issue weeks ago, I must have misunderstood. Doesn't this issue of "having two vehicles with the same id" lead to an internal problem/bug ? If so, maybe we should just add a verification to the API, no?

senhalil (approved):

lib/interpreters/periodic_visits.rb
braktar: Oh ! Yes ! It comes in another PR (https://github.com/braktar/optimizer-api/pull/14/files) which requires : https://github.com/Mapotempo/optimizer-api/pull/353

braktar: Let's keep this one (#273) for the end. First we need #353 and then [https://github.com/braktar/optimizer-api/pull/14/](https://github.com/braktar/optimizer-api/pull/14

senhalil: - [ x ] https://github.com/Mapotempo/optimizer-api/pull/353 is merged

senhalil: You can rebase & move 14 to this side

braktar: Yes ! The FIXME has no sense with the vehicle name fix

senhalil: ping to remind rebase and the FIXME fix :)  then we can merge this.

senhalil (approved):
@braktar is there any behavior changes here? Do we need to to launch the benchmarks of scheduling? If not, :rocket:

braktar:
There is 2 fixes in context of scheduling : Rests now have a correct timewindow according to the vehicle. And we now allow multiple rests with the same day index.

PR #275, opened 2021-09-17
Move zip_condition for consistency check

braktar:


fonsecadeline (approved):

senhalil (approved):
Is it normal that the total time is not correct when there is zipping? Unzip should correct the totals (travel, route, solution, etc) while "unzipping" ideally. Do we have a ticket for this?

PR #276, opened 2021-09-17
Try car & car_interurban to prevent RouterError

senhalil:
In clustering we need depot-point et point-depot distances and to prevent long matrix calculation we use `:car` as the router_mode by default. However, `:car` doesn't cover every location Mapotempo supports so if router returns OutOfSupportedAreaOrNotSupportedDimensionErr, we switch to car_interurban to make sure to cover all supported locations.

 Closes optimizer-api/-/issues/868


fonsecadeline (approved):

fab-girard:
Router is going to provide a `car` mode everywhere it's possible

PR #277, opened 2021-09-20
Edit cases where vehicles skills are initialized with empty skills

fonsecadeline:
VRPs sent by mapotempo-web send [] as default vehicle skills.

We used to keep this value in whole code.

When we convert sticky_vehicles into skills this produces a problem : https://gitlab.com/mapotempo/optimizer-api/-/issues/869

### DIFF
--- a/models/vehicle.rb
+++ b/models/vehicle.rb
@@ -110,7 +110,7 @@ module Models
         hash[:unavailable_days].delete_if{ |index| !work_day_indices.include?(index.modulo(7)) }
       end
 
-      hash[:skills] = [[]] unless hash.has_key?(:skills) # If vehicle has no skills, it has the empty skillset
+      hash[:skills] = [[]] if hash[:skills].to_a.empty? # If vehicle has no skills, it has the empty skillset
 
       super(hash)
     end
--- a/models/vrp.rb
+++ b/models/vrp.rb
@@ -237,7 +237,7 @@ module Models
       vrp.add_sticky_vehicle_if_routes_and_partitions
       vrp.expand_unavailable_days
       vrp.provide_original_info
-      vrp.sticky_as_skills
+      vrp.sticky_as_skills # TODO: this should be done on hash in order to completely remove sticky_vehicles from service model
     end
 
     def self.convert_position_relations(hash)
--- a/test/api/v01/vrp_test.rb
+++ b/test/api/v01/vrp_test.rb
@@ -68,6 +68,35 @@ class Api::V01::VrpTest < Minitest::Test
     end
   end
 
+  def test_generated_vehicle_skills
+    problem = VRP.basic
+    [nil, [], [[]],
+     'skill1', [['skill1']],
+     'skill1,skill2', [['skill1', 'skill2']]
+    ].each_with_index{ |skill_set, v_i|
+      problem[:vehicles] << problem[:vehicles].first.dup unless v_i == 0
+      problem[:vehicles][v_i][:id] = "vehicle_#{v_i}"
+      problem[:vehicles][v_i][:skills] = skill_set
+    }
+    problem[:services].first[:skills] = ['skill1', 'skill2'] # otherwise skills will be ignored in vehicles
+
+    OptimizerWrapper.stub(
+      :define_main_process,
+      lambda { |services_vrps, _job|
+        vrp = services_vrps[0][:vrp]
+        [[[]], [[]], [[]],
+         [[:skill1]], [[:skill1]],
+         [[:skill1, :skill2]], [[:skill1, :skill2]]
+        ].each_with_index{ |expected_skills, v_i|
+          assert_equal vrp.vehicles[v_i].skills, expected_skills
+        }
+        {}
+      }
+    ) do
+      submit_vrp api_key: 'demo', vrp: problem
+    end
+  end
+
   def test_exceed_params_limit
     vrp = VRP.toy
     vrp[:points] *= 151


models/vrp.rb
```diff
@@ -237,7 +237,7 @@ def self.expand_data(vrp)
       vrp.add_sticky_vehicle_if_routes_and_partitions
       vrp.expand_unavailable_days
       vrp.provide_original_info
-      vrp.sticky_as_skills
+      vrp.sticky_as_skills # TODO : this should be done on hash in order to completemly remove sticky_vehicles from service model
```
braktar, line 240:
```suggestion
      vrp.sticky_as_skills # TODO: this should be done on hash in order to completely remove sticky_vehicles from service model
```

braktar (changes requested):

braktar (approved):

models/vehicle.rb
```diff
@@ -110,7 +110,7 @@ def self.create(hash)
         hash[:unavailable_days].delete_if{ |index| !work_day_indices.include?(index.modulo(7)) }
       end
 
-      hash[:skills] = [[]] unless hash.has_key?(:skills) # If vehicle has no skills, it has the empty skillset
+      hash[:skills] = [[]] if !hash.key?(:skills) || hash[:skills] == [] # If vehicle has no skills, it has the empty skillset
```
fonsecadeline, line 113:
```suggestion
      hash[:skills] = [[]] if hash[:skills].to_a.empty? # If vehicle has no skills, it has the empty skillset
```

api/v01/entities/vrp_input.rb
```diff
@@ -516,11 +516,15 @@ module VrpVehicles
     optional(:distance, type: Integer, desc: 'Maximum tour distance. Not available with periodic heuristic.')
     optional(:maximum_ride_time, type: Integer, desc: 'Maximum ride duration between two route activities')
     optional(:maximum_ride_distance, type: Integer, desc: 'Maximum ride distance between two route activities')
-    optional :skills, type: Array[Array[Symbol]],
+    optional :skills, type: Array[Array[Symbol]], default: [[]],
```
braktar, line 519: Using default values within the active_hash model used to generate array with the same id for all objets of the problem. Could check if the object_id of skills obtained using this default value by two vehicles is different ?

fonsecadeline: Now skills.object_id are different. Do you think it will be too costly to duplicate these ? In my opinion it is cleaner like this.

test/test_helper.rb
```diff
@@ -80,6 +80,9 @@ def self.coerce(vrp)
 
     vrp[:vehicles].each{ |vehicle|
       vehicle.delete(:original_id)
+      next unless vehicle[:skills].to_a.empty? # nil or []
+
+      vehicle[:skills] = [[]]
```
senhalil, line 85: This doesn't correct the case if `vehicle[:skills]` is `["skill1", "skill2"]`.


We already have the correct fix in https://github.com/Mapotempo/optimizer-api/pull/274
you can cherry-pick the commit https://github.com/Mapotempo/optimizer-api/pull/274/commits/de47b5b00724d158bfa26be1452a98c39ebbe591 that way we wouldn't have a conflict.

models/vehicle.rb
```diff
@@ -110,8 +110,6 @@ def self.create(hash)
         hash[:unavailable_days].delete_if{ |index| !work_day_indices.include?(index.modulo(7)) }
       end
 
-      hash[:skills] = [[]] unless hash.has_key?(:skills) # If vehicle has no skills, it has the empty skillset
```
senhalil, line 113: This is needed. Otherwise we will have a single `[[]]` object if I am not mistaken.

api/v01/entities/vrp_input.rb
```diff
@@ -516,11 +516,15 @@ module VrpVehicles
...
-                        val.is_a?(String) ?
-                        [val.split(/,/).map!(&:strip).map!(&:to_sym)] :
-                        val&.map{ |set| set.map(&:to_sym) }
+                        if val.is_a?(String)
+                          [val.split(/,/).map!(&:strip).map!(&:to_sym)]
+                        elsif val == [] # Mapotempo-Web case
+                          [[]]
+                        else
+                          val&.map{ |set| set.map(&:to_sym) }
+                        end
```
senhalil, line 527: coerce_with is called for each "element" of the array, so returning `[[]]` is not correct.

models/vehicle.rb
fonsecadeline: So you mean having a default value in vrp_input generates one single object ? We do have a different object_id when we use default value in vrp_input. Is that not enough? That would explain the problem I mentionned on test_using_two_solver

api/v01/entities/vrp_input.rb
senhalil: If mapotempo-web is sending `[]` then this coerce_with is not called (because the array is empty -- i.e., there is no `val` to call the coerce with).

fonsecadeline: When mapotempo-web sends [ ] then val is [] and we do call coerce on [], otherwise I guess test_generated_vehicle_skills would fail.

models/vehicle.rb
senhalil: I just saw @braktar's comment, it looks like grape fixed its single default object bug. So we can transfer the default to the grape you can ignore the above comment. 

However,  to fix mt-web case (i.e., `[]`) we need a  `hash[:skills] = [[]] if hash[:skills].empty?`  "conversion" somewhere. Before "everything" (i.e., before consistency and retrocomp checks) might be the best place.

api/v01/entities/vrp_input.rb
senhalil: Sorry, I just verified that this is the case with a byebug, I mixed "values" logic with the "coerce_with" logic -- one is called for each value of the array, the other is called for the value as whole :facepalm:

models/vehicle.rb
fonsecadeline: Is it okay for you as is i.e. in coerce_with ?

api/v01/entities/vrp_input.rb
senhalil: What would this coerce do for `["skill1", "skill2"]` instead of `[["skill1", "skill2"]]` (I don't know if mt-web can send such skills)?

models/vehicle.rb
senhalil: That should be enough. We want not just the outer but the inner array to have different object_ids but if `hash[:vehicles][x][:skills].object_id` is different for different x's then `hash[:vehicles][x][:skills][0].object_id` would be different to. So I don't know.

api/v01/entities/vrp_input.rb
fonsecadeline: I don't know either, @braktar do you know ?
Since that wasn't in vehicle.rb create I just ignored this case, assuming it was already managed.

braktar: In mt-web, vehicle skills are transmitted the following way : `data[:use_skills] ? [vehicle[:skills]] : nil`

senhalil: I am not so sure that it is managed. I think, if it is okay for you, we can fix this coerce_with so that it handles all cases correctly -- so that we are sure that only correct vehicle[:skills] enters the API and all coerce logic is in one place.

senhalil: Thanks @braktar, I don't see how they can send us `[]` in that case? When the vehicle has no skills, they should be sending `[[]]` or `[nil]` (a la marge).

braktar (approved):

PR #283, opened 2021-09-23
Split solve respects vehicle relations inc. trips

senhalil:
Fixes a sentry of a mt-web user.

Closes optimizer-api/-/issues/852

### DIFF
--- a/lib/interpreters/split_clustering.rb
+++ b/lib/interpreters/split_clustering.rb
@@ -238,6 +238,8 @@ module Interpreters
         unless sides.size == 2 && sides.none?(&:empty?)
           # this might happen under certain cases (skills etc can force all points to be on one side)
           # and not necessarily a problem but it should happen very rarely (in real instances)
+          # it might also happen because the vehicles are in a linking relation (like vehicle_trips)
+          # which forces them to be stay on one side.
           log 'There should be exactly two clusters in split_solve_core!', level: :warn
           ss_data[:cannot_split_further] = true
         end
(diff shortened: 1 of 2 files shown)


lib/interpreters/split_clustering.rb
```diff
@@ -386,7 +389,13 @@ def self.create_representative_vrp(split_solve_data)
...
       split_solve_data[:original_vrp].relations.select{ |r| FORCING_RELATIONS.include?(r.type) }.each{ |relation|
         if relation.linked_vehicle_ids.any? && relation.linked_services.none?
-          relations << { type: :same_route, linked_ids: relation.linked_vehicle_ids }
+          linked_ids = []
+          relation.linked_vehicle_ids.map{ |v_id|
+            s_id = "0_representative_vrp_s_#{v_id}"
+            linked_ids << s_id if services.any?{ |s| s[:id] == s_id }
+          }
+
+          relations << { type: :same_route, linked_ids: linked_ids } if linked_ids.size > 1
```
fonsecadeline, line 398: I am not sure I understand how this would work : how do we link those ids with corresponding vehicles ?

senhalil: It is done by `s_id = "0_representative_vrp_s_#{v_id}"` because in the representative vrp, there are ("representative" read "fake") services with the vehicle id in their name -- i.e., `"0_representative_vrp_s_#{v_id}"` is the id of the service representing the services assigned* to the vehicle `v_id`.  

*: assigned by the very first split so that we can continue the split logic by splitting "vehicles" to half instead of services.

If some real vehicles (vehicles from the original vrp) are in a forcing relation, their ids appear in  `relation.linked_vehicle_ids` and  if there are services assigned to these vehicles during the very first split, then there will be representative services representing these vehicle in the representative vrp and the original forcing vehicle relation will be represented with a `:same_route` relation on the services representing these vehicles.

(if in the representative vrp, there are no services representing at least two of such vehicles `linked_ids.size > 1` then we don't need to represent the original relation in the representative vrp because the vehicles will not be in the following splits. These relations will continue to be present in the sub_vrps.)

fonsecadeline: I did not remember we already had "0_representative_vrp_s_#{v_id}" logic that is why I got lost. Thank you for the explanation.

fonsecadeline (approved):

lib/interpreters/split_clustering.rb
senhalil: No worries, the logic is a bit hidden in fact (I am not v happy with it but it is quite faster and more reliable like this). Maybe a future refacto will make the hidden logic more obvious.

braktar (approved):

PR #284, opened 2021-09-24
Return original_vehicle_id as route.id

fonsecadeline:


senhalil:
Is this to ensure that csv is importable or for some other reason? If it is for mt-web, can you add a simple test? If it is not too much hassle a unit test calling `build_csv_activity` and checking that we have  `route[:original_vehicle_id]`

senhalil (approved):
A test would be useful (if this change is due to mt-web import) otherwise it is :rocket: for me

fonsecadeline:
> Is this to ensure that csv is importable or for some other reason? If it is for mt-web, can you add a simple test? If it is not too much hassle a unit test calling `build_csv_activity` and checking that we have `route[:original_vehicle_id]`

See my email from 24/09 with object "Conclusions debug import sur mapotempo-web" for more details.

senhalil:
I just saw the email and was here to write that :grinning:

lib/output_helper.rb
```diff
@@ -161,7 +161,7 @@ def self.find_type_and_complete_id(activity)
     def self.build_csv_activity(name, route, activity)
       type, complete_id = find_type_and_complete_id(activity)
       [
-        route && route[:vehicle_id],
+        route && route[:original_vehicle_id],
```
senhalil, line 164: @fab-girard asked if you can add a comment like the following so that we can go back when mt-web supports the first version you implemented 
`# TODO: Can be reverted when mt-web supports importing multiple plans with multiple route references`

fonsecadeline, line 164:
```suggestion
        route && route[:original_vehicle_id], # TODO: Can be reverted when mt-web supports importing multiple plans with multiple route references
```

PR #287, opened 2021-09-28
Ability to define a specific router api key

fab-girard:


braktar (approved):

optimizer_wrapper.rb
```diff
@@ -15,11 +15,16 @@
 # along with Mapotempo. If not, see:
 # <http://www.gnu.org/licenses/agpl.html>
 #
+# frozen_string_literal: true
+
```
senhalil, line 19: Is this voluntary? We modify string variables so I don't know if freezing them can cause an issue. If there is no risk we need this for all files. Shall we do all now?

optimizer_wrapper.rb
```diff
@@ -81,7 +63,7 @@ def self.wrapper_vrp(api_key, services, vrp, checksum, job_id = nil)
 
     services_vrps = split_independent_vrp(vrp).map{ |vrp_element|
       {
-        service: services[:services][:vrp].find{ |s|
```
senhalil, line 84: Thanks for this rename!

test/api/v01/api_test.rb
```diff
@@ -16,9 +16,11 @@
 # <http://www.gnu.org/licenses/agpl.html>
 #
 require './test/test_helper'
+require './test/api/v01/helpers/count_helper'
 
 class Api::V01::ApiTest < Minitest::Test
   include Rack::Test::Methods
+  include TestHelper
```
senhalil, line 23: When do we need this `include TestHelper`? It exists only in 3 test files and I suspect it might not be necessary since we `require './test/test_helper'`?

test/test_helper.rb
```diff
@@ -70,6 +62,16 @@
 
 require 'zlib'
 
+ENV['APP_ENV'] ||= 'test'
+require File.expand_path('../../config/environments/' + ENV['APP_ENV'], __FILE__)
+Dir[File.dirname(__FILE__) + '/../config/initializers/*.rb'].sort.each{ |file| require file }
+
+require './util/error.rb'
+require './util/config.rb'
+Dir[File.dirname(__FILE__) + '/../models/*.rb'].sort.each{ |file| require file }
```
senhalil, line 71: It looks like these `Dir[...]` and `File.expand_path` can be simplified by `require_all` https://github.com/jarmo/require_all#usage

In fact this is the only functionality of `require_all` gem. We had a discussion before about this and we were both confused by `require_all` not requiring everything "automatically" but I just read their document and it doesn't do auto loading, all it does is to simplify writing the above complex require logic in a simpler fashion. 

(I noticed this a while back I didn't have time to fix these requires, do you want to attack it in this PR if not we can create a quality ticket?)

models/concerns/distance_matrix.rb
```diff
@@ -63,7 +63,7 @@ def compute_need_matrix(&block)
         # set matrix_time and matrix_distance depending of dimensions order
         log "matrix computation #{matrix_points.size}x#{matrix_points.size}"
         tic = Time.now
-        router_matrices = OptimizerWrapper.router.matrix(OptimizerWrapper.config[:router][:url], mode, dimensions, matrix_points, matrix_points, options)
+        router_matrices = router.matrix(OptimizerWrapper.config[:router][:url], mode, dimensions, matrix_points, matrix_points, options)
```
senhalil, line 66: Why don't we set the url inside the router object like we do for the `api_key` when we are creating the router object -- something like 
```
OptimizerWrapper.router(OptimizerWrapper.config[:router][:api_key], OptimizerWrapper.config[:router][:url])
```
instead of
```
OptimizerWrapper.router(OptimizerWrapper.config[:router][:api_key])
```

That way we wouldn't have to do `OptimizerWrapper.config[:router][:url]` everywhere. What do you think?

senhalil (changes requested):
I believe there is a bug, the rest are comments.

optimizer_wrapper.rb
fab-girard: no risk, should be added in all files, we have a code/task to add this un all file, but it's not available in optim-api

lib/interpreters/multi_modal.rb
```diff
@@ -31,7 +31,7 @@ def initialize(vrp, selected_service)
     def generate_isolines
       @original_vrp.subtours.collect{ |tour|
         tour.transmodal_stops.collect{ |stop|
-          isoline = JSON.parse(OptimizerWrapper.router.isoline(OptimizerWrapper.config[:router][:url], tour.router_mode, (tour.time_bounds ? :time : :distance), stop.location.lat, stop.location.lon, (tour.time_bounds || tour.distance_bounds)))
+          isoline = JSON.parse(vrp.router.isoline(OptimizerWrapper.config[:router][:url], tour.router_mode, (tour.time_bounds ? :time : :distance), stop.location.lat, stop.location.lon, (tour.time_bounds || tour.distance_bounds)))
```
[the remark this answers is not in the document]
fab-girard: this code is not covered by any test in this case, it's a problem, we need to follow coverage of the project. I'll fix it

test/test_helper.rb
fab-girard: I suggest to change that in another PR (issue to create), I've juste moved some code and add a line in this one

models/concerns/distance_matrix.rb
fab-girard: the router_wrapper.rb file is the same for 3 projects (it should be moved one day in a gem), the url cannot be different for the moment in none of these 3 projects

test/api/v01/api_test.rb
fab-girard: I think this include is not necessary, I had it because there was the same include in `vrp_test.rb`, but it should be removed according to me, I repush

lib/interpreters/multi_modal.rb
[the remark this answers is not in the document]
fab-girard: i've pushed the fix

test/test_helper.rb
fab-girard: this include is mandatory

test/api/v01/api_test.rb
fab-girard: this include is mandatory

test/test_helper.rb
fab-girard: @senhalil you create the issue? I need to merge this PR

lib/interpreters/multi_modal.rb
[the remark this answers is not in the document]
senhalil: I believe multi-modal is written as a poc by @braktar and I am not sure if it is well tested ;)

test/api/v01/api_test.rb
senhalil: Hmm, it is weird that few test files needed while most of it can do without it. We will look into it when we have time

test/test_helper.rb
senhalil: https://gitlab.com/mapotempo/optimizer-api/-/issues/876

models/concerns/distance_matrix.rb
senhalil: Yeah it was just an idea, in that case of course it can be done for all projects (if it makes sense). 

All I meant that the router object can handle its own url and the matrix function doesn't really need the url as an argument. Maybve during the passage to the gem, the public interface of the router object can be improved :shrug:

PR #290, opened 2021-09-29
Fix default quantity value

braktar:


test/wrapper_test.rb
```diff
@@ -3086,6 +3086,11 @@ def test_assert_inapplicable_relations
     refute_includes OptimizerWrapper.config[:services][:vroom].inapplicable_solve?(vrp),
                     :assert_no_relations_except_simple_shipments
 
+    problem[:services][0][:quantities] = [{ unit_id: 'kg', value: nil }]
+    vrp = TestHelper.create(problem)
+    assert_includes OptimizerWrapper.config[:services][:vroom].inapplicable_solve?(vrp),
+                    :assert_no_relations_except_simple_shipments
+
```
senhalil, line 3093: Asking just tpo be sure. This test breaks without the default value correction, right?

senhalil (approved):

test/wrapper_test.rb
braktar: Yes, it breaks !

PR #301, opened 2021-10-22
Add sentry

fab-girard:


braktar (approved):

util/job_manager.rb
```diff
@@ -38,7 +38,16 @@ def perform
       Job.current_job_id = self.uuid
       tick('Starting job') # Important to kill job before any code
 
+      puts "#{self.uuid} - Starting job... " + options['checksum']
```
senhalil, line 41: Wouldn't it be better if we use the log directly (so that the job id etc is automatic and conforms the syntax)?
```suggestion
      log "Starting job... #{options['checksum']}"
```

util/job_manager.rb
```diff
@@ -70,6 +79,7 @@ def perform
           Result.set(self.uuid, p)
         end
       }
+      puts "#{self.uuid} - Elapsed time: #{(Time.now - job_started_at).round(2)}s Vrp size: #{services_vrps.size} Key print: #{key_print}"
```
senhalil, line 82:
```suggestion
      log "Elapsed time: #{(Time.now - job_started_at).round(2)}s Vrp size: #{services_vrps.size} Key print: #{key_print}"
```

senhalil (commented):
1. Should we use `log` instead of `puts`? I believe the original config-commits are written with `puts` because we didn't have the logger at the time and Seb didn't clean config-commits when he implemented the logger.

2. @fab-girard  if you like, there is one more commit that we can include in this PR, it adds the `Ending job...` log -- its counterpart `Starting job...` is already included in this PR. The commit is called "More infos in jobs logs".

PR #304, opened 2021-10-26
Docker: use anchor in docker-compose file

alain-andre:
Use the anchor and fix a few things in Docker and Passenger.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -1,42 +1,62 @@
 version: '3.7'
+
+# keep command outside anchor because of performance problems
+x-default-redis: &default-redis
+  image: redis:${REDIS_VERSION:-6.2.5-alpine}
+  deploy:
+    restart_policy:
+      condition: any
+  volumes:
+    - ${APP_HOME:-..}/redis:/data
+
+x-default-environment: &default-environment
+  PASSENGER_APP_ENV: test
+  REDIS_COUNT_HOST: redis-count
+  REDIS_HOST: redis-cache
+  REDIS_RESQUE_HOST: redis
+
+x-environment-resque: &environment-resque
+  <<: *default-environment
+  APP_ENV: test
+  ROUTER_API_KEY: ${ROUTER_API_KEY:-demo}
+  ROUTER_URL: ${ROUTER_URL:-http://localhost:4899/0.1}
+
+x-default-service: &default-service
+  image: ${REGISTRY:-registry.mapotempo.com}/mapotempo-${EDT:-ce}/optimizer-api:${OPTIMIZER_VERSION:-latest}
+  environment:
+    <<: *default-environment
+  networks:
+    - redis_cache
+    - redis_count
+    - resque
+
 services:
   redis:
-    image: redis:${REDIS_VERSION:-6.2.5-alpine}
+    <<: *default-redis
     hostname: redis
-    volumes:
-      - ${APP_HOME:-..}/redis:/data
     command: redis-server --appendonly yes
     networks:
       - resque
 
   redis-cache:
-    image: redis:${REDIS_VERSION:-6.2.5-alpine}
+    <<: *default-redis
     hostname: redis-cache
-    volumes:
-      - ${APP_HOME:-..}/redis:/data
     command: redis-server --save ""
     networks:
       - redis_cache
 
   redis-count:
-    image: redis:${REDIS_VERSION:-6.2.5-alpine}
+    <<: *default-redis
     volumes:
       - ${APP_HOME:-..}/redis-count:/data
     hostname: redis-count
     command: redis-server --appendonly yes
-    deploy:
-      restart_policy:
-        condition: any
     networks:
       - redis_count
 
   api:
-    image: registry.test.com/mapotempo/optimizer-api:latest
-    environment:
-      PASSENGER_APP_ENV: test
-      REDIS_COUNT_HOST: redis-count
-      REDIS_HOST: redis-cache
-      REDIS_RESQUE_HOST: redis
+    <<: *default-service
+    hostname: api
     networks:
       - base
       - redis_cache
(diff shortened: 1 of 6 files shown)


docker/docker-compose.yml
```diff
@@ -1,61 +1,74 @@
...
+    - ${APP_HOME:-..}/redis:/data
+  command: redis-server --appendonly yes
+
+x-environment: &environment
+  PASSENGER_APP_ENV: test
+  REDIS_COUNT_HOST: redis-count
+  REDIS_HOST: redis-cache
+  REDIS_RESQUE_HOST: redis
+
+x-default: &default-api
```
senhalil, line 17: Can we have a uniform naming convention?
```suggestion
x-redis: &default-redis
  image: redis:${REDIS_VERSION:-6.2.5-alpine}
  deploy:
    restart_policy:
      condition: any
  volumes:
    - ${APP_HOME:-..}/redis:/data
  command: redis-server --appendonly yes

x-environment: &default-environment
  PASSENGER_APP_ENV: test
  REDIS_COUNT_HOST: redis-count
  REDIS_HOST: redis-cache
  REDIS_RESQUE_HOST: redis

x-api: &default-api
```

docker/docker-compose.yml
```diff
@@ -1,61 +1,74 @@
...
+        memory: ${MEMORY_API_LIMIT:-10G}
+    update_config:
+      failure_action: rollback
+      order: start-first
+    restart_policy:
+      condition: any
+  volumes:
+    - /etc/localtime:/etc/localtime:ro
+  environment:
+    <<: *environment
```
senhalil, line 31:
```suggestion
    <<: *default-environment
```

docker/docker-compose.yml
```diff
@@ -1,61 +1,74 @@
...
+    <<: *default-api
+    command: bundle exec rake resque:workers
     environment:
-      APP_ENV: test
-      COUNT: 5
-      QUEUES: DEFAULT
-      REDIS_COUNT_HOST: redis-count
-      REDIS_HOST: redis-cache
-      REDIS_RESQUE_HOST: redis
+      <<: *environment
```
senhalil, line 69:
```suggestion
      <<: *default-environment
```

senhalil (commented):
We can name the anchors more uniformly; otherwise, the tests timing is :heavy_check_mark:

senhalil:
@alain-andre I want to approve this PR but if each time another perf test breaks :/ our tests are not this flaky normally. I understand that the usage of anchors changes nothing but this PR also tocuhes the limits and other settings `update_config` etc. Any of these might be the culprit :shrug: If these already exists in the client server (the one that uses Docker image) then we can fix these flaky tests in this PR and call it a day. What do you think?

alain-andre:
> @alain-andre I want to approve this PR but if each time another perf test breaks :/ our tests are not this flaky normally. I understand that the usage of anchors changes nothing but this PR also tocuhes the limits and other settings `update_config` etc. Any of these might be the culprit 🤷 If these already exists in the client server (the one that uses Docker image) then we can fix these flaky tests in this PR and call it a day. What do you think?

I totally agree. I'm going to split this PR into just a refacto and then one with more complex settings.

alain-andre:
@fab-girard @senhalil It's now ok, removing the `appendonly yes` fixes the pipeline. I do not think we really need to append on persisting redis ; asynchronous writing would do the trick wouldn't it?

alain-andre:
> @fab-girard @senhalil It's now ok, removing the `appendonly yes` fixes the pipeline. I do not think we really need to append on persisting redis ; asynchronous writing would do the trick wouldn't it?

OK. After discussing with Fabien, I put the commands back into the compose but out of the anchors because that's what slows down the tests too much. @senhalil

docker/docker-compose.yml
```diff
@@ -1,61 +1,74 @@
...
+  REDIS_HOST: redis-cache
+  REDIS_RESQUE_HOST: redis
+
+x-environment-resque: &environment-resque
+  <<: *default-environment
+  APP_ENV: test
+  ROUTER_API_KEY: ${ROUTER_API_KEY:-demo}
+  ROUTER_URL: ${ROUTER_URL:-http://localhost:4899/0.1}
+
+x-default: &default-service
```
senhalil, line 24: Can we use a uniform naming convention?
```suggestion
x-environment-resque: &default-environment-resque
  <<: *default-environment
  APP_ENV: test
  ROUTER_API_KEY: ${ROUTER_API_KEY:-demo}
  ROUTER_URL: ${ROUTER_URL:-http://localhost:4899/0.1}

x-service: &default-service
```

Just to make sure this is voluntary, it looks like the anchor `x-environment-resque` is never referenced, do we need it?

alain-andre: You are write, something is wrong because of my going back and forth.

senhalil: Now I understand the naming. I  thought you had a specific naming convention (that adds the "default" to the name) but apparently that is not the case. Than I suggest being verbose and naming the anchors accordingly.That is instead of

```
x-redis: &default-redis

x-environment: &default-environment

x-environment-resque: &environment-resque

x-default: &default-service
```
more like this
```
x-default-redis: &default-redis

x-default-environment: &default-environment

x-resque-environment: &resque-environment

x-default-service: &default-service
```

At the end of the day, if you fixed the other issue, this naming issue doesn't matter that much but it is better to have a coherent naming convention to not to get lost after some time has passed.

senhalil (commented):
@alain-andre it looks like there is a conlict, can you rebase and add `SplitClusteringTest#test_instance_same_point_day` to the `methods_to_retry` list in test_helper.rb please? So that we can see if the test is flaky or if there still is a perf issue. 

Otherwise, I think this PR is good to go :rocket: unless you want to touch up the anchor names.

senhalil (approved):

senhalil (changes requested):
Dockerfile has a problem with the deafult values

senhalil (approved):

PR #305, opened 2021-10-27
Docker: manage cors in app

alain-andre:
Manage CORS in app instead of nginx


server.ru
```diff
@@ -24,7 +24,12 @@ require 'action_dispatch'
 use Rack::Cors do
   allow do
     origins '*'
-    resource '*', headers: :any, methods: :any
+    resource '*',
+      headers: :any,
+      methods: :any,
+      expose: ['Cache-Control', 'Content-Encoding', 'Content-Type'],
+      max_age: 1728000,
+      credentials: false
```
senhalil, line 32: On the deleted file `credentials` and `methods` are different

```
  add_header 'Access-Control-Allow-Credentials' 'true' always;

senhalil (commented):
It is probably nothing but to make sure..

server.ru
alain-andre: Yes it's because then gem asks to specify the origin when you set `credentials` to `true`. We don't need it in our case : https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Credentials

senhalil (approved):

PR #307, opened 2021-11-12
Fix #882 violated multi pickup delivery

senhalil:
Makes sure that simplified relations are not ignored in the protobuf collection.

- [x] It needs the following optimizer-ortools PR since simplified complex shipments are connected via `:sequence` relations and this PR fixes an issue with it https://github.com/Mapotempo/optimizer-ortools/pull/33

Closes optimizer-api/-/issues/882

### DIFF
--- a/wrappers/wrapper.rb
+++ b/wrappers/wrapper.rb
@@ -1112,6 +1112,8 @@ module Wrappers
             expanded_services << new_service
           }
 
+          # For some reason, or-tools performance is better when the sequence relation is defined in the inverse order.
+          # Note that, activity.duration's are set to zero except the last duplicated service (so we model exactly same constraint). 
           sequence_relations << Models::Relation.create(type: :sequence, linked_ids: sequence_relation_ids.reverse)
         }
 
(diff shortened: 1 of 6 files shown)


wrappers/wrapper.rb
```diff
@@ -1112,7 +1112,7 @@ def simplify_complex_multi_pickup_or_delivery_shipments(vrp, result = nil, optio
             expanded_services << new_service
           }
 
-          sequence_relations << Models::Relation.create(type: :sequence, linked_ids: sequence_relation_ids)
+          sequence_relations << Models::Relation.create(type: :sequence, linked_ids: sequence_relation_ids.reverse)
```
braktar, line 1115: Could you mention in a comment why we do a reverse ?

senhalil:
```suggestion
          # For some reason, or-tools performance is better when the sequence relation is defined in the inverse order.
          # Note that, activity.duration's are set to zero except the last duplicated service (so we model exactly same constraint). 
          sequence_relations << Models::Relation.create(type: :sequence, linked_ids: sequence_relation_ids.reverse)
```

braktar (approved):

PR #309, opened 2021-11-15
Introduce Solution Model

braktar:


api/v01/vrp.rb
```diff
@@ -170,48 +171,42 @@ class Vrp < APIBase
...
 
-            unless solution.nil? || solution[:result].nil? || solution[:result].is_a?(Array)
-            # TODO: solution[:result] should always be an array, find out why it is not and
-            # remove this check and other similar ones -- i.e.,  [solution[:result]].flatten(1).
-              solution[:result] = [solution[:result]]
-            end
-
             if output_format == :csv && (job.nil? || job.completed?) # At this step, if the job is nil then it has already been retrieved into the result store
-              present(OutputHelper::Result.build_csv(solution[:result]), type: CSV)
+              present(OutputHelper::Result.build_csv(job_object[:result].map(&:deep_symbolize_keys!)), type: CSV)
```
senhalil, line 194: `deep_symbolize_keys` is not free, we can load the object with symbolised key directly, Oj.load shoudl have an option for that.

api/v01/vrp.rb
```diff
@@ -170,48 +171,42 @@ class Vrp < APIBase
             id = params[:id]
             job = Resque::Plugins::Status::Hash.get(id)
             stored_result = APIBase.dump_vrp_dir.read([id, params[:api_key], 'solution'].join('_'))
-            solution = stored_result && Oj.load(stored_result) rescue Marshal.load(stored_result) # rubocop:disable Security/MarshalLoad, Style/RescueModifier
+            job_object = stored_result && Oj.load(stored_result) rescue Marshal.load(stored_result) # rubocop:disable Security/MarshalLoad, Style/RescueModifier
```
senhalil, line 174: why not `result_object` or `solution_object`  or just `result` / `solution`?

api/v01/vrp.rb
senhalil: actually if we can do `job_object[:result]` why the rest is not already symbolized :thinking: ?

models/concerns/as_json.rb
```diff
@@ -0,0 +1,189 @@
...
+#
+# Mapotempo is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY
+# or FITNESS FOR A PARTICULAR PURPOSE.  See the Licenses for more details.
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with Mapotempo. If not, see:
+# <http://www.gnu.org/licenses/agpl.html>
+#
+require 'active_support/concern'
```
senhalil, line 18: not needed
```suggestion
```

models/concerns/validate_data.rb
```diff
@@ -27,6 +27,8 @@ def check_consistency(hash)
     hash[:relations] ||= []
     @hash = hash
 
+    ensure_no_conflicting_skills
+
```
senhalil, line 31: Is this correct? It looks like a function not related to model solution dev.

models/concerns/as_json.rb
senhalil: There are few others like this that can be removed. All gem `require` operations are done in environement.rb automatically via Gemfile.

senhalil (commented):
Needs a rebase

api/v01/vrp.rb
braktar: Actually, the object stored in redis is neither a single solution nor a single result.
It holds all the properties of the job and which contains a key called :result which is collection of solutions.
By changing the naming, I want to separate the internal objects from the objects manipulated on the api side to avoid confusion

senhalil: Yes but the part that is dumped (and loaded here) is a result, is it not?

braktar: We directly store the content of the job returned through Redis

models/concerns/validate_data.rb
braktar: :+1:  It should not be there

api/v01/vrp.rb
senhalil: I think I understand the differentiation you want to make but I don't think `job_object` is the way to go, because we have a class Job and real objects of this class which is confusing.

I know that it is not directly a single result nor a solution but at the end of the day we do get/set this "object" via `OptimizerWrapper::Result` class that's why I suggest calling it result_object:
```ruby
result_object ||= OptimizerWrapper::Result.get(id)
```

With the suffix `_object` we give a heads-up that it is not exactly `solution[:result]`. Would that work for you?

PR #310, opened 2021-11-15
Use vroom instead of sim_annealing

braktar:


senhalil:
@braktar can you rebase this branch, it shows conflict because of bundle update

senhalil:
> @braktar can you rebase this branch, it shows conflict because of bundle update

I just did fix the conflicts

senhalil:
@braktar I tried to fix the conflicts but it looks like it needs your attention. I am not even sure if we want to manage and merge these PRs separately or from the "main" one. 

If we would like to do it via the main PR, we can mark these "separate" ones as draft maybe. What do you think?

PR #311, opened 2021-11-15
Model solution optimizer logic

braktar:


lib/cleanse.rb
```diff
@@ -17,108 +17,94 @@
...
-          quantity.unit.id
+        useful_units = (current.loads.map{ |l|
+          l.quantity.unit.id
         }.compact & capacities)
 
-        current_fill_units = current.quantities.collect{ |quantity|
-          quantity.unit.id if quantity.fill
+        current_fill_units = current.loads.map{ |l|
+          l.quantity.unit.id if l.quantity.fill
         }.compact
```
senhalil, line 66: Can you replace `l` with `load`, it is harder to read since it looks like :open_mouth:  `1.quantity`

braktar: as load is a method of ruby Kernel, i will avoid to use it, may `loa` suits ?

senhalil: ah :face_exhaling:  effectively.. in that case, as you wish, you can ignore my comment.

PR #312, opened 2021-11-15
Fix: split_independent_vrp respects skills and keeps relations

senhalil:


test/wrapper_test.rb
```diff
@@ -2883,31 +2883,87 @@ def test_check_distances_if_distance_and_time_with_lateness
                  "all should be infeasible due to distance constraint"
   end
 
-  def test_number_of_service_vrps_generated_in_split_independent
+  def test_split_independent_vrp_does_not_reset_relations
+    problem = VRP.independent_skills
+    # The services in the same realtion would stay on the same vehicle after the split_independent_vrp even without
```
braktar, line 2888:
```suggestion
    # The services in the same relation would stay on the same vehicle after the split_independent_vrp even without
```

test/wrapper_test.rb
```diff
@@ -2883,31 +2883,87 @@ def test_check_distances_if_distance_and_time_with_lateness
...
+    # the relations (due to skills). Here we test only if the relations stay intact after split_independent_vrp.
+    # split_independent_vrp does not take into account relations but if the services are not feasible in the first place
```
braktar, line 2890: Is it possible to ease the understandability of this phrase by reducing the number negations ?
```suggestion
    # split_independent_vrp only takes into account relations if the services are feasible in the first place
```

test/wrapper_test.rb
```diff
@@ -2883,31 +2883,87 @@ def test_check_distances_if_distance_and_time_with_lateness
...
+    # i.e., both cannot be served by the same vehicle, then split cannot do anything about it.
```
braktar, line 2891:
```suggestion
    # i.e., both services of the same relation cannot be served by the same vehicle, then split cannot do anything about it.
```

test/wrapper_test.rb
senhalil: In fact split_independent_vrp never takes into account relation. I improved the explanation. Unresolve this conversation if it is still not clear.

braktar (approved):

PR #313, opened 2021-11-16
Add milliseconds to logger

fab-girard:


senhalil (changes requested):
The regexp of `test_unfounded_avancement_message_change` needs to be corrected

senhalil (approved):

PR #319, opened 2021-12-07
Fix bundle update and docker issues

senhalil:


Rakefile
```diff
@@ -11,6 +11,8 @@ namespace :resque do
   end
 
   task clean_working_job_ids: :prune_dead_workers do
+    next if ENV['APP_ENV'] == 'production'
```
braktar, line 14: Don't we want to prune dead workers even in production ?

senhalil: Containers cannot see legit workers of other cont.

senhalil: This task is not really needed on the server-side anymore (thanks to the fact that docker workers continue until they finish the job at hand and then set the job status to completed...) it is still of use on local to prevent accumulation of jobs (ergo the `if ENV['APP_ENV'] == 'production'` condition)

braktar (approved):

alain-andre (approved):

PR #321, opened 2021-12-13
Model solution 2

braktar:


models/concerns/as_json.rb
```diff
@@ -0,0 +1,239 @@
...
+# You should have received a copy of the GNU Affero General Public License
+# along with Mapotempo. If not, see:
+# <http://www.gnu.org/licenses/agpl.html>
+#
+module ContainedPointAsJson
+  extend ActiveSupport::Concern
+
+  def as_json(options = {})
+    point = super
+
```
senhalil, line 23: AcriveModel initilises options with `nil`
and we need to call `super` with options so that the lower level as_json gets the options  
https://apidock.com/rails/v3.2.1/ActiveModel/Serializers/JSON/as_json

models/base.rb
```diff
@@ -64,6 +85,15 @@ def self.has_many(name, options = {})
           "#{name[0..-4]}y_ids".to_sym
         end
 
+      case options[:as_json]
+      when :ids
+        json_fields << ids_function_name.to_s
```
senhalil, line 90: Is there a specific reason for not using symbol?

models/base.rb
```diff
@@ -52,8 +50,31 @@ def initialize(hash)
...
+      hash = {}
+      self.class.json_fields.each{ |field_name|
+        hash[field_name] = self.send(field_name) if self.send(field_name)
+      }
+      hash
+    rescue => e
+      puts e
+      puts e.backtrace
+
+      byebug
```
senhalil, line 69: forgotten

models/base.rb
braktar: The hash keys returned by as_json are strings

lib/interpreters/split_clustering.rb
```diff
@@ -288,7 +283,8 @@ def self.split_solve_sub_vrp(service_vrp, job = nil, &block)
 
       service_vrp[:vrp] = create_sub_vrp(ss_data)
 
-      OptimizerWrapper.define_process(service_vrp, job, &block)
+      solution = OptimizerWrapper.define_process(service_vrp, job, &block)
+      solution.parse(service_vrp[:vrp])
```
senhalil, line 287: why do the solution need the vrp to be able to parse itself? Since a solution belongs to a specific VRP, I think we can have the "reference" of the VRP inside the solution object so that, we don't need to pass the vrp to it for different operations, what do you think?

lib/interpreters/split_clustering.rb
```diff
@@ -614,21 +613,24 @@ def self.update_matrix_index(vrp)
...
+      vrp_hash = JSON.parse(vrp.empty_json, symbolize_names: true)
+      sub_vrp = Models::Vrp.create(vrp_hash)
+      sub_vehicles = []
+      sub_routes = []
       # TODO: Within Periodic Vehicles require to have unduplicated ids
       if available_vehicles_indices
-        sub_vrp.vehicles.delete_if.with_index{ |_v, v_i| !available_vehicles_indices.include?(v_i) }
-        sub_vrp.routes.delete_if{ |r|
+        sub_vrp.vehicles = vrp.vehicles.select.with_index{ |_v, v_i| available_vehicles_indices.include?(v_i) }
+        sub_vrp.routes = vrp.routes.select{ |r|
```
senhalil, line 631: I am not sure about this. The above warning looks old, and I feel like the behavior change might dangerous, lets discuss this one.

senhalil (changes requested):
13/98 done

models/solution/route_info.rb
```diff
@@ -0,0 +1,45 @@
...
+  class Solution < Base
+    class Route < Base
+      class Info < Base
+        field :total_time
+        field :total_travel_time
+        field :total_waiting_time
+
+        field :total_distance
+
+        field :total_travel_value
```
senhalil, line 30: maybe these can become function as well

models/solution/solution_info.rb
```diff
@@ -0,0 +1,40 @@
...
+module Models
+  class Solution < Base
+    class Info < Base
+      field :total_time, default: 0
+      field :total_travel_time, default: 0
+      field :total_waiting_time, default: 0
+
+      field :total_distance, default: 0
+
+      field :total_travel_value
```
senhalil, line 29: total_x function?

models/solution/stop.rb
```diff
@@ -0,0 +1,97 @@
...
+#
+require './models/base'
+require './models/solution/parsers/stop_parser'
+
+module Models
+  class Solution < Base
+    class Stop < Base
+      include SolutionStopAsJson
+
+      # field :point_id
```
senhalil, line 26: comment, voluntary? if so what is the significance

models/solution/stop_info.rb
```diff
@@ -0,0 +1,41 @@
...
+      class Info < Base
+        field :id, as_json: :none
+        field :day_week_num
+        field :day_week
+
+        field :travel_distance
+        field :travel_time
+        field :travel_value
+
+        field :waiting_time, default: 0
```
senhalil, line 32: how can we make this more automatic, do we really need this field in the class, maybe we can calculated it with the info we have once at the very end (eventually inside the grape entity)?

senhalil: And maybe we can get rid of the "parse" logic this way/..

senhalil (changes requested):

wrappers/ortools.rb
```diff
@@ -561,21 +485,21 @@ def run_ortools(problem, vrp, thread_proc = nil, &block)
 
       correspondant = { 'path_cheapest_arc' => 0, 'global_cheapest_arc' => 1, 'local_cheapest_insertion' => 2, 'savings' => 3, 'parallel_cheapest_insertion' => 4, 'first_unbound' => 5, 'christofides' => 6 }
 
-      raise StandardError, "Inconsistent first solution strategy used internally: #{vrp.preprocessing_first_solution_strategy}" if vrp.preprocessing_first_solution_strategy.any? && correspondant[vrp.preprocessing_first_solution_strategy.first].nil?
+      raise StandardError, "Inconsistent first solution strategy used internally: #{vrp.configuration.preprocessing.first_solution_strategy}" if vrp.configuration.preprocessing.first_solution_strategy.any? && correspondant[vrp.configuration.preprocessing.first_solution_strategy.first].nil?
```
senhalil, line 488:
```
config = vrp.configuration
preprocessing = config.preprocessing
resolution = config.resolution
```
 to simplify

senhalil (changes requested):

lib/interpreters/split_clustering.rb
braktar: I have removed the above comment. The behaviour has change in order to avoid to keep any link between old and duplicated objects

senhalil: but here we keep a link don't we? `sub_vrp.vehicles = vrp.vehicles.select` would use the vehicles of the original vrp in the sub problems, no?

braktar: yes, we select parts of the upper level vrp, but only to transform these into a json and then a hash using `sub_vrp_hash = JSON.parse(sub_vrp.to_json, symbolize_names: true)`

wrappers/wrapper.rb
```diff
@@ -998,31 +894,26 @@ def prioritize_first_available_trips_and_vehicles(vrp, result = nil, options = {
...
-
-          cost_diff = route[:cost_details].fixed - vehicle[:fixed_cost_before_adjustment]
+          next unless route.vehicle[:fixed_cost_before_adjustment] && route.cost_info&.fixed&.positive?
 
-          route[:cost_details].fixed -= cost_diff
-          result[:cost_details].fixed -= cost_diff
+          route.cost_info.fixed = route.vehicle[:fixed_cost_before_adjustment]
         }
-        result[:cost] = result[:cost_details].total
+        solution.update_costs
```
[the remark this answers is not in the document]
braktar: See internal ticket 892

models/concerns/as_json.rb
```diff
@@ -0,0 +1,43 @@
...
+module SolutionStopAsJson
+  extend ActiveSupport::Concern
+
+  def as_json(options = nil)
+    stop = super
+
+    return stop unless self.is_a? Models::Solution::Stop
+
+    stop.delete('id') if self.type == :depot
+    stop
```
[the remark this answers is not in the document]
braktar: See internal ticket 893

models/configuration.rb
```diff
@@ -28,7 +28,7 @@ class Configuration < Base
   class Preprocessing < Base
     field :force_cluster, default: false
     field :max_split_size, default: nil
-    field :partition_method, default: nil
+    field :partition_mode, default: nil
     field :partition_metric, default: nil
```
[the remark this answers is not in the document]
braktar: See internal ticket 894

models/service.rb
```diff
@@ -49,15 +51,21 @@ class Service < Base
...
+    has_many :sticky_vehicles, class_name: 'Models::Vehicle', as_json: :ids
     has_many :quantities, class_name: 'Models::Quantity'
-    has_many :relations, class_name: 'Models::Relation'
+    has_many :relations, class_name: 'Models::Relation', as_json: :none
+
+    def self.create(hash)
+      hash[:skills] = [] if hash[:skills].to_a.empty?
+
+      super(hash)
+    end
```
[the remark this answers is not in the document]
braktar: the logic implemented in base.rb does not handle skills, as the field may be transmitted with a nil value due to the input in csv format.

models/solution/route_info.rb
braktar: See internal ticket 896

models/solution/solution_info.rb
braktar: See internal ticket 896

models/solution/stop_info.rb
braktar: The internal ticket 897

models/vrp.rb
```diff
@@ -143,6 +74,74 @@ def self.create(hash, delete = true)
...
+      else
+        self.services.map{ |service| Models::Solution::Stop.new(service, reason: unassigned_hash[service.id]) }
+      end
+    end
+
+    def unassigned_rests
+      self.vehicles.flat_map{ |vehicle|
+        vehicle.rests.map{ |rest| Models::Solution::Stop.new(rest) }
+      }
+    end
```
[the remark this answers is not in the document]
braktar: I checked, and these methods are well located and are mostly related to the cases where the solution is empty due to the lack of vehicle or a vrp with no services

models/solution/parsers/route_parser.rb
```diff
@@ -0,0 +1,220 @@
...
+        previous_index = matrix_index if stop.type != :rest
+      }
+
+      if @route.info.end_time && @route.info.start_time
+        @route.info.total_time = @route.info.end_time - @route.info.start_time
+      end
+      @route.info.total_travel_time = total[:time].round if dimensions.include?(:time)
+      @route.info.total_distance = total[:distance].round if dimensions.include?(:distance)
+      @route.info.total_travel_value = total[:value].round if dimensions.include?(:value)
+
```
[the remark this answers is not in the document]
braktar: https://gitlab.com/mapotempo/optimizer-api/-/issues/896

api/v01/vrp.rb
```diff
@@ -172,48 +173,42 @@ class Vrp < APIBase
...
 
-            unless solution.nil? || solution[:result].nil? || solution[:result].is_a?(Array)
-            # TODO: solution[:result] should always be an array, find out why it is not and
-            # remove this check and other similar ones -- i.e.,  [solution[:result]].flatten(1).
-              solution[:result] = [solution[:result]]
-            end
-
             if output_format == :csv && (job.nil? || job.completed?) # At this step, if the job is nil then it has already been retrieved into the result store
-              present(OutputHelper::Result.build_csv(solution[:result]), type: CSV)
+              present(OutputHelper::Result.build_csv(result_object[:result]), type: CSV)
```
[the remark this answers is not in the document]
braktar: See internal ticket 898

lib/interpreters/split_clustering.rb
braktar: See internal ticket 899

lib/heuristics/dichotomious_approach.rb
```diff
@@ -76,66 +82,78 @@ def self.dichotomious_heuristic(service_vrp, job = nil, &block)
...
+              sub_service_vrp[:vrp].configuration.resolution.split_number = sub_service_vrps[0][:vrp].configuration.resolution.split_number + 1
+              sub_service_vrp[:vrp].configuration.resolution.total_split_number =
+                sub_service_vrps[0][:vrp].configuration.resolution.total_split_number
             end
-            if sub_service_vrp[:vrp].resolution_minimum_duration
-              sub_service_vrp[:vrp].resolution_minimum_duration *= sub_service_vrp[:vrp].services.size / service_vrp[:vrp].services.size.to_f * 2
+            if sub_service_vrp[:vrp].configuration.resolution.duration
+              sub_service_vrp[:vrp].configuration.resolution.duration *=
+                sub_service_vrp[:vrp].services.size / service_vrp[:vrp].services.size.to_f * 2
+            end
```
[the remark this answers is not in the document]
braktar: actually, there is no other change here than the configuration edit and a back to the line

senhalil (approved):

senhalil:
@braktar in zone.rb vehicles is defined as full objects `has_many :vehicles, class_name: 'Models::Vehicle'`  but it should be with `as_json: :ids`, right?

senhalil:
`Models::Solution` has `has_many :unassigned` as a member but it is not plural so it violates the "plural_name => singular_name_ids" vehicles/vehicle_ids, capacities/capacity_ids logic. Which messes the plural_id_function_name generation. Can we (internally) do `unassigned_services` since internally we only have `service` object in any case? What do you think?

PR #324, opened 2021-12-30
Fix wrong number of visits returned in result

senhalil:
Fixes the issue of "lost services" in beta and prod.
api-optimizer-beta/issues/19133 and api-optimizer-prod/issues/18249

And fixes api-optimizer-beta/issues/19214 by bumping balanced_vrp_clustering

### DIFF
--- a/wrappers/wrapper.rb
+++ b/wrappers/wrapper.rb
@@ -907,9 +907,31 @@ module Wrappers
     end
 
     def simplify_constraints(vrp)
+      vrp_subfields_to_check = %i[points relations rests routes services units vehicles].freeze
+
       simplifications.each{ |simplification|
+        expected_vrp_subfield_counts = vrp_subfields_to_check.map{ |subfield| [subfield, vrp.send(subfield).size] }.to_h
+
         simplification_activated = self.send(simplification, vrp, nil, mode: :simplify)
-        log "#{simplification} simplification is activated" if simplification_activated
+
+        if simplification_activated
+          log "#{simplification} simplification is activated"
+        else
+          actual_vrp_subfield_counts = vrp_subfields_to_check.map{ |subfield| [subfield, vrp.send(subfield).size] }.to_h
+
+          next unless expected_vrp_subfield_counts != actual_vrp_subfield_counts
+
+          tags = {
+            simplification: simplification,
+            expected_counts: expected_vrp_subfield_counts,
+            actual_counts: actual_vrp_subfield_counts,
+          }
+
+          log_msg = 'Lost objects in a non-active simplification routine'
+
+          log log_msg, tags.merge(level: :warn)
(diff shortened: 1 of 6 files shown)


wrappers/wrapper.rb
```diff
@@ -1038,7 +1060,10 @@ def simplify_complex_multi_pickup_or_delivery_shipments(vrp, result = nil, optio
         # the original service id with the new expanded service ids in the routes (and rewind it afterwards)
         # but if the complex relations are already in feasible initial routes then it might not worth it.
         # So we can wait for a real use case arrives and we have an instance to test.
-        return nil if services_in_multi_shipment_relations.any?{ |s| vrp.routes.any?{ |r| r.mission_ids.include?(s.id) }}
+        if services_in_multi_shipment_relations.any?{ |s| vrp.routes.any?{ |r| r.mission_ids.include?(s.id) }}
+          vrp.services += services_in_multi_shipment_relations
```
braktar, line 1064: If services of the same shipment are in different routes, won't it generate duplicated services ?

senhalil: I just checked this but couldn't see an issue, maybe I didn't get your question well :/ but from where I look, `services_in_multi_shipment_relations` is defined as below so we separate original `vrp.services` into "complex_services" and "non_complex_services" (the first becomes `services_in_multi_shipment_relations` the latter replaces the `vrp.services`)

```
services_in_multi_shipment_relations, vrp.services =
      vrp.services.partition{ |s| s.relations.count{ |r| r.type == :shipment } > 1 }
```

so with `          vrp.services += services_in_multi_shipment_relations` operation we just put back what was there in the first place. Does this answer your question?

braktar: My though is that if we separate the problem in two parts, pickup and delivery 1 is in sub problem 1 and initial route 1, delivery 2 is in sub problem 2 and initial route 2. Won't we duplicate all the services due to this invalid initial routes ?

senhalil: I don't understand, who is doing such a separation but if the [pick1, del1] and [pick1, del2] are the shipment relations, then how come pick1 and del1 are in subproblem1 and del2 is in another subproblem? Neither max_split nor split_independent can make such a split, because of respecting the relations. If that is not the case, we should fix that.

But apart from all this, I still don't see the issue. We just cut/partition vrp.services in two parts (complex, non-complex) then (if any of the complex shipments are in a route) we put these two parts back together. I don't see how we can duplicate anything. You might want to take a look at 10 lines before (L1054). Does this answer your question? If not, maybe I still don't get the question :/

braktar: If it is not possible to have such case due to relations, then yes, there is no issue ;)

braktar (approved):

braktar (approved):

PR #341, opened 2022-02-03
Bump optimizer-ortools to v1.11.0 - Time horizon rework

braktar:


test/real_cases_test.rb
```diff
@@ -311,7 +310,8 @@ def test_ortools_open_timewindows
       assert result[:routes].sum{ |r| r[:total_travel_time] } <= 13225, "Too long travel time: #{result[:routes].sum{ |r| r[:total_travel_time] }}"
 
       # Check elapsed time
-      assert result[:elapsed] < 15000, "Too long elapsed time: #{result[:elapsed]}"
+      # The rework of time horizon in optimizer-ortools has decreased the computation time drastically
+      assert result[:elapsed] < 1000, "Too long elapsed time: #{result[:elapsed]}"
```
senhalil, line 314: I just realized there might be even more tests like this. The easiest way I can think of is i) activate timing report ii) fix the seed (so that tests are run at the same order) iii) run this setup on github-actions with `v1.10` and `v1.11` and iv) compare the output via libreoffce-calc to find the tests with considerable amount of time improvement.

We can multiply the time limits with the ratio of the change. What do you think @braktar ?

braktar: I tried to spot such improvement in other tests. But there is not much differences on other instances

senhalil (approved):

PR #342, opened 2022-02-03
Day color selector

AntoineCAMILLE:
color-seleel
color-select
color-sele
color-sel


public/js/result.js
```diff
@@ -295,6 +450,12 @@ function vehicleToGeojson(vehicle) {
 
   L.geoJSON({ type: "FeatureCollection", features: vehicle.points }, {
     pointToLayer: function(geoJsonPoint, latlng) {
+      if (geoJsonPoint.properties.day && geoJsonPoint.properties.work_day) {
+        const day = (geoJsonPoint.properties.day - 1) % 7;
```
braktar, line 454:
```suggestion
        const day = (geoJsonPoint.properties.day) % 7;
```
Starting the week with monday instead of tuesday is much more natural.
```

braktar (changes requested):

public/js/result.js
AntoineCAMILLE: The days order is define by the day number 1 so if it's tuesday it starts with tuesday, i think this order provides more informations than just making the week starts with monday
I don't understand the suggestion could you explain the reason ? (the `-1` is for making day 1 as day 0 so it fits easily in an Array)

senhalil: optimizer-api always makes the monday day_index=0, tuesday => day_index=1, etc.. This is not about the actual planning, day or the horizon. These days are "work_day_indices" and not "real_day_indices". So because of the `-1` , the interface will always start with a Tuesday and this does not provide more information but it make the interface look a bit broken.

Eventually we can make the UX locale dependent and make `+1` for en-US (since they like to start the week with a Sunday) but the locality is out of scope of this dev.

braktar: The schedule range may start with day 0. Day 0 is always a monday.
Even if the schedule range is [4, 26] or [04-02-2022, 26-02-2022] (which is converted to [4, 26] as we are a friday)
if day%7 is zero we are sure it is a monday.

It doesn't really matter if the total horizon starts a monday, a tuesday or any other day. As the problem is periodic it doesn't really matter which week day is the first day.

AntoineCAMILLE:
optim-aeaaqgvg

braktar (approved):

AntoineCAMILLE:
Screenshot_20220207_180243

PR #345, opened 2022-02-07
VROOM cannot handle empty or fill quantities

braktar:


senhalil (approved):

senhalil:
If it is due to performance, maybe we can add `test_vrp_ten_routes_with_rest` to the "re-try" list

PR #347, opened 2022-02-14
Once OR-Tools has finished pids should be reset

braktar:


wrappers/ortools.rb
```diff
@@ -582,6 +582,7 @@ def run_ortools(problem, vrp, thread_proc = nil, &block)
       stdin&.close
       stdout_and_stderr&.close
       pipe&.close
+      thread_proc&.call([])
```
senhalil, line 585: What does this do exactly?

senhalil (approved):

wrappers/ortools.rb
braktar: Once the resolution is finished, the pid is not reseted. So, a DELETE request try to kill this pid which no more exist.

PR #349, opened 2022-02-14
Fix internal vrp import

braktar:


api/v01/entities/vrp_input.rb
```diff
@@ -591,7 +591,7 @@ module VrpVehicles
 
   params :router_options do
     optional :router_mode, type: String, desc: '`car`, `truck`, `bicycle`, etc... See the Router Wrapper API doc.'
-    exactly_one_of :matrix_id, :router_mode
+    at_least_one_of :matrix_id, :router_mode
```
senhalil, line 595: Is there a  use case for a user to have both of these params?

lib/interpreters/split_clustering.rb
```diff
@@ -683,7 +683,7 @@ def self.build_partial_service_vrp(service_vrp, partial_service_ids, available_v
       sub_vrp_hash = JSON.parse(sub_vrp.to_json, symbolize_names: true)
 
       vehicle_ids = sub_vrp_hash[:vehicles]&.map{ |v| v[:id] } || []
-      sub_vrp_hash[:services].each{ |service|
+      sub_vrp_hash[:services]&.each{ |service|
```
senhalil, line 677: I feel like I am responsible for this -- by suggesting "not dumping the fields with default values". Do you think this is going to be an issue?

Ideally, we were going to handle all the hash related operation before (or during) model.create and inside the code base have the gurantee that we always have sensible defaults (for example an empty array), or at least the idea was to aim for this. 

But it looks like because we skip the defaults (which was done to prevent dumps having lots of unnecessary fields) we are back to square zero. 

I didn't think of this internal use case when we were discussing "skipping the defaults" may be we need a `dump_with_defaults: true/false` switch which is by default false and at places like this we can activate it.

On the other hand, it looks like we don't have many issue with this so tell me what you think about all this :pray:

models/base.rb
```diff
@@ -122,6 +122,9 @@ def self.has_many(name, options = {})
       # vehicles -> vehicle_ids | capacities -> capacity_ids | matrices -> matrix_ids
       ids_function_name = "#{ActiveSupport::Inflector.singularize(name.to_s)}_ids".to_sym
 
+      add_default_value(name, options[:options] || [])
+      add_default_value(ids_function_name, options[:options] || [])
```
senhalil: If we are sure to have a default value, thanks to this logic 

may be we should modify 

```ruby
  redefine_method(name) do
      self[name] ||= []
    end
```

to

```ruby
  redefine_method(name) do
      self[name] || self.class.default_attributes[field_name] # if we are sure about the `add_default_value`
    end
```

so that the "default" value is defined in one location otherwise we might overwrite the default value in the name= or name() functions. Is that make sense?

models/quantity.rb
```diff
@@ -19,8 +19,8 @@
 
 module Models
   class Quantity < Base
-    field :fill, default: false
-    field :empty, default: false
+    field :fill # Active_hash did not record default values set to false
+    field :empty # because the condition to reach add_default_value is `if options[:default]`
```
senhalil: :facepalm:  :face_with_spiral_eyes: :face_with_head_bandage: :face_exhaling:

What do you suggest? We can set the default value in the `quantity.create`? But I guess the main question is "do we need it?" if the default value is `nil`, it is the same as `false` and we will prevent ourselves from dumping `false` by mistake since `nil`s are already eliminated. So if this is good for you, it is good for me too.

senhalil: But then we always need to remember to make `!!` with fill and empty -- and I don't really like that :/ there will be bugs. In any case, tell me what you think about (i) having the defaults in create or (ii) fixing the active_hash bug with `options.has_key?[:default]`

api/v01/entities/vrp_input.rb
braktar: It could happen if a user transmits its own matrix, but still want the route polylines to be returned

lib/interpreters/split_clustering.rb
braktar: Removing the `&` will force us to remove `allow_blank: false` on services from API input. This doesn't sounds to be a good idea regarding our will to impose a more clean vrp definition.

If this `&` is one of the only counterpart, it is a small price to pay

api/v01/entities/vrp_input.rb
senhalil: Then we might want to do `allow_blank: false` for both of these,  no? Otherwise `nil` `nil` for both passes this verification

lib/interpreters/split_clustering.rb
senhalil: I don't see the connection to the API immediately :thinking: We can have both, no? Until we arrive split_clustering.rb, services is already initialized with empty array `[]` so if the user POSTs a vrp without any services, there is no problem -- so I don't see the price to pay -- did I misunderstand you?

The issue, if I got it correctly, is related to the internal dump/load via as_json -- we skip the defaults so if the services are empty, it is not dumped so the `services` key doesn't exist in the hash, so this forces us to have the `&`. 

My suggestion was, when using `as_json` internally dumping everything, including the defaults because that way we don't need to constantly think about "if this key exists or not". The goal was to have everything initialized at the beginning and not worry about it inside the code-base, and what we did for `as_json` (while only thinking about the final "dump-ability") violates this.

braktar: My point was that if we want to use the internal vrps in order to analyse a problem. It requires to edit the vrps manually, or to allow blank for many objects. From my point of view as_json should ease our life both by avoiding marshal dump/load and by making the debug as easy as possible

models/quantity.rb
braktar: There is only an issue with our protobuf because we expect boolean values.

Anyway, as we use our own fork for active_hash, we indeed can change the condition to allow the value `false`.
As this is the default value, we will in any case never return it with as_json.

Moreover `!!` is often used in Mapotempo Web to turn nil values into booleans

lib/interpreters/split_clustering.rb
senhalil: Now I understand your comment, I missed the "if we were to dump the default values" context.

However, what I suggested was to have a `dump_defaults:false/true` flag, so that in the code we have the "full" model with defaults but if we dump for debugging or storage, it is dumped with `dump_defaults:false` which means we don't need to modify the endpoint nor the vrp.

I don't know :/ maybe this is too "heavy" for not much gain. I was just trying to prevent the internal `nil doesn't have x method` errors, it is hard to predict these bugs because it is vrp dependant :/

models/quantity.rb
```diff
@@ -30,5 +30,11 @@ class Quantity < Base
     # validates_numericality_of :setup_value
 
     belongs_to :unit, class_name: 'Models::Unit', as_json: :id
+
+    def initialize(hash)
+      hash[:empty] = false unless hash.key?(:empty)
+      hash[:fill] = false unless hash.key?(:fill)
```
braktar: It won't allow to fix the issue of the default value returned with as_json :disappointed:

api/v01/vrp.rb
```diff
@@ -162,6 +162,32 @@ class Vrp < APIBase
           end
         end
 
+        resource :validate do
+          desc 'Validate VRP problem', {
+            nickname: 'submit_vrp',
```
senhalil: this is the same nickname with the POST, is it voluntary (wouldn't it lead to a clash)? If not, we can have `validate_vrp` as the nickname
```suggestion
            nickname: 'validate_vrp',
```

lib/interpreters/split_clustering.rb
senhalil: Just to be sure and so that we have it as a conclusion here; you decided to not the dump the defaults (internally or externally) and use a `&` internally when we are handling a hash created via `as_json`. 

Is that it?

models/quantity.rb
```diff
@@ -30,5 +30,13 @@ class Quantity < Base
     # validates_numericality_of :setup_value
 
     belongs_to :unit, class_name: 'Models::Unit', as_json: :id
+
+    def self.field(name, options = {})
+      # Active_hash did not record default values set to false
+      # because the condition to reach add_default_value is `if options[:default]`
+      super(name, options)
+      add_default_value(:empty, false)
+      add_default_value(:fill, false)
```
senhalil: it is a good idea to to override `field` for this purpose  :ok_hand: instead of create. I have 2 remarks:
1. don't you get `field` method is redefined warning by doing this? You might want to use `redefine_method` function to prevent these warnings.
2. what is the return value of `add_default_value()` function. if super is not the last call, the `field` function of quantity and others will behave differently (unless `add_default_value` returns the whole object). Can you verify this please?

models/relation.rb
```diff
@@ -43,16 +43,18 @@ class Relation < Base
...
+      # Raise condition required due to the use of lib/helper deep_copy
+      raise 'linked_services should not belong to vrp hash' if hash.key?(:linked_services)
 
       # TODO: remove it after the linked_ids is replaced with linked_service_ids in the api definition
       if hash.key?(:linked_ids)
         hash[:linked_service_ids] = hash[:linked_ids]
       elsif hash.key?(:linked_service_ids)
         hash[:linked_ids] = hash[:linked_service_ids]
-      elsif hash.key?(:linked_services)
-        hash[:linked_ids] = hash[:linked_services].map(&:id)
```
senhalil, line 66: This was to make sure that until we finally fix the `linked_ids` vs `linked_vehicle_ids` disperency -- i.e., firmly pass to `linked_service_ids` -- the model would stay inter-operable. What if I just create the relation with `linked_services` -- i.e., with the proper objects without re-generation then we need to set the linked_ids manually, no ?

models/relation.rb
```diff
@@ -43,16 +43,18 @@ class Relation < Base
...
     def self.create(hash)
       # TODO: remove it after the linked_ids is replaced with linked_service_ids in the api definition
-      exclusive = [:linked_service_ids, :linked_ids, :linked_services].freeze
-      raise "#{exclusive} fields are mutually exclusive" if hash.keys.count{ |k| exclusive.include? k } > 1
+      if hash.key?(:linked_service_ids) && hash.key?(:linked_ids) && hash[:linked_service_ids] != hash[:linked_ids]
+        raise 'linked_service_ids and linked_ids should have the same content if both used'
+      end
+
+      # Raise condition required due to the use of lib/helper deep_copy
+      raise 'linked_services should not belong to vrp hash' if hash.key?(:linked_services)
```
senhalil: Why do we need this? As far as I remember, deep_copy should be able to handle linked_services/linked_service_ids/linked_ids correctly.

senhalil: Note that it is possible to create a vehicle object with a hash which has `start_point_id` (pointing to a real point object with this id`) but also with a hash which has `start_point` (which points to a real point object directly and creator of vehicle will set the `start_point_id` correctly using this proper point object). So this is the same thing with respect to `linked_services` vs `linked_service_ids`, no?

senhalil: The problem araises if you supply both `start_point_id` and `start_point` and they reference to two diffrent objects. I believe in this case one of them overrides the other but I am not sure which one (it might be order dependent) but overall I think `linked_service_ids` and `linked_services` should work exactly the same I think. 

Except that we have this extra `linked_ids` to maintain :face_exhaling: but that shouldn't make us change the behavior of another field of the model.

senhalil: (if we can help it)

wrappers/wrapper.rb
```diff
@@ -981,13 +981,10 @@ def simplify_complex_multi_pickup_or_delivery_shipments(vrp, solution = nil, opt
...
             }
 
-            new_shipment_relation = Helper.deep_copy(
-              relation,
-              override: {
-                linked_ids: nil, # this line can be removed when linked_ids is replaced with linked_service_ids
-                linked_services: relation.linked_services.map{ |s| s.id == service.id ? new_service : s }
-              }
-            )
+            relation_hash = relation.as_json
```
senhalil, line 986: Seeing `as_json` being useful makes me happy :hugs:

senhalil (commented):
Last crinkles

lib/interpreters/split_clustering.rb
braktar: yes !

models/quantity.rb
braktar: 1. Where should i see these warnings ?

2. `field` method is not expected to return any value

models/relation.rb
braktar: I tried to comply with the previous logic in place. But for relations, we did not use deep_copy anymore. I
 could erase this raise if you think it is not useful anymore.

models/relation.rb
braktar: In case we generate a new relation from a hash. Which field is prime ? `:linked_services` ?

models/relation.rb
senhalil: If we erase the raise, we will ignore the `linked_services` silently, no? If it is not going to be ignored we can erase it for sure.


Even if we do not use deep_copy we can have linked_services and linked_service_ids at the same time, we just need to figure out which one has the higher "priority" -- using start_point vs start_point_id in a vehicle.create() -- and then let linked_ids get overwritten via the correct info.

I feel like the easiest would be just get rid of linked_ids from the model and just do the linked_ids -> linked_service_ids "swap" before the first vrp.create  (at the same time we do shipment -> service swap) and call it a day. What do you think? 

That way we can mark linked_ids depreciated and add mutually_exclusive linked_service_ids (which is more inline with linked_vehicle_ids) so that we can completely remove it a v0.2?

senhalil: This would save us the relation_hash.delete(:linked_ids) logic and we can completely remove this custom logic of relation.create.

models/relation.rb
senhalil: I don't know, it depends how it is handled by active_hash. For example we do nothing splecial for vehicle.start_point -- what happens if the has has start_point and start_point_id pointing two diffrent objects. I suspect the one that comes later overwrites the first one in `hash.each{ |key, value| .... }` loop

models/relation.rb
braktar: Get rid of linked_ids is way more ambitious and requires at least its own dedicated PR. As there is the whole shipment logic to adapt : https://github.com/Mapotempo/optimizer-api/blob/062360a7b59a1e30e9e74596c12828434d93d107/test/models/vrp_test.rb#L150

deep_copy of relations has been replaced by .as_json which did not dump the services but only the linked_service_ids. So even if we want to directly assign services, i assume it should be done outside/after the create.

models/quantity.rb
senhalil: 1. when you launch a test with an instance with a quantity for example? If there is no method redefinition warning, are you sure this method is called? Can you put a byebug and try?
2. your version of the field returns a value which is specifically `false`. I don't know what active_hash returns but my expectation is that we return the same thing and not a random false.

models/relation.rb
senhalil: I didn't quite get the shipment part and why it is ambitious. We can do the linked_ids -> linked_service_ids replacement after the shipments turned into services but before anything else. 

Is it not as easy as "find_all linked_ids, replace_all with linked_service_ids" inside the code-code and a "ensure_compatibility" kind of logic that does the swap in vrp.rb (and a coercer in tsphelper) plus a mutually_exclusive `linked_service_ids` in the vrp_input.rb? 

We are trying to converge to a solution for these linked_services/linked_service_ids/linked_ids since multiple days, and what I am trying to prevent is having this extra reassignment/delete and custom create logic. These are there all because I didn't replace all linked_ids with linked_service_ids in the first place and they are a clear pain point.

braktar: My point is that it turns a +58 -20 9 files PR into a +199 -165 18 files for a change which was not mandatory for the point addressed. Anyway, it is now done

models/quantity.rb
braktar: 1. There is now warning about the redefinition of `field`. But you are right this method is never called. But the behavior is correct without it. The default getter returns the default value `false`, and it doesn't returns the value within as_json as it is not part of the attributes
2. active_hash didn't return a particular value for method field. Its last step is to create a function : https://github.com/active-hash/active_hash/blob/1c95f992af3ec94c07e19846f042e12bd0b11dd1/lib/active_hash/base.rb#L224

models/service.rb
```diff
@@ -62,5 +62,13 @@ class Service < Base
...
     has_many :quantities, class_name: 'Models::Quantity'
     has_many :relations, class_name: 'Models::Relation', as_json: :none
+
+    def self.create(hash, _options = {})
+      hash[:skills] = [] if hash[:skills].to_a.empty?
+      hash[:skills].map!(&:to_sym)
+      hash[:skills].sort!
+
+      super(hash)
+    end
```
senhalil: I think the reason why we don't have a default value for skills is the fact that we simply do not set a default value for `skills` -- before we had `field :skills, default: []` but this got removed by Model::Solution dev (if I am not mistaken). 

Do you remember why you removed the defaults? 

If we save us coming back to having this create logic just because of this default value, it would be better. I believe with the new convert/type logic if we have a `default: []` everything should work. What do you think?

senhalil:
Can we merge [Linked_ids and linked_service_ids are currently most of the time dupl…](https://github.com/Mapotempo/optimizer-api/pull/349/commits/aa6dd1b131ee03e1ff6ebd1f9b990a000f369fde) and  [Get rid of linked_ids](https://github.com/Mapotempo/optimizer-api/pull/349/commits/1cd69c21399f86c79e1f3f9f2ea62f4af0ada937)? The last one removes/modifies most of everything done by the first one.

models/service.rb
braktar: `default: []` give the very same object `[]` to every service with no skills

senhalil: We can continue here to have the trace. 

I cannot reproduce the issue on local at one before the last commit of this PR. For example, the following simple test doesn't generate duplicate objects but unique `[]` objects `service.skills`:

```ruby
def test_basic
  vrp = TestHelper.create(VRP.basic)
  puts vrp[:services].map{|s| s.skills.object_id }
end
```

This should have generated a single `[]` skills object for all services if the missing `create` was the problem, right?

senhalil:
@braktar The following bug still active on this branch unfortunately 
`[2022-03-17T18:17:53.073031172+01:00] 1b369389f94ec1e252bcb9fbf605c5e5 - FATAL: ActiveHash::RecordNotFound: Couldn't find Models::Rest with ID="r190465419"`

braktar:
`Couldn't find Models::Rest with ID="r190465419"` is now fixed

models/vrp.rb
```diff
@@ -383,7 +382,8 @@ def self.convert_expected_string_to_symbol(hash)
       }
 
       hash[:services].each{ |service|
-        service[:skills] = service[:skills]&.map(&:to_sym)
+        service[:skills] ||= [] # Service skills should always be initialized
```
senhalil: I can't understand why we need this `service[:skills] ||= []` and `vehicle[:skills] = [[]]`? Is it for skipping the `service.has_key?(:skill)` logic? The initialization of a member should not be obligatory (or handled here), the service class can initialize its own members itself. Am I missing something?

braktar: It has currently no need. It is mostly defined as a warning for subsequent dev we could do to filter/edit skills

senhalil: I was concerned with the behavior change but I guess this is safe and your explanation is :ok_hand:

models/service.rb
```diff
@@ -64,7 +64,7 @@ class Service < Base
     has_many :relations, class_name: 'Models::Relation', as_json: :none
 
     def self.create(hash, _options = {})
-      hash[:skills] = [] if hash[:skills].to_a.empty?
+      hash[:skills] = [] if hash[:skills].to_a.empty? # Service skills should always be initialized
```
senhalil: I am looking into why we need these

braktar:
`Each mission_ids should refer to an existant id` should be fixed

senhalil (approved):

PR #350, opened 2022-02-14
Setup duration is not a travel time

braktar:


models/solution/stop.rb
```diff
@@ -66,6 +66,7 @@ def vrp_result(options = {})
         hash.merge!(info.vrp_result(options))
         hash['point_id'] = self.activity.point_id
         hash['detail'] = hash.delete('activity')
+        hash['detail']['setup_duration'] = 0 if info.travel_time == 0
```
senhalil, line 69: (Probably a bit out of scope of this dev!) Do we have to show setup_duration even if it doesn't exist in the original problem (sent by the user)? Because having a "not_applied" setup duration (because this is not the first activity at the location) and "not having" any setup duration returns the same result detail at the moment.

senhalil (commented):
Just to be sure: We need `setup_duration_is_not_travel_time` because Or-tools includes/treats setup as travel right? 

If that is the case, wouldn't it be better to handle this "correction" inside ortools.rb? 

My gut feeling is that it would be better like that because I suspect vroom will not include setup_duration in travel_time (but i haven't checked, do you know?)

braktar:
I've been thinking a bit further on it.
The route stops refers to service activities. Whenever we use vroom or ortools, we have operation which can happen afterwars and use, for example, insert_stop, so, if we set the setup duration in a intermediate step we could lose some information.
I think more reasonable to only set the applied setup duration during the construction of the returned solution. What do you think ?

senhalil:
> we have operation which can happen afterwars and use, for example, insert_stop, so, if we set the setup duration in a intermediate step we could lose some information.

That's true! Then we find a way to make the behavior uniform -- i.e., solver can be vroom/ortools/etc and we want to be able to insert/pop-out stops from the routes.

It looks like to me that the easiest way still, have the setup_duration as info in the stop (and not within the travel time) because that way if we pop-in/out some stop we need to verify the correctness of the setup-duration information of the previous and next stops. If this info is embedded within the travel_time, we still need to check the previous and next stops but we also need to fix the travel_time. Does that make sense?

models/solution/stop.rb
```diff
@@ -60,6 +60,10 @@ def initialize(object, options = {})
       end
 
       def vrp_result(options = {})
+        if info.travel_time&.positive?
```
senhalil, line 63: If I understood it correctly, you moved the setup_duration "parsing"/"correction" to the very end so that we don't have to worry about correcting (applying, rewinding, etc) this info. Is that it?

braktar: Exact !

wrappers/wrapper.rb
```diff
@@ -1421,11 +1418,12 @@ def simplify_service_setup_duration_and_vehicle_setup_modifiers(vrp, solution =
 
             next if setup_duration.zero?
 
-            time_correction = coef_setup * setup_duration + additional_setup
+            # The vehicle adjustement is performed by vrp_result
+            stop.activity.setup_duration = setup_duration
+            travel_time_correction = stop.activity.setup_duration_on(vehicle)
```
senhalil, line 1423: Just above there is `setup_duration = stop.activity[:simplified_setup_duration].to_i` I don't know why I did the rounding there but it should be done here in fact (because the matrix is modified as such so rewind/patch should of the inverse):

```suggestion
            travel_time_correction = stop.activity.setup_duration_on(vehicle).to_i
```

senhalil (commented):
I think accept a "bug" that comes from my original implementation, it is :rocket: Do you mind fixing it?

senhalil (approved):

PR #353, opened 2022-02-14
Marshal dump repetitions

braktar:


models/base.rb
```diff
@@ -60,6 +60,17 @@ def vrp_result_fields
       end
     end
 
+    def duplicate_safe(options = {})
+      object_hash = JSON.parse(self.to_json, symbolize_names: true)
+
+      options.each{ |key, value|
+        next unless object_hash.key?(key)
+
+        object_hash[key] = value
```
senhalil, line 69: If we decide to not to dump the keys with default values this will lead to an issue, we need to check if `key` exists as a filed or a member function via the original object `self` itself.

models/configuration.rb
```diff
@@ -80,7 +80,7 @@ class Resolution < Base
     field :batch_heuristic, default: false
     field :repetition, default: nil
 
-    def self.create(hash)
+    def self.create(hash, _options = {})
```
senhalil, line 83: Why do we need these ignored `_options` arguments?

models/vrp.rb
```diff
@@ -516,8 +516,8 @@ def self.deduce_unavailable_days(hash, element, start_index, end_index, type)
     def self.deduce_possible_days(hash, element, start_index)
       convert_possible_dates_into_indices(element, hash, start_index)
 
-      element[:first_possible_days] = element[:first_possible_day_indices].to_a.slice(0..(element[:visits_number] || 1) - 1)
-      element[:last_possible_days] = element[:last_possible_day_indices].to_a.slice(0..(element[:visits_number] || 1) - 1)
+      element[:first_possible_days] ||= element[:first_possible_day_indices].to_a.slice(0..(element[:visits_number] || 1) - 1)
+      element[:last_possible_days] ||= element[:last_possible_day_indices].to_a.slice(0..(element[:visits_number] || 1) - 1)
```
senhalil, line 520: Is this related to [Marshal dump repetitions](https://github.com/Mapotempo/optimizer-api/pull/353/files#)? It looks like a proper bug fix. (Not asking for you to separate by the way, just keep it in mind for the future that if the fix is separate, it doesn't wait for the independent refacto and we can merge it faster)

lib/interpreters/compute_several_solutions.rb
```diff
@@ -18,6 +18,16 @@
 
 module Interpreters
   class SeveralSolutions
+    def self.duplicate_service_vrp(service_vrp, vrp_hash = nil)
+      service_vrp.map{ |key, value|
+        if key == :vrp
+          vrp_hash ||= JSON.parse(value.to_json, symbolize_names: true)
```
senhalil, line 24: Why don't we use `vrp.as_json` instead of `JSON.parse(vrp.to_json, symbolize_names: true)`? I think `as_json` already returns a new hash object with each call and we can make sure that `as_json` returns proper "symbolized" keys, no? What do you think? That way, we skip two intermediate parse operation in `hash->to_json->parse->hash` -- depending on the size of the VRP, this might take time and memory.

models/concerns/validate_data.rb
```diff
@@ -420,7 +420,7 @@ def check_maximum_lateness(periodic_heuristic)
 
     if @hash[:services].any?{ |service|
       ([service[:activity]] + service[:activities].to_a).compact.any?{ |activity|
-        activity[:timewindows]&.any?{ |tw| tw[:maximum_lateness] }
+        activity[:late_multiplier]&.positive? && activity[:timewindows]&.any?{ |tw| tw[:maximum_lateness] }
```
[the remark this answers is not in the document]
braktar: Until now, if we had late_multiplier to 0, it was meaning no lateness, did you change this behavior ?

models/vrp.rb
senhalil: What if I dump the vrp via `as_json`, make a change (like modify `visits_number`) and then call `vrp.create` this would leave the incorrect `:first_possible_day_indices` in the new object. As long as we don't have a way to mark an object "dirty/clean" I am afraid the best way is to recalculate it -- considering the fact that it is a very rudimentary calculation, it won't have any performance effects. What do you say?

lib/interpreters/compute_several_solutions.rb
braktar: ruby as_json returns explicitly string keys : https://apidock.com/rails/ActiveModel/Serializers/JSON/as_json 
I probably need to introduce an option for our use case

senhalil: I am not sure if to_json minds if the hash has string or symbol keys but I just realised, we implemented the wrong function. We should have implemented `to_hash` (which calls `as_json` at the moment but we can delete that) instead of `as_json` because that is our intention -- our object is a glorified hash and it can be represented as a hash -- i..e, to_hash (which is different then `to_h`).

If we rename the `as_json` to `to_hash` (and remove the other `to_hash`) and have the hash have `:symbol` keys then the native `Hash::as_json` function will do the necessary for handling the string business and we can use `to_hash` for dump/load logic skipping the json logic (which is useful for file dumps). What do you say?
 
https://justincypret.com/blog/activesupport-as-json-vs-to-json#adapt-objects-to-hash

models/base.rb
```diff
@@ -74,11 +74,27 @@ def inspect
...
+
+      options.each{ |key, value|
+        next unless object_hash.key?(key) && self.class.default_attributes[key] != value
+
+        object_hash[key] = value
+      }
+      self.class.create(object_hash, options)
+    end
+
     def as_json(options = {})
```
senhalil, line 88: Noting so that we don't forget: Implementing `to_hash` might be even better than `as_json` so that we use the default `hash.as/to_json` functionality.

senhalil (changes requested):

models/base.rb
```diff
@@ -74,11 +74,27 @@ def inspect
...
+    def duplicate_safe(options = {})
+      object_hash = self.as_json
+
+      options.each{ |key, value|
+        next unless object_hash.key?(key) && self.class.default_attributes[key] != value
+
+        object_hash[key] = value
+      }
+      self.class.create(object_hash, options)
+    end
```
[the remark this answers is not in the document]
braktar: Seems to be an artifact :x:

models/base.rb
senhalil: This doesn't work as I expected. We decided to left it aside.

models/concerns/as_json.rb
```diff
@@ -20,9 +20,9 @@ module VrpAsJson
...
     vrp = {
-      'name' => self.name,
-      'configuration' => self.configuration.as_json(options),
-      'points' => [], 'services' => [], 'vehicles' => []
+      name: self.name,
+      configuration: self.configuration.as_json(options),
+      points: [], services: [], vehicles: []
     }
 
     vrp.to_json
```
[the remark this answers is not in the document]
braktar: the hash -> call to_json -> call Json.parse -> hash logic as been removed now. 

having empty_vrp within vrp class sounds a bit odd. Having it related to as_json in a concern does not shock me.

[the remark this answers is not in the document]
senhalil: Hmm, you are right about the oddness but having this custom logic in `as_json` is quite odd as well. Plus, in either case, this is not an empty json nor an empty vrp. Its name is generic (empty_json) but if we include the concern anywhere else (i.e., vehicles, services) it would not work. 

Instead of this, I suggest we save us the trouble and just do the following which prevents any confusion and actually less code (1 line). Since it is used only in one location, as long as we don't need this logic at +10x location we are better off.

```ruby
Models::VRP.create({ name:  original.name, configuration: original.configuration.as_json }) 
```
I assume the points, services, vehicles, etc will be automatically initialized to empty arrays, right?

models/vrp.rb
```diff
@@ -57,6 +55,10 @@ def self.create(hash, options = {})
       }.merge(hash[:configuration] || {})
       Models.delete_all if options[:delete]
 
+      hash[:vehicles] ||= []
+      hash[:services] ||= []
+      hash[:points] ||= []
```
senhalil, line 60: I thought fields of type `has_many` are automatically initialized with an empty array? Is that not the case?

lib/interpreters/compute_several_solutions.rb
```diff
@@ -18,6 +18,16 @@
...
 module Interpreters
   class SeveralSolutions
+    def self.duplicate_service_vrp(service_vrp, vrp_hash = nil)
+      service_vrp.map{ |key, value|
+        if key == :vrp
+          vrp_hash ||= value.as_json
+          value = Models::Vrp.create(vrp_hash)
+        end
+        [key, value]
+      }.to_h
```
senhalil, line 28: I find this more expressive because of explicit `dup` and override of `[:vrp]`, what do you think?
```suggestion
      vrp_hash ||= service_vrp[:vrp].as_json
      service_vrp.dup.tap{ |hash| hash[:vrp] = Models::Vrp.create(vrp_hash) }
```

models/base.rb
```diff
@@ -77,7 +77,11 @@ def inspect
     def as_json(options = {})
       hash = {}
       self.class.json_fields.each{ |field_name|
-        hash[field_name] = self.send(field_name).as_json if self.respond_to?(field_name)
+        next unless self.respond_to?(field_name) &&
+                    (self.class.default_attributes.nil? ||
+                     self.class.default_attributes[field_name] != self.send(field_name))
```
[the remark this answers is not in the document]
braktar: I couldn't agree more on `unless` !

models/vrp.rb
braktar: The problem comes from the multiple validation/filter/transformations

senhalil (approved):
After rebase :rocket:

models/vrp.rb
senhalil: Oh yeah, the filters are based on the fact that grape validates that there are at least these members. I agree the easiest is what you already suggested :+1:

PR #354, opened 2022-02-14
Empty solution for periodic generates duplicate id error

braktar:


models/vrp.rb
```diff
@@ -80,6 +80,7 @@ def expand_vehicles_for_consistent_empty_result
     end
 
     def empty_solution(solver, unassigned_with_reason = [], already_expanded = true)
+      Models.delete_all # TODO: Remove once periodic heuristic use the proper object creation instead of duplicate_safe
```
senhalil, line 83: Is this safe? Can't we fix the `duplicate_safe` of periodic so that it doesn't lead to objects with duplicate IDs? I suspect we might be able to fix it pretty easily the way we did it for `Helper.deep_copy` https://github.com/Mapotempo/optimizer-api/blob/1ef4a82756fe7484fe8b6eb69439476d6592af8a/lib/helper.rb#L65

What do you think? I find it unsafe to need to reset all objects mid-run :fearful:

senhalil: Ah :( Why do you think it start to be an issue now? Because now we have proper ID validation in `Model::Solution` ?

braktar: Yes, that's the reason. Previously the solution was not validated at all. Now, the solution create active_hash json with ids, specially for services and rests.

senhalil: This made me think of something that we haven't discussed before, with "repetition" or "several_solutions" (or even with dicho etc) we can have a service (stop) appear in multiple solutions during the same optimization. If I am not mistaken, the new Model::Solution class will block us from doing so, no? How are we gonna handle this? 

I think it is good to have the ID validation within a single Solution object (or more completely, within the objects of all solution from the same "level"  -but this is more complex-), but if this prevents us from having the same service appearing in two solution then we have much deeper issue then `empty_solution`, right?

braktar: A `Models.delete_all` is performed right after each resolution by define_main_process (https://github.com/Mapotempo/optimizer-api/blob/1ef4a82756fe7484fe8b6eb69439476d6592af8a/optimizer_wrapper.rb#L104) so it is not a issue for repetition/several_solutions

senhalil: Then we keep this in mind as you noted in the `TODO` and continue our lives :)

senhalil (approved):

PR #355, opened 2022-02-14
Remove tmp folder

braktar:


senhalil (commented):
@braktar did you verify that this folder is for sure not used? I have a `mapotempo-optimizer-api` fodler inside `tmp` but I don't remember creating it, it must have been generated by the code at some point. I guess it must be okay since the CI (which is based on docker image) passes but wanted to check with you before merging.

braktar:
The config refers to the system tmp, local tmp is never used

senhalil (approved):

PR #356, opened 2022-02-14
Let synchronous resolution return a geojson

braktar:


senhalil (changes requested):
There is at least one more `ENV['OPTIM_GENERATE_GEOJSON_POLYLINES']` to be replaced with `OptimizerWrapper.config[:restitution][:allow_polylines]` -- it should be in `vrp_input.rb`. 

@braktar can you verify that there is none left after that one is replaced and "fix-up" the main commit `Let synchronous resolution return a geojson` with this correction then it is :rocket:

senhalil (approved):

PR #359, opened 2022-02-14
Remove cost field

braktar:


senhalil (commented):
Question and a minor renaming; otherwise great, we don't need to keep updating the cost anymore.

senhalil (approved):
The [first attempt](https://github.com/Mapotempo/optimizer-api/actions/runs/1841953858/attempts/1) failled with time-out. I don't see why this PR would do that, so I relaunched the tests. If the CI pass, we can merge it :rocket:

PR #362, opened 2022-02-14
Update avancement

braktar:


optimizer_wrapper.rb
```diff
@@ -87,7 +89,8 @@ def self.define_main_process(services_vrps, job = nil, &block)
     several_service_vrps = Interpreters::SeveralSolutions.expand_similar_resolutions(services_vrps)
     several_solutions = several_service_vrps.collect.with_index{ |current_service_vrps, solution_index|
       callback_main = lambda { |wrapper, avancement, total, message, cost = nil, time = nil, solution = nil|
-        msg = "#{"solution: #{solution_index + 1}/#{several_service_vrps.size} - " if several_service_vrps.size > 1}#{message}" unless message.nil?
+        msg = message && several_service_vrps.size > 1 && "solution: #{solution_index + 1}\
+                                                          /#{several_service_vrps.size} - #{message}" || message
```
senhalil, line 93:
```suggestion
        msg =
          if message && several_service_vrps.size > 1
            "solution: #{solution_index + 1}/#{several_service_vrps.size} - #{message}"
          else
            message
          end
```

optimizer_wrapper.rb
```diff
@@ -98,7 +101,8 @@ def self.define_main_process(services_vrps, job = nil, &block)
 
         service_vrp_repeats.each_with_index{ |repeated_service_vrp, repetition_index|
           repeated_results << define_process(repeated_service_vrp, job) { |wrapper, avancement, total, message, cost, time, solution|
-            msg = "#{"repetition #{repetition_index + 1}/#{service_vrp_repeats.size} - " if service_vrp_repeats.size > 1}#{message}" unless message.nil?
+            msg = message && service_vrp_repeats.size > 1 && "repetition #{repetition_index + 1}" \
+                                                             "/#{service_vrp_repeats.size} - #{message}" || message
```
senhalil, line 105: We have exact same logic above, maybe we should turn it into a function?

lib/interpreters/split_clustering.rb
```diff
@@ -257,14 +260,23 @@ def self.split_solve_core(service_vrp, job = nil, &block)
         # RECURSIVELY CALL split_solve_core AND MERGE SOLUTIONS
         solutions = sides.collect.with_index{ |side, index|
           service_vrp[:split_level] = split_level + 1
+          service_vrp[:split_denominators] << 2**service_vrp[:split_level]
+          service_vrp[:split_sides] << index
```
senhalil, line 264: I remember discussing basing this on the number of services or the optimization duration, was that going to happen in another dev?

lib/interpreters/split_clustering.rb
```diff
@@ -257,14 +260,23 @@ def self.split_solve_core(service_vrp, job = nil, &block)
         # RECURSIVELY CALL split_solve_core AND MERGE SOLUTIONS
         solutions = sides.collect.with_index{ |side, index|
           service_vrp[:split_level] = split_level + 1
+          service_vrp[:split_denominators] << 2**service_vrp[:split_level]
```
senhalil, line 263: I am not sure if this would work. `split_solve` is a recursive function and `split_level` increases, decreases, re-increases many times during the optimization process (if my memory servers me well) so this would make the bar  jump back and fort in a big optimization with multiple levels. Plus some branches dive "deeper" (more levels) than some other branches so the denominator calculation is not correct :( (The problem is even worse for dicho)

Lets base the calculation on services_seen(planned+unplanned)/total_services, this has multiple advantages:
- it is correct from the time point of view
- it never jumps back and fort
- the same logic works for dicho, split, repeats, several_solutions, etc.

And instead of saying `Handling sub-process X of Y`, we can say `X services out of Y are handled`. 

Note that the first one gives no information what so ever about the "completion" level -- it might say 1 out of 10 subproblems solved -- i.e., 10% but we might have 100X more time to go or may be we have completed half of the optimization by solving this first sub-problem.

senhalil (commented):
I am not sure this works as expected, can we discuss this again?

lib/interpreters/split_clustering.rb
senhalil: @braktar do we need to discuss this subject?

braktar: The way I implement the avancement using the denominators and the sides didn't not make the bar jump back and forth.

I would like to keep the current logic based on the number of sub_problems processed and keep the progression by planned for the next rework.

Still, there is some issues with the current implementation I need to fix these first before further discussions and review :slightly_smiling_face:

lib/interpreters/split_clustering.rb
braktar: Yes, it will be done in another dev

lib/interpreters/split_clustering.rb
senhalil: Deal :handshake:  ping me when you want me to review

senhalil: Also for the bar to jump back and fort you need a big optimisation so you might want to try with (i) either a big file (ii) or a reasonably sized file with a very small max_split_size..

lib/interpreters/split_clustering.rb
```diff
@@ -278,10 +278,10 @@ def self.split_solve_core(service_vrp, job = nil, &block)
             }.sum
 
             msg =
-              if message.include?('max split')
+              if message.include?('recursive split process')
                 message
               else
-                add = "max split #{(service_vrp[:split_denominators].last * avc).to_i}/#{service_vrp[:split_denominators].last}"
+                add = "recursive split process #{(service_vrp[:split_denominators].last * avc).to_i}/#{service_vrp[:split_denominators].last}"
```
senhalil: Can we do `split solve process` that way we can keep the split_solve name that we have/know/love :heart:

senhalil (approved):
I think calculating the advancement with dicho denominator might be misleading -> https://github.com/Mapotempo/optimizer-api/pull/362#discussion_r809880529

PR #363, opened 2022-02-15
Schedule vehicle name

braktar:


senhalil:
Closed automatically because of `Fix #363` note in the commit

lib/interpreters/periodic_visits.rb
```diff
@@ -186,15 +186,24 @@ def generate_services(vrp)
...
-        global_day_index: vehicle_day_index,
-        skills: associate_skills(vehicle, vehicle_day_index),
-        rests: generate_rests(vehicle, vehicle_day_index, rests_durations),
-        sequence_timewindows: [],
-      )
-      @equivalent_vehicles[vehicle.id] << new_vehicle.id
+      new_vehicle_hash = JSON.parse(vehicle.to_json, symbolize_names: true)
+      new_vehicle_hash.delete(:id)
+      new_vehicle_hash.delete(:start_point_id)
+      new_vehicle_hash.delete(:end_point_id)
```
senhalil, line 192: May be we can have something like `except: [:id, :start_point_id, :end_point_id, :sequence_timewindows]` in as_json -- similar to  https://apidock.com/rails/ActiveModel/Serializers/JSON/as_json

I am not sure if this is automatic. Since we are implementing the `as_json` function, I guess we need to implement it via  `options[:except]`. What do you think is it feasible? It would save us from deleting fields manually.

test/lib/heuristics/periodic_test.rb
```diff
@@ -243,10 +243,11 @@ def test_returned_vehicles_ids_with_two_stage_cluster
...
       problem[:configuration][:schedule] = { range_indices: { start: 0, end: 83 } }
 
-      solutions = OptimizerWrapper.wrapper_vrp('demo', { services: { vrp: [:demo] }},
-                                               TestHelper.load_vrp(self, problem: problem), nil)
+      solutions = OptimizerWrapper.wrapper_vrp(
+        'demo', { services: { vrp: [:demo] }}, TestHelper.load_vrp(self, problem: problem), nil
+      ).vrp_result
       assert solutions[0]
-      route_vehicle_ids = solutions[0].routes.collect{ |route| route.vehicle.id }
+      route_vehicle_ids = solutions[0]['routes'].collect{ |route| route['vehicle_id'] }
```
senhalil, line 250: Why do we revert from method to string hash? Issue with rebase?

senhalil (changes requested):
It looks like this PR needs a rebase and I have some suggestions, tell me what you think.

test/lib/heuristics/periodic_test.rb
braktar: Seems to be an artifact from a rebase

models/service.rb
```diff
@@ -63,6 +63,7 @@ class Service < Base
 
     def self.create(hash, _options = {})
       hash[:skills] = [] if hash[:skills].to_a.empty?
+      hash[:skills].map!(&:to_sym)
```
senhalil, line 66: It looks like there are quite a few other fields/parameters which have symbol type in `vrp_input.rb` and because of as_json/to_json they will loose their type after dump/load :/ 

Can you do a similar `:to_sym` call for these fields in this commit please -- then we can rename the commit to something like "Prevent as_json transforming symbols into strings"? 

Hopefully there will not be more members in this situation but it looks like we might want automatize this à-la-façon-de `as_json: :id` in base.rb -- like `type: symbol` that way the call to `:to_sym` would be done automatically in `initialize()`.

models/base.rb
```diff
@@ -76,7 +76,8 @@ def inspect
     def as_json(options = {})
       hash = {}
       self.class.json_fields.each{ |field_name|
-        next if !respond_to?(field_name) ||
+        next if options[:except]&.include?(field_name.to_sym) ||
```
senhalil, line 79: Can't we remove the `.to_s` part from `json_fields << name.to_s` so that we don't have to do `field_name.to_sym`? Similarly for `vrp_result_fields`?

models/activity.rb
```diff
@@ -38,6 +38,11 @@ class Activity < Base
                                   # the code would continue to accept invalid time windows thorugh API because
                                   # vrp.valid? doesn't call the validator of activity
                                   # We need to implement a check inside Api::V01::Vrp and fix the ActivityTest::test_timewindows accordingly
+    def self.create(hash)
+      hash[:position]&.to_sym
```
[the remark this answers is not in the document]
braktar: I will put the `type: X` idea in a future dev

models/vehicle.rb
```diff
@@ -106,7 +106,7 @@ class Vehicle < Base
 
     def vrp_result(_options = {})
       {
-        vehicle_id: self.id,
+        vehicle_id: self.original_id.to_s + (self.global_day_index ? "_#{self.global_day_index}" : ''),
```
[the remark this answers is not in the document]
braktar: The actual id may have nothing related to the original id anymore to avoid  any `id` conflict. We use the active_hash self defined ids.

wrappers/ortools.rb
```diff
@@ -281,14 +281,14 @@ def solve(vrp, job, thread_proc = nil, &block)
       vrp.relations.each{ |relation|
         relation.split_regarding_lapses.each{ |portion_linked_ids, portion_vehicle_ids, portion_lapse|
           current_linked_ids = (portion_linked_ids.map!(&:to_s) & services.map(&:id)).uniq if portion_linked_ids
-          current_linked_vehicles = (portion_vehicle_ids.map!(&:to_s) & vehicles.map(&:id)).uniq if portion_vehicle_ids
+          current_linked_vehicles = vrp.vehicles.select{ |v| portion_vehicle_ids.include?(v.id) }.map(&:id) if portion_vehicle_ids
```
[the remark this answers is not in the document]
braktar: It was forcing the type of vehicle id to string, which is integer when using the active hash self defined ids

lib/interpreters/periodic_visits.rb
```diff
@@ -212,6 +216,9 @@ def build_vehicle(vrp, vehicle, vehicle_day_index, rests_durations)
 
     def generate_vehicles(vrp)
       rests_durations = Array.new(vrp.vehicles.size, 0)
+
+      # Repopulate Models::Unit
+      vrp.units.each{ |unit| Models::Unit.insert(unit) } if Models::Unit.all.empty?
```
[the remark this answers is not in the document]
braktar: Ok ! The problem comes from the tests and not the code base ;)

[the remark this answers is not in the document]
braktar: Ok, the main problem comes from the fact that load_vrps load multiple vrps, but make a delete_all before loading any new vrp. So whenever there is actually multiple vrp to load, the objects refered into the firsts are erased

lib/interpreters/split_clustering.rb
```diff
@@ -1008,12 +1005,18 @@ def self.duplicate_vehicle(vehicle, timewindow, schedule)
       available_days.collect{ |day|
         next unless (schedule[:start]..schedule[:end]).any?{ |day_index| day_index % 7 == day }
 
-        tw = timewindow ? Marshal.load(Marshal.dump(timewindow)) : Models::Timewindow.create({})
+        tw = Models::Timewindow.create(timewindow&.as_json&.symbolize_keys || {})
```
senhalil, line 1008: `nil.as_json => nil` and keys are already symbol

```suggestion
        tw = Models::Timewindow.create(timewindow.as_json || {})
```

lib/interpreters/periodic_visits.rb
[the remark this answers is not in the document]
senhalil: Ah :face_exhaling:  yeah, `load_vrps` unfortunately does something that cannot be done via API, can't we 
1. move the `load_vrps` to a `yield/block` type of logic so that, we create the model objects just before the test  (instead of creating all before hand within load_vrps)
2. or at least, do the model::object.inser on the test size via a test_helper function that takes the vrp object and "populates" the model database?
What do you say, easy/hard?

models/service.rb
senhalil: There is apparently a few `type: Date` and  `type: DateTime` as well. And we need to keep in mind "solution" side of things -- for example,  there is `field :schedule_start_date` in `solution_configuration.rb` which comes from the original `Configuration::Schedule` so we need to apply the `type: Date` to both of them

lib/interpreters/periodic_visits.rb
senhalil: @braktar you already did this, right? In  case you kept it alive, I didn't resolve it but if that is not hte case you can resolve it.

senhalil (approved):
I think we need to find a general solution to the following logic but otherwise :rocket: 
```ruby
vrp.X.each{ |i| Models::X.insert(i) } if Models::X.all.empty?
```
@braktar if you have any ideas, how we can do that lets discuss it on a ticket?

PR #364, opened 2022-02-18
Pd perf

braktar:
Pré-requis : 

- [x] https://github.com/Mapotempo/optimizer-api/pull/363

### DIFF
--- a/models/vrp.rb
+++ b/models/vrp.rb
@@ -77,10 +77,10 @@ module Models
     end
 
     def self.as_json(options = {})
-      # Remove unknown linked_service_ids from relations
       hash = super(options)
 
       service_ids = hash[:services].map{ |service| service[:id] }
+      # Remove unknown linked_service_ids from relations
       hash[:relations].delete_if{ |relation|
         relation[:linked_service_ids].delete_if{ |id| service_ids.exclude?(id) }
 
--- a/models/vrp.rb
+++ b/models/vrp.rb
@@ -76,6 +76,19 @@ module Models
       vrp
     end
 
+    def self.as_json(options = {})
+      # Remove unknown linked_service_ids from relations
+      hash = super(options)
+
+      service_ids = hash[:services].map{ |service| service[:id] }
+      hash[:relations].delete_if{ |relation|
+        relation[:linked_service_ids].delete_if{ |id| service_ids.exclude?(id) }
+
+        relation[:linked_service_ids].empty?
+      }
+      hash
+    end
+
     def expand_vehicles_for_consistent_empty_result
       periodic = Interpreters::PeriodicVisits.new(self)
       periodic.generate_vehicles(self)
--- a/test/models/service_test.rb
+++ b/test/models/service_test.rb
@@ -18,9 +18,8 @@
 require './test/test_helper'
 
 module Models
-  class ServiceTest < Minitest::Test
+  class ServiceTest < IsolatedTest
     def test_skills
-      Models.delete_all
       service1 = { id: 'service_1' }
       service2 = { id: 'service_2' }
 
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -55,6 +55,8 @@ Minitest::Retry.use!(
 )
 
 class IsolatedTest < Minitest::Test
+  # It cleans the active_hash database before and after (around) the test. 
+  # Needed for tests calling Model::X.create directly to prevent ID errors.
   include Rack::Test::Methods
 
   def around
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -54,6 +54,17 @@ Minitest::Retry.use!(
   ]
 )
 
+class IsolatedTest < Minitest::Test
+  include Rack::Test::Methods
+
+  def around
+    Models.delete_all
+    yield
+  ensure
+    Models.delete_all
+  end
+end
+
 module TestHelper # rubocop: disable Style/CommentedKeyword, Lint/RedundantCopDisableDirective, Metrics/ModuleLength
   def self.coerce(vrp)
     # This function is called both with a JSON and Models::Vrp
--- a/wrappers/wrapper.rb
+++ b/wrappers/wrapper.rb
@@ -471,12 +471,19 @@ module Wrappers
 
       unless service_already_marked_unfeasible
         service.relations.each{ |relation|
-          next unless Models::Relation::ALL_OR_NONE_RELATIONS.include?(relation.type.to_sym) # TODO: remove to_sym when https://github.com/Mapotempo/optimizer-api/pull/145 is merged
-
-          relation.linked_services&.each{ |service_in|
+          remove_start_index =
+            case relation.type
+            when *Models::Relation::ALL_OR_NONE_RELATIONS
+              0
+            when *Models::Relation::POSITION_TYPES
+              relation.linked_service_ids.index(service.id)
+            else
+              next
+            end
+          remove_end_index = -1
+          relation.linked_services[remove_start_index..remove_end_index].each{ |service_in|
             next if service_in == service
-
-            add_unassigned(unfeasible, vrp, service_in, "In a relation with an unfeasible service: #{service.id}")
+            add_unassigned(unfeasible, vrp, service_in, "In a #{relation.type} relation with an unfeasible service: #{service.id}")
           }
         }
         vrp.routes.each{ |route|
@@ -604,8 +611,6 @@ module Wrappers
           end
         end
       }
-
-      relation.linked_services -= to_delete_services
 
       to_delete_services.each{ |service|
         add_unassigned(unfeasible, vrp, service, "Service belongs to a relation of type #{relation.type} which makes it infeasible")
--- a/wrappers/wrapper.rb
+++ b/wrappers/wrapper.rb
@@ -573,6 +573,88 @@ module Wrappers
       unfeasible
     end
 
+    def sequence_relation_with_no_reachable_vehicle?(unfeasible, vrp, relation)
+      return false unless Models::Relation::POSITION_TYPES.include?(relation.type)
+
+      vrp.vehicles.each{ |v|
+        unreachable_service_id = last_service_reachable_sequence(vrp, v, relation.linked_services)
+        next unless unreachable_service_id
+
+        service_met = false
+        relation.linked_services.each{ |service|
+          service_met = true if service.id == unreachable_service_id
+          next unless service_met
+
+          service.vehicle_compatibility[v.id] = false
+        }
+      }
+
+      previous_removed = false
+      to_delete_services = []
+      relation.linked_services.each{ |service|
+        previous_removed ||= vrp.vehicles.none?{ |vehicle|
+          service.vehicle_compatibility[vehicle.id]
+        }
+
+        if previous_removed
+          if Models::Relation::ALL_OR_NONE_RELATIONS.include?(relation.type)
+            to_delete_services = relation.linked_services
+            break false
+          else
+            to_delete_services << service
+          end
+        end
+      }
+
+      relation.linked_services -= to_delete_services
+
+      to_delete_services.each{ |service|
+        add_unassigned(unfeasible, vrp, service, "Service belongs to a relation of type #{relation.type} which makes it infeasible")
+      }
+      relation.linked_services.empty?
+    end
+
+    def last_service_reachable_sequence(vrp, vehicle, services)
+      vehicle_timewindow = vehicle.timewindow || Models::Timewindow.new(start: 0)
+      vehicle_duration = [vehicle.duration, vehicle_timewindow.safe_end(vehicle.cost_late_multiplier&.positive?) - vehicle_timewindow.start].compact.min
+      successive_activities = services.map{ |service|
+        [service.id, service.activity && [service.activity] || service.activities]
+      }
+      return nil if successive_activities.all?{ |_id, acts| acts.any?{ |a| a.timewindows.none?(&:end) } }
+
+      matrix = vrp.matrices.find{ |m| m.id == vehicle.matrix_id }
+      successive_activities = services.map{ |service|
+        [service.id, service.activity && [service.activity] || service.activities]
+      }
+
+      depot_approach = vehicle.start_point && matrix.time ? successive_activities.first.last.map{ |act| matrix.time[vehicle.start_point.matrix_index][act.point.matrix_index] }.min : 0
+      earliest_arrival = vehicle_timewindow.start + depot_approach
+
+      successive_activities.each.with_index{ |(id, a_activities), a_index|
+        return id if earliest_arrival > (a_activities.map{ |act| act.timewindows.map{ |tw| tw.safe_end(act.late_multiplier&.positive?) }.max }.max ||
+                                            vehicle_timewindow.safe_end(vehicle.cost_late_multiplier&.positive?))
+
+        last_duration = a_activities.map{ |a| a.duration_on(vehicle) }.min
+        depot_return = vehicle.end_point && matrix.time ? a_activities.map{ |act| matrix.time[act.point.matrix_index][vehicle.end_point.matrix_index] }.min : 0
+        earliest_depot_arrival = earliest_arrival + depot_return + last_duration
+
+        return id if earliest_depot_arrival > vehicle_timewindow.safe_end(vehicle.cost_late_multiplier&.positive?) ||
+                     earliest_depot_arrival - vehicle_timewindow.start > vehicle_duration
+
+        break if a_index == successive_activities.size - 1
+
+        _b_id, b_activities = successive_activities[a_index + 1]
+
+        earliest_arrival = a_activities.map{ |a_act|
+          b_activities.map{ |b_act|
+            travel_time = matrix.time && matrix.time[a_act.point.matrix_index][b_act.point.matrix_index]
+            travel_time += a_act.duration_on(vehicle) + travel_time > 0 ? a_act.setup_duration_on(vehicle) : 0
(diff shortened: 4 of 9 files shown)


wrappers/wrapper.rb
```diff
@@ -712,6 +771,15 @@ def check_distances(vrp, unfeasible)
       unfeasible = check_unreachable(vrp, :distance, unfeasible)
       unfeasible = check_unreachable(vrp, :value, unfeasible)
 
+      vrp.relations.delete_if{ |relation|
+        next unless relation_with_no_reachable_vehicle(vrp, relation)
+
+        relation.linked_services.each{ |service|
+          add_unassigned_internal(unfeasible, vrp, service, "Activities of the relation #{relation.type} cannot be reached successively by any vehicle")
```
senhalil: add_unassigned

wrappers/wrapper.rb
```diff
@@ -570,6 +571,64 @@ def detect_unfeasible_services(vrp)
       unfeasible
     end
 
+    def relation_with_no_reachable_vehicle(vrp, relation)
+      return false unless Models::Relation::POSITION_TYPES.include?(relation.type)
+
+      vrp.vehicles.each{ |v|
```
senhalil, line 585: service.compatible_vehicles

senhalil: for all services in `relation.linked_services`

wrappers/wrapper.rb
```diff
@@ -570,6 +571,64 @@ def detect_unfeasible_services(vrp)
...
+        next if reachable_service_sequence?(vrp, v, relation.linked_services)
+
+        relation.linked_services.each{ |service|
+          service.vehicle_compatibility[v.id] = false
+        }
+      }
+
+      relation.linked_services.any?{ |service|
+        vrp.vehicles.none?{ |v| service.vehicle_compatibility[v.id] }
+      }
```
senhalil: We need to check if there is no vehicle that can serve all linked_service_ids.

test/models/service_test.rb
```diff
@@ -19,8 +19,14 @@
 
 module Models
   class ServiceTest < Minitest::Test
-    def test_skills
+    def around
+      Models.delete_all
+      yield
+    ensure
       Models.delete_all
+    end
```
senhalil: If we need them here, don't we need this in every test?

wrappers/wrapper.rb
```diff
@@ -572,6 +573,66 @@ def detect_unfeasible_services(vrp)
...
 
+    def relation_with_no_reachable_vehicle(vrp, relation)
+      return false unless Models::Relation::POSITION_TYPES.include?(relation.type)
+
+      vrp.vehicles.each{ |v|
+        next if reachable_service_sequence?(vrp, v, relation.linked_services)
+
+        relation.linked_services.each{ |service|
+          service.vehicle_compatibility[v.id] = false
+        }
```
senhalil: In theory/practice, we can plan the beginning of an order/sequence relation and skip the tail of it. Here we completely remove that possibility? Is it voluntary?

senhalil (commented):
I think we are doing this backwards, first marking/removing relations and then marking all its services infeasible. 

I think we should 
1- First accumulate "vehicle_incompatibility" information of services to the services that are coming after them. 
2- Then add_unassinged services which are deemed infeasible and 
3- Only then filter the services from relations and remove the relations which left empty at the end.

What do you think about this @braktar ?

Plus I think we should do this relation check at the very very end after we already eliminated vehicle_compatibility with respect to every other check so that if a relation is not possible because services do not have a "matching" compatible vehicle we can detect it

test/models/service_test.rb
braktar: Most of other tests use Models::Vrp.create which clear by himself the Models store.

We usually only need this for model tests. Otherwise we need a custom class which inherits from Minitest::Test and from all our test classes needs to inherit.

We can also only inherit from this custom class for models tests. 

What do you think ?

wrappers/wrapper.rb
```diff
@@ -478,6 +478,7 @@ def add_unassigned(unfeasible, vrp, service, reason)
 
             add_unassigned(unfeasible, vrp, service_in, "In a relation with an unfeasible service: #{service.id}")
           }
+          vrp.relations.delete(relation)
```
senhalil: Why do we delete a relation if there is an infeasible service in it? There might be other feasible services in it, no?

senhalil: :face_exhaling: :facepalm: ALL_OR_NONE_RELATIONS

braktar: we are in the context of a `ALL_OR_NONE_RELATIONS` so it requires to delete all the services and the relation as well otherwise we have a relation without the corresponding services in the problem

test/models/service_test.rb
senhalil: If the other tests are okay, we can leave it like this at the moment but I like your inheritance idea only for the group of tests who need it.

test/test_helper.rb
```diff
@@ -54,6 +54,17 @@
   ]
 )
 
+class IsolatedTest < Minitest::Test
```
senhalil, line 57: A little comment goes a long way :relaxed: 
```suggestion
class IsolatedTest < Minitest::Test
  # It cleans the active_hash database before and after (around) the test. 
  # Needed for tests calling Model::X.create directly to prevent ID errors.
```

wrappers/wrapper.rb
```diff
@@ -572,6 +573,88 @@ def detect_unfeasible_services(vrp)
...
+      vehicle_duration = [vehicle.duration, vehicle_timewindow.safe_end(vehicle.cost_late_multiplier&.positive?) - vehicle_timewindow.start].compact.min
+      successive_activities = services.map{ |service|
+        [service.id, service.activity && [service.activity] || service.activities]
+      }
+      return nil if successive_activities.all?{ |_id, acts| acts.any?{ |a| a.timewindows.none?(&:end) } }
+
+      matrix = vrp.matrices.find{ |m| m.id == vehicle.matrix_id }
+      successive_activities = services.map{ |service|
+        [service.id, service.activity && [service.activity] || service.activities]
+      }
```
senhalil: Duplicate code
```      
      successive_activities = services.map{ |service|
        [service.id, service.activity && [service.activity] || service.activities]
      }
```

wrappers/wrapper.rb
```diff
@@ -572,6 +573,88 @@ def detect_unfeasible_services(vrp)
...
+    def sequence_relation_with_no_reachable_vehicle?(unfeasible, vrp, relation)
+      return false unless Models::Relation::POSITION_TYPES.include?(relation.type)
+
+      vrp.vehicles.each{ |v|
+        unreachable_service_id = last_service_reachable_sequence(vrp, v, relation.linked_services)
+        next unless unreachable_service_id
+
+        service_met = false
+        relation.linked_services.each{ |service|
+          service_met = true if service.id == unreachable_service_id
```
senhalil, line 591: This condition should be as follows no? Otherwise, we mark only the unreachable_service but not the services "following" it. 
```suggestion
          service_met ||= true if service.id == unreachable_service_id
```

wrappers/wrapper.rb
```diff
@@ -572,6 +573,88 @@ def detect_unfeasible_services(vrp)
...
+          if Models::Relation::ALL_OR_NONE_RELATIONS.include?(relation.type)
+            to_delete_services = relation.linked_services
+            break false
+          else
+            to_delete_services << service
+          end
+        end
+      }
+
+      relation.linked_services -= to_delete_services
```
senhalil: I am not sure if we need to do this. The linked_service_ids/linked_services are parsed in ortools.rb so that only the ids/services that are present in the problem are passed to the solver.

senhalil: What happens if the sub-problem is merged with another sub-problem and now the services are feasible and the relation is feasible?

I think it is dangerous to modify the relations. 

I think we should be able to handle all this with `unassigned` logic which correctly handles the case of merged sub-problems.

wrappers/wrapper.rb
senhalil: If all services of linked_services of a relation are already marked infeasible and removed from the vrp.services list, do we really need to delete the relation? This is related to my last question, what happens to this relation becomes feasible after we merge the sub-problems with another sub-problem? Can't we get away without deleting/modifying the relations?

wrappers/wrapper.rb
```diff
@@ -572,6 +573,88 @@ def detect_unfeasible_services(vrp)
...
+
+      to_delete_services.each{ |service|
+        add_unassigned(unfeasible, vrp, service, "Service belongs to a relation of type #{relation.type} which makes it infeasible")
+      }
+      relation.linked_services.empty?
```
senhalil: no need to return this anymore right?

wrappers/wrapper.rb
senhalil: But also are we sure that this is the most elegant way to do this?

wrappers/wrapper.rb
```diff
@@ -572,6 +573,88 @@ def detect_unfeasible_services(vrp)
...
+        _b_id, b_activities = successive_activities[a_index + 1]
+
+        earliest_arrival = a_activities.map{ |a_act|
+          b_activities.map{ |b_act|
+            travel_time = matrix.time && matrix.time[a_act.point.matrix_index][b_act.point.matrix_index]
+            travel_time += a_act.duration_on(vehicle) + travel_time > 0 ? a_act.setup_duration_on(vehicle) : 0
+            [b_act.timewindows.map(&:start)&.min || 0, earliest_arrival + travel_time].max
+          }.min
+        }.min
+      }
```
senhalil, line 656: it looks like we duplicated all the checks here in an elementary way. can't we generalize the original checks to prevent duplication?

wrappers/wrapper.rb
```diff
@@ -713,6 +796,10 @@ def check_distances(vrp, unfeasible)
       unfeasible = check_unreachable(vrp, :time, unfeasible)
       unfeasible = check_unreachable(vrp, :distance, unfeasible)
       unfeasible = check_unreachable(vrp, :value, unfeasible)
+      vrp.relations.each{ |relation|
+        # If relation becomes empty it is removed by add_unassigned
+        sequence_relation_with_no_reachable_vehicle?(unfeasible, vrp, relation)
+      }
```
senhalil: It is better if this block comes after the next (vrp.services `no_compatible_vehicle_with_compatible_tw`) block, no? Otherwise, a service might get marked as unfeasible after the relation check and the services "following" this service won't get marked infeasible.

wrappers/wrapper.rb
braktar: The other solution is to entirely migrate activity to activities

wrappers/wrapper.rb
braktar: service_met stays true from the service with the unreachable_service_id on a given vehicle
It is reset to false only if we change the vehicle. `||=` may only avoid a re-affectation. It won't change the value.

wrappers/wrapper.rb
braktar: If services are removed from the vrp and put in the unassigned list, if we perform as vrp.as_json, the relation will contain non existing service_ids which will generate error within the Vrp.create

wrappers/wrapper.rb
senhalil: Decided to do the generalization in another ticket.

wrappers/wrapper.rb
senhalil: Decided to handle the relation elimination in `as_json` so that if we optimize the dumped vrp (`vrp.as_json`),  it will always give the same solution.

wrappers/wrapper.rb
senhalil: Decided to handle the relation elimination in as_json so that if we optimize the dumped vrp (vrp.as_json), it will always give the same solution.

models/vrp.rb
```diff
@@ -76,6 +76,19 @@ def self.create(hash, options = {})
       vrp
     end
 
+    def self.as_json(options = {})
+      # Remove unknown linked_service_ids from relations
+      hash = super(options)
+
+      service_ids = hash[:services].map{ |service| service[:id] }
+      hash[:relations].delete_if{ |relation|
```
senhalil:
```suggestion
      hash = super(options)

      service_ids = hash[:services].map{ |service| service[:id] }
      # Remove unknown linked_service_ids from relations
      hash[:relations].delete_if{ |relation|
```

wrappers/wrapper.rb
```diff
@@ -572,6 +572,85 @@ def detect_unfeasible_services(vrp)
...
+          if Models::Relation::ALL_OR_NONE_RELATIONS.include?(relation.type)
+            to_delete_services = relation.linked_services
+            break false
+          else
+            to_delete_services << service
+          end
+        end
+      }
+
+      relation.linked_services -= to_delete_services
```
senhalil: @braktar did you forget to remove this?

wrappers/wrapper.rb
```diff
@@ -572,6 +572,85 @@ def detect_unfeasible_services(vrp)
...
+
+        if previous_removed
+          if Models::Relation::ALL_OR_NONE_RELATIONS.include?(relation.type)
+            to_delete_services = relation.linked_services
+            break false
+          else
+            to_delete_services << service
+          end
+        end
+      }
```
senhalil, line 613: Why do we need this (especially the ALL_OR_NONE part)? This is already handled inside `add_unassinged` function if I am not missing something, right?

Plus I think we can handle the `POSITION_TYPES` there too.

Actual
```
    def add_unassigned(unfeasible, vrp, service, reason)
      # calls add_unassigned_internal for every service in an "ALL_OR_NONE_RELATION" with the service
      service_already_marked_unfeasible = !!unfeasible[service.id]

      unless service_already_marked_unfeasible && reason.start_with?('In a relation with an unfeasible service: ')
        add_unassigned_internal(unfeasible, vrp, service, reason)
      end

      unless service_already_marked_unfeasible
        service.relations.each{ |relation|
          next unless Models::Relation::ALL_OR_NONE_RELATIONS.include?(relation.type.to_sym) # TODO: remove to_sym when https://github.com/Mapotempo/optimizer-api/pull/145 is merged

          relation.linked_services&.each{ |service_in|
            next if service_in == service

            add_unassigned(unfeasible, vrp, service_in, "In a relation with an unfeasible service: #{service.id}")
          }
        }
        vrp.routes.each{ |route|
          route.mission_ids.delete_if{ |mission_id| mission_id == service.id }
        }
      end

      unfeasible
    end
```
untested proposition
```
    def add_unassigned(unfeasible, vrp, service, reason)
      # calls add_unassigned_internal for every service in an "ALL_OR_NONE_RELATION" with the service or for following services in a POSITION_TYPES relation
      service_already_marked_unfeasible = !!unfeasible[service.id]

      unless service_already_marked_unfeasible && reason.start_with?('In a relation with an unfeasible service: ')
        add_unassigned_internal(unfeasible, vrp, service, reason)
      end

      unless service_already_marked_unfeasible
        service.relations.each{ |relation|
          if Models::Relation::ALL_OR_NONE_RELATIONS.include?(relation.type)
            relation.linked_services&.each{ |service_in|
              next if service_in == service

              add_unassigned(unfeasible, vrp, service_in, "In a #{relation.type} relation with an unfeasible service: #{service.id}")
            }
          elsif Models::Relation::POSITION_TYPES.include?(relation.type)
            original_unfeasible_service_has_past = false
            relation.linked_services&.each{ |service_in|
              next if service_in != service

              if !original_unfeasible_service_has_past
                original_unfeasible_service_has_past = true
                next
              end

              add_unassigned(unfeasible, vrp, service_in, "In a #{relation.type} relation with an unfeasible service: #{service.id}")
            }
          end
        }
        vrp.routes.each{ |route|
          route.mission_ids.delete_if{ |mission_id| mission_id == service.id }
        }
      end

      unfeasible
    end
```

senhalil: Which made me realise that maybe we should move the logic below to the vrp.as_json as well, no? It is the same logic. We lose the connection between the solution to the original problem and to the dumped one after solve -- at the end of the day route cannot make a service infeasible so maybe we can let this slip or we can mention it in the ticket you created, up to you. 

```
        vrp.routes.each{ |route|
          route.mission_ids.delete_if{ |mission_id| mission_id == service.id }
        }
```

wrappers/wrapper.rb
```diff
@@ -572,6 +572,85 @@ def detect_unfeasible_services(vrp)
...
+    end
+
+    def last_service_reachable_sequence(vrp, vehicle, services)
+      vehicle_timewindow = vehicle.timewindow || Models::Timewindow.new(start: 0)
+      vehicle_duration = [vehicle.duration, vehicle_timewindow.safe_end(vehicle.cost_late_multiplier&.positive?) - vehicle_timewindow.start].compact.min
+      successive_activities = services.map{ |service|
+        return service.id if service.vehicle_compatibility[vehicle.id] == false
+        [service.id, service.activity && [service.activity] || service.activities]
+      }
+      return nil if successive_activities.all?{ |_id, acts| acts.any?{ |a| a.timewindows.none?(&:end) } }
```
senhalil, line 627: What about the case an order relation is infeasible due to vehicle capacity? We normally handle capacity violating services with detect_unfeasible. Don't we want to generalise all this logic to these ALL_OR_NOTHING or POSITION_TYPE relations. 

This may be out-of-scope of this PR and maybe better placed in the generalisation ticket -- can you share me the link or add this message to the ticket (unless you already have this in the ticket, grands-esprit) so that we keep it in mind during generalization.

senhalil (approved):

PR #367, opened 2022-02-28
Backport - Setup duration related to matrix index

braktar:


wrappers/wrapper.rb
```diff
@@ -1497,8 +1503,8 @@ def simplify_service_setup_duration_and_vehicle_setup_modifiers(vrp, result = ni
             # then the following logic needs to be updated. Basically we need to do each_with_index
             # and apply the setup duration increment to every pair except index == point.matrix_index
             # even if they were 0 in the first place.
-            matrix.time.each{ |row|
-              row[point.matrix_index] += (coef_setup * setup_duration + additional_setup).to_i if row[point.matrix_index] > 0
+            matrix.time.each.with_index{ |row, row_index|
```
senhalil, line 1506: Now that it became true, shall we delete the warning? 
```suggestion
            matrix.time.each_with_index{ |row, row_index|
```

wrappers/wrapper.rb
```diff
@@ -1531,8 +1537,8 @@ def simplify_service_setup_duration_and_vehicle_setup_modifiers(vrp, result = ni
             coef_setup = vehicle[:simplified_coef_setup] || 1
             additional_setup = vehicle[:simplified_additional_setup].to_i
 
-            matrix.time.each{ |row|
-              row[point.matrix_index] -= (coef_setup * setup_duration + additional_setup).to_i  if row[point.matrix_index] > 0
+            matrix.time.each.with_index{ |row, row_index|
```
senhalil, line 1540:
```suggestion
            matrix.time.each_with_index{ |row, row_index|
```

senhalil (approved):

senhalil (approved):

test/fixtures/minimum_duration_lapse_shipments.json
```diff
@@ -523,19 +502,19 @@
           ]
         },
         "delivery": {
-          "point_id": "point_7",
+          "point_id": "point_59",
```
senhalil, line 505: Do we need to dump this file?

test/real_cases_periodic_test.rb
```diff
@@ -162,7 +162,7 @@ def test_performance_13vl
 
       # voluntarily equal to watch evolution of periodic algorithm performance
       assert_equal expected, seen, 'Do not have the expected number of total visits'
-      assert_equal 301, unassigned_visits.sum, 'Do not have the expected number of unassigned visits'
+      assert_equal 285, unassigned_visits.sum, 'Do not have the expected number of unassigned visits'
```
senhalil, line 165: Any idea why we have such an improvement? Matrix?

wrappers/wrapper.rb
```diff
@@ -1572,18 +1578,22 @@ def simplify_service_setup_duration_and_vehicle_setup_modifiers(vrp, result = ni
...
+            next if activity[:service_id].nil?
 
             setup_duration = services_grouped_by_point_id[activity[:point_id]].first.activity[:simplified_setup_duration].to_i
 
             next if setup_duration.zero?
 
             time_correction = coef_setup * setup_duration + additional_setup
 
-            total_travel_time_correction += time_correction
+            total_travel_time_correction += time_correction.round
```
senhalil: We need to round only once at the end, otherwise we accumulate rounding errors and totals do not represent the reality.

wrappers/wrapper.rb
```diff
@@ -1537,8 +1543,8 @@ def simplify_service_setup_duration_and_vehicle_setup_modifiers(vrp, result = ni
             coef_setup = vehicle[:simplified_coef_setup] || 1
             additional_setup = vehicle[:simplified_additional_setup].to_i
 
-            matrix.time.each{ |row|
-              row[point.matrix_index] -= (coef_setup * setup_duration + additional_setup).to_i  if row[point.matrix_index] > 0
+            matrix.time.each_with_index{ |row, row_index|
+              row[point.matrix_index] -= (coef_setup * setup_duration + additional_setup).to_i if point.matrix_index != row_index
```
senhalil, line 1547: point_id can be different but matrix_ids can be the same (if we do collect the matrix via unique lat/lon pairs to prevent large matrices) since the rewind/patch are based on point_id, can we base this condition on point_id as well -- with a TODO note that this logic should be replicated via point_id on the optimizer-ortools as well (which is actually based on matrix_index.

wrappers/wrapper.rb
```diff
@@ -1557,7 +1563,7 @@ def simplify_service_setup_duration_and_vehicle_setup_modifiers(vrp, result = ni
       when :patch_result
         # patches the result
         # the travel_times need to be decreased and setup_duration need to be increased by
-        # (coef_setup * setup_duration + additional_setup) if setup_duration > 0 and travel_time > 0
+        # (coef_setup * setup_duration + additional_setup) if setup_duration > 0 and matrix_indices are different
```
senhalil, line 1566: point_ids ?

wrappers/wrapper.rb
```diff
@@ -1608,16 +1618,22 @@ def shift_route_times(route, shift_amount, shift_start_index = 0)
...
         activity[:end_time] += shift_amount if activity[:end_time]
-        activity[:waiting_time] -= [shift_amount, activity[:waiting_time]].min if activity[:waiting_time]
-        activity[:departure_time] += shift_amount if activity[:departure_time]
+
+        if activity[:waiting_time]
+          waiting_resorb = [activity[:waiting_time] - current_shift, 0].max
+          current_shift = [current_shift - waiting_resorb, 0].max
+          activity[:waiting_time] -= waiting_resorb
+        end
+        activity[:departure_time] += current_shift if activity[:departure_time]
```
senhalil: I couldn't understand here very well :man_shrugging: what is `waiting_resorb` and what do we accomplish here? The original idea (if I am not mistaken) shift everything with a fix amount, why do we need to handle the departure_time specifically if the begin_time and end_time are pushed `+shift_amount` (and the fact that `departure_time` and `end_time` are the same thing) then `departure_time` should be pushed equally right? 

I couldn't clearly see how changing the behavior of setup_duration (travel_time==0 --> point_id==point_id) effects this operation. :thinking:

wrappers/wrapper.rb
```diff
@@ -1572,18 +1578,22 @@ def simplify_service_setup_duration_and_vehicle_setup_modifiers(vrp, result = ni
...
 
             next if setup_duration.zero?
 
             time_correction = coef_setup * setup_duration + additional_setup
 
-            total_travel_time_correction += time_correction
+            total_travel_time_correction += time_correction.round
             activity[:detail][:setup_duration] = time_correction.round
-            activity[:travel_time] -= activity[:detail][:setup_duration]
+            activity[:travel_time] -= time_correction.round
```
senhalil, line 1596: Any reason for this modification? it is the same function and it connects it to the setup_duration which already set to the same value the line above `activity[:detail][:setup_duration] = time_correction.round`

wrappers/wrapper.rb
senhalil: it looks like you did based it on the point_id in most places, we can make it uniform with that TODO -- or better a ticket in the current milestone so that we can  finilize the optim-orttols side before this interim/urgent dev arrives to a stable version. What do you think?

wrappers/wrapper.rb
```diff
@@ -1481,7 +1484,10 @@ def simplify_service_setup_duration_and_vehicle_setup_modifiers(vrp, result = ni
                       } || vrp.services.any?{ |s| s.activity.nil? }
 
         vehicles_grouped_by_matrix_id = vrp.vehicles.group_by(&:matrix_id)
-        vrp.services.group_by{ |s| s.activity.point }.each{ |point, service_group|
+        grouped_services = vrp.services.group_by{ |s| s.activity.point }
+        return nil if grouped_services.map{ |point, _s_g| point.matrix_index }.uniq.size != grouped_services.size
```
senhalil, line 1488: We should add a TODO (and mention this location in the ticket that I just mentioned) so that we remove this logic once the optim-ortools is based on point_id as well.

wrappers/wrapper.rb
senhalil: maybe https://github.com/Mapotempo/optimizer-api/pull/369 would have been a better place to discuss this but I didn't know you already have that PR

senhalil: Can we do all the review here and you can mirror the changes to #369 so that we don't double the review afford? Is that okay for you?

wrappers/wrapper.rb
braktar: The idea is that if waiting_time > 0. Then this waiting time will absorb as much as possible this shift so the next steps won't have to be shifted of the entire shift.

It is not really related to the change of travel_time == 0 to point_id == point_id, but as I changed the waiting time computation else where I had to fix it here too

wrappers/wrapper.rb
braktar: As we set time_correction to this very same value it was only to standardize its use

wrappers/wrapper.rb
senhalil: My confusion was because of my assumptions, on the other PR, I realized that you are using this function in other places so there can be waiting_time.

However, I still don't understand why we apply the `current_shift` "logic" (i.e., "absorb" corrected shift) only to `departure_time` but not also to `end_time`

Also, I think we should correct the `[:start]` with `current_shift` as well, no?

senhalil: :point_up:  this might be a bug :thinking:

senhalil (changes requested):

wrappers/wrapper.rb
braktar: Yes ! I have fixed this logic in #369 but didn't transpose it here. Nice catch !

senhalil (approved):

PR #369, opened 2022-03-02
Setup duration related to matrix index

braktar:


models/solution/parsers/route_parser.rb
```diff
@@ -150,24 +156,35 @@ def self.compute_route_waiting_times
...
+          extra_time = stop.info.begin_time - previous_end - used_travel_time
+          # setup_duration consumed by the current rest is at most the next stop setup_duration or
+          # the extra_time between this stop and the previous one
+          considered_setup = [extra_time, potential_setup].min
+          consumed_setup_time += considered_setup
         else
-          used_travel_time = (stop.info.travel_time || 0) - consumed_travel_time - consumed_setup_time
+          potential_setup = previous_activity&.point_id != stop.activity.point_id &&
+                            stop.activity.setup_duration_on(@route.vehicle) || 0
+          used_travel_time = (stop.info.travel_time || 0) - consumed_travel_time #- consumed_setup_time
```
senhalil: Did you want to comment this?

models/solution/parsers/route_parser.rb
```diff
@@ -150,24 +156,35 @@ def self.compute_route_waiting_times
...
+          considered_setup = [potential_setup - consumed_setup_time, 0].max
         end
-        considered_setup = stop.info.travel_time&.positive? &&
-                           (stop.activity.setup_duration.to_i - consumed_setup_time) || 0
-        arrival_time = previous_end + used_travel_time + considered_setup + consumed_setup_time
-        stop.info.waiting_time = stop.info.begin_time - arrival_time
+        arrival_time = previous_end + used_travel_time
+        stop.info.waiting_time = [stop.info.begin_time - (arrival_time + considered_setup), 0].max
+
+        consumed_setup_time = 0 unless stop.type == :rest
```
senhalil, line 185: I saw the comment but it is still not evident for me how this `consumed_setup_time` works :(

test/fixtures/minimum_duration_lapse_shipments.json
```diff
@@ -50,13 +50,6 @@
           "lon": -0.9744
         }
       },
-      {
```
senhalil, line 53: don't we need a redump for this file?

test/real_cases_periodic_test.rb
```diff
@@ -163,7 +163,7 @@ def test_performance_13vl
 
       # voluntarily equal to watch evolution of periodic algorithm performance
       assert_equal expected, seen, 'Do not have the expected number of total visits'
-      assert_equal 301, unassigned_visits.sum, 'Do not have the expected number of unassigned visits'
```
senhalil, line 166: any idea where the improvement comes from ?

wrappers/wrapper.rb
```diff
@@ -1338,8 +1343,8 @@ def simplify_service_setup_duration_and_vehicle_setup_modifiers(vrp, solution =
             # then the following logic needs to be updated. Basically we need to do each_with_index
             # and apply the setup duration increment to every pair except index == point.matrix_index
             # even if they were 0 in the first place.
-            matrix.time.each{ |row|
-              row[point.matrix_index] += first_activity.setup_duration_on(vehicle).to_i if row[point.matrix_index] > 0
+            matrix.time.each.with_index{ |row, row_index|
+              row[point.matrix_index] += first_activity.setup_duration_on(vehicle).to_i if point.matrix_index != row_index
```
senhalil, line 1343: Shall we add a comment so that here and only here we use the index

wrappers/wrapper.rb
```diff
@@ -1377,8 +1382,8 @@ def simplify_service_setup_duration_and_vehicle_setup_modifiers(vrp, solution =
             coef_setup = vehicle[:simplified_coef_setup] || 1
             additional_setup = vehicle[:simplified_additional_setup].to_i
 
-            matrix.time.each{ |row|
-              row[point.matrix_index] -= (coef_setup * setup_duration + additional_setup).to_i  if row[point.matrix_index] > 0
+            matrix.time.each_with_index{ |row, row_index|
+              row[point.matrix_index] -= (coef_setup * setup_duration + additional_setup).to_i if point.matrix_index != row_index
```
senhalil, line 1384: and here :)

senhalil (commented):
I think we are :pinching_hand: close to :rocket:

test/fixtures/minimum_duration_lapse_shipments.json
braktar: We have just removed some points and replaced them by their doppelganger. Do you see any particular reason to redump the file ?

test/real_cases_periodic_test.rb
braktar: My assumption is that before services at the same position but non related were performed in any case in the same row even if there was no particular reason to do so. Now there is no discrimination between these and their neighbors. So we can explore solutions with a more "global" perspective.

test/fixtures/minimum_duration_lapse_shipments.json
senhalil: No, if you think it is not needed, then it is not needed. I asked it automatically because the json was modified but there was no dump.

However, how does this work? I believe if you do not re-dump, the test will use the old dump which has the old ids and points etc  and we are not sure if th --- scratch that ---- the dumps are now ONLY matrix, so no problems right?

test/real_cases_periodic_test.rb
senhalil: This is great news!

test/fixtures/minimum_duration_lapse_shipments.json
braktar: The load of matrix dump also contain the association of the locations to the matrix indices. So I see no issue here :wink:

senhalil (approved):

PR #370, opened 2022-03-02
 Modiy empty/fill logic to fix current_load calculation

senhalil:


wrappers/ortools.rb
```diff
@@ -151,9 +151,9 @@ def solve(vrp, job, thread_proc = nil, &block)
               OrtoolsVrp::TimeWindow.new(start: tw.start, end: tw.end || 2147483647, maximum_lateness: tw.maximum_lateness)
             },
             quantities: vrp.units.collect{ |unit|
-              is_empty_unit = problem_units.find{ |unit_status| unit_status[:unit_id] == unit.id }[:empty]
+              is_fill_unit = problem_units.find{ |unit_status| unit_status[:unit_id] == unit.id }[:fill]
               q = service.quantities.find{ |quantity| quantity.unit == unit }
-              q&.value.to_f * (is_empty_unit ? -1 : 1)
+              q&.value.to_f * (q&.empty ? -1 : 1) * (is_fill_unit && !q&.fill ? -1 : 1)
```
braktar: I didn't get why we should change the sign of the quantities if we have a fill_unit.
The quantity values are assumed to be negative and the quantity provided with the fill positive

config/environments/development.rb
```diff
@@ -90,7 +90,7 @@ module OptimizerWrapper
       batch_heuristic: ENV['OPTIM_DBG_BATCH_HEURISTIC'] == 'true'
     },
     restitution: {
-      allow_polylines: ENV['OPTIM_GENERATE_GEOJSON_POLYLINES'] == 'true'
+      allow_polylines: ENV['OPTIM_GENERATE_GEOJSON_POLYLINES'] != 'false'
```
braktar, line 93: Why changing this condition and only this one ?

wrappers/ortools.rb
senhalil: I remember we said something like "its easier if the user always supplies positive quantities and we handle the negativity :)." Do I remember incorrectly?

The main reason is that this way the user is just obliged to mark one of the units fill and we will handle the rest. However, I am not sure if this is compatible with the case when we have  normal deliveries but no refilling :/ On the otherhand, we don't ask users to give negative quantities for deliveries and positive for collect (do we?) so I find it easier for the user to supply positive values and just mark the unit `fill: true` and we handle the rest. Otherwise the user is always free to suply negative values :thinking: 

Hmm, may be I can add a logic to make sure that the quantity.value of a `fill` unit is negative for all except the fill service itself and it is positive for the latter. How is that? You are right if the user has already multiplied the quantity, this might lead to unwanted behavior. 

This made me realize, (not in the PR but) why don't we do this "correction" at the model creation stage and assume that all the quantity and capacity values are correct and forget the empty/fill logic (expect for or-tools model, where we relax the slack variable) ?

config/environments/development.rb
senhalil: The others are debug outputs that are activated for debugging purposes but this is related to the vrp/json file itself. On local we never launch `polylines:true` or `polylines:[...]` unless we really need the geojson; however, if you forget to launch the server with `OPTIM_GENERATE_GEOJSON_POLYLINES` set then you get nothing -- this creates unnecessary friction.  This env var was originally created as a protection for the servers so I reverted on only local.

wrappers/ortools.rb
senhalil: @braktar I made the modification I suggested, do you have any other suggestion? This way ff the user did send the "good" values for services with empty/fill units no change but if not we multiply with -1 when necessary so that we make sure that if it is
- an empty unit then the amount is negative for the empty service and positive for the proper services
- a fill unit then the amount is positive for the fill service and negative for the proper services

braktar: Mapotempo Web consider delivery as positive values and collect with negative values.
On our side, We've defined the concepts the opposite. 

In the current case, I don't see the need to reverse the sign of quantity values (except for empty operations)
We may have some positive and negative quantity value (i.e: to provide waste material to reprocessing sites) within the same problem and fill XOR empty operations.

If we automatically reverse or values. Or expect only positive value, we should merge empty and fill fields.

So in the end, as discussed yesterday, indeed we can limit/reduce the scope until the need is met by a user :wink:

senhalil: Effectively, mt-web does the opposite of reality :face_exhaling:  but this change fortunately doesn't effect them :relieved: since they keep track of their quantities themselves.

I thought we were on the same page about the fill as on we are for empty. This is the limited scope, the part that is out of scope is moving this correction to the beginning of the API.

I don't understand how a positive amount on a service that has a fill "unit" make sense... if the fill operation itself has positive quantity and the proper services has positive quantity for this unit, then everything is a fill and there is no delivery and the current load just increases. It looks like the model works because of the relaxation of slack variables for the fill nodes but the current_load is calculated incorrectly.

We should not merge empty/fill since they are not the same actually and we need to keep them separate for easy of use on the api level but we can eliminate them in the worker/model level (if we do the correction of values at the api entrance/model creation) since they are exactly reverse of each other (in reality) and this dev is making sure that this is the case for the code as well and we model the actuality so that the vehicle loads make sense.

let me know if anything doesn't make sense, maybe I misunderstood something

braktar (approved):

PR #378, opened 2022-03-10
Move ortools relations tests & Add shipment soft mtwpdvrp

braktar:


senhalil:
@braktar I don't remember discussing this dev, why do we want to separate ortools_test.rb into separate files? I agree that the ortools_test.rb file is too long but the reason for that is too many "manual" VRPs and splitting the file in multiple doesn't fix the issue IMHO. At the moment we have at least ortools.rb -> ortools_test.rb connection and we can launch all related tests in one file. What do you say?

braktar:
My first wish here was to add a soft mtwpdvrp test for ortools
Then i realized that ortools_test was already huuuuge.
So I've extracted all the tests related to relations and started to minify a good part of them by moving to VRP.basic and other "common" vrps.

braktar:
Plus there is already a `ortools_multi_trips_test` file. Having the `ortools_` prefix to execute ortools test suit seems a good to me.

senhalil:
> there is already a `ortools_multi_trips_test` file. Having the `ortools_` prefix to execute ortools test suit seems a good to me.

`ortools_multi_trips_test` was created with a similar argument without proper oversight  (instead of fixing the issue itself) but be my guest if you want to base your decision on that.

> So I've extracted all the tests related to relations and started to minify a good part of them by moving to VRP.basic and other "common" vrps.

I think you did very good by minifying all the tests (this is the necessary fix) but you could do this modification in place which would have the same effect without scattered tests. 

---

Here are the stats of the analysis I did a while back (Jan 2021) :
Screenshot from 2022-03-25 17-50-51

1 - It is clear that GitLab doesn't split tests into multiple files and break the `code.rb <=> code_test.rb` logic even if a test file becomes thousands of lines.

2- We have a few test files with this issue (all created before VRP.basic/toy/etc utility was put in place and these files never got cleaned) if we clean them the way you did by simplifying them with the VRP.problem utility, I can see that ortools_test.rb, wrapper_test.rb and interpreter_test.rb will all be a lot smaller -- there are many +100-line tests with only 10-15 lines of actual content.  

Let me know what you think, I think we should even merge ortools_multi_trips_test.rb back into ortools.rb and keep the `code.rb <=> code_test.rb` logic and split a test file only if we are ready to split the class/code into multiple files as well. The length of the file changes literally nothing -- it does only decrease the chance of coincidental merge conflict chance which is the rarest and easiest to remedy (keep both sides) type of merge conflict.

senhalil:
is this related to this dev?
```
 =ERROR Models::VehicleTest#test_symbol_skills (5.39s)
Minitest::UnexpectedError:         ActiveHash::IdError: Duplicate ID found for record {:cost_fixed=>0, :cost_distance_multiplier=>0, :cost_time_multiplier=>1, :cost_value_multiplier=>0, :cost_setup_time_multiplier=>0, :coef_service=>1, :coef_setup=>1, :additional_service=>0, :additional_setup=>0, :router_mode=>:car, :router_dimension=>:time, :speed_multiplier=>1, :area=>[], :speed_multiplier_area=>[], :motorway=>true, :track=>true, :toll=>true, :max_walk_distance=>750, :shift_preference=>:minimize_span, :unavailable_days=>#<Set: {}>, :skills=>[[:string]], :skill_ids=>[], :original_skills=>[[:string]], :original_skill_ids=>[], :sequence_timewindows=>[], :sequence_timewindow_ids=>[], :capacities=>[], :capacity_ids=>[], :rests=>[], :rest_ids=>[], :id=>"vehicle_1", :original_id=>"vehicle_1"}
            /srv/app/vendor/bundle/ruby/2.5.0/bundler/gems/active_hash-10a84f020aad/lib/active_hash/base.rb:176:in `validate_unique_id'
            /srv/app/vendor/bundle/ruby/2.5.0/bundler/gems/active_hash-10a84f020aad/lib/active_hash/base.rb:183:in `create'
            /srv/app/models/vehicle.rb:124:in `create'
            /srv/app/test/models/vehicle_test.rb:121:in `test_symbol_skills'
            ```

senhalil (approved):
The broken test can not be related to this dev. Merging

PR #379, opened 2022-03-14
Wrong number of visits because of split_independent

senhalil:
Fixes two production bugs -- api-optimizer-prod/issues/19424/ and api-optimizer-prod/issues/19425

Closes https://gitlab.com/mapotempo/optimizer-api/-/issues/912

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -17,9 +17,6 @@ Style/FrozenStringLiteralComment:
     - 'Gemfile'
     - 'Rakefile'
 
-# Layout/LineLength:
-#   Enabled: false
-
 Style/NumericPredicate:
   Enabled: false
 
@@ -32,8 +29,11 @@ Layout/EmptyLines:
 Layout/ExtraSpacing:
   Severity: warning
 
+Layout/IndentationConsistency:
+  Severity: warning
+
 Layout/IndentationWidth:
   Severity: warning
 
-Layout/IndentationConsistency:
-  Severity: warning
+# Layout/LineLength:
+#   Enabled: false
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -42,6 +42,3 @@ Layout/IndentationConsistency:
 
 Layout/IndentationWidth:
   Severity: warning
-
-# Layout/LineLength:
-#   Enabled: false
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -17,6 +17,14 @@ Style/FrozenStringLiteralComment:
     - 'Gemfile'
     - 'Rakefile'
 
+# Disabled since it defers/pushes final verifications after a return which hides the verification logic.
+#
+# It is disabled by other tech companies with similar reasons:
+# https://github.com/airbnb/ruby/blob/12435e8136d2adf710de999bc0f6bef01215df2c/rubocop-airbnb/config/rubocop-style.yml#L363
+# https://github.com/Shopify/ruby-style-guide/blob/f604033657c54fa0dc28b5914c1e6c2641e6f361/rubocop.yml#L469
+Style/GuardClause:
+  Enabled: false
+
 Style/NumericPredicate:
   Enabled: false
 
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,4 @@ build/
 .byebug_history
 *.aof
 *.code-workspace
+TAGS
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@
 - Split duration among partitions correctly [#336](https://github.com/Mapotempo/optimizer-api/pull/336)
 - Fix find_best_heuristic selection logic [#337](https://github.com/Mapotempo/optimizer-api/pull/337)
 - Prevent periodic heuristic overwriting supplied initial routes [#318](https://github.com/Mapotempo/optimizer-api/pull/318)
+- Split independent VRP respects the skills of services in relations  [#379](https://github.com/Mapotempo/optimizer-api/pull/379)
 
 ## [v1.8.2] - 2022-01-19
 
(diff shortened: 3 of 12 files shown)


optimizer_wrapper.rb
```diff
@@ -400,6 +400,21 @@ def self.split_independent_vrp(vrp)
     # - there is multimodal subtours
     return [vrp] if (vrp.vehicles.size <= 1) || vrp.services.empty? || vrp.subtours&.any? # there might be zero services
 
+    # Accumulate the skills of services in a relation of ALL_OR_NONE_RELATIONS type
+    vrp.relations.each{ |relation|
+      next unless Models::Relation::ALL_OR_NONE_RELATIONS.include?(relation.type)
```
braktar: We did not exactly want to only address "all or none relations" but the relations which force services to be in the same route in way or another.
It includes : `same_route`, `sequence`, `order`, `shipment`

`meetup` expects different vehicles to reach the defined missions at the same time.

optimizer_wrapper.rb
```diff
@@ -400,6 +400,21 @@ def self.split_independent_vrp(vrp)
...
+
+      if Models::Relation::ON_SERVICES_TYPES.include?(relation.type)
```
braktar: Why is this a different condition than above ?

braktar (changes requested):

optimizer_wrapper.rb
senhalil: I am not sure we can do this skill accumulation trick vis-a-vis forcing relations as you suggested because (I remember asking you this a while back that) it is possible to plan only some part of an `order` relation as long as we respect the order but not plan some of the services -- isn't that the case anymore? Also, `same_route` relation applies only to planned services and so it is possible to plan a single one of them. But if we accumulate the skills, we might prevent this possibility.

However, for a `shipment`, we cannot just plan the pickup or just the delivery and they have to be on the same route -- so we don't change the feasible solution space by doing the skill accumulation. I believe it was a joint decision to include `sequence` in the all_or_none list during split_solve refacto -- even though we can skip the tail end of the services -- so I kept the same logic but it is not super appropriate to do so either.

I must have misunderstood `meetup` relation -- effectively if this is the definition, we cannot (and should not) accumulate the skills. It means that we need to find a way to "force" these services and the vehicles that can serve them to stay on the same subproblem (! which is easier said then done !) and keep the skill intact  :( 

In this case, seeing that we immediately need the shipment fix (and the other relations are not in high demand) and that we need a proper over-haul of the split_independent logic -- which is too big to do as a quick hotfix -- I suggest I do the skill accumulation fix only for `shipment` relation and at a "higher", more general location (because it is always correct to do this expansion) and create a ticket for the refacto of split_independent to handle all "forcing" relations in a generic fashion (the way we did for split_solve) in the next sprint. Is that good for you?

senhalil: In fact I just realised, split_independent is as high as it gets but I might do it in vrp.create after shipment expansion unless you have a suggestion for a better location?

optimizer_wrapper.rb
senhalil: We might add new "all_or_none" type relations which is on "vehicles" in the future and this makes sure that we only apply the skill accumulation to relations related to services and get alerted about the need for human intevention before the dev hits the servers.

optimizer_wrapper.rb
braktar: `same_route`, `order` and `sequence` have the same behavior  
mission at rank `n` should be assigned to allow mission at rank `n+1` to be active in the solution.
So for any mission, if it is assigned all the ancestors/predecessors should be assigned as well.

Do you expect that for a relation with services 1,2 and 3 we only assign services 1 and 2 and let service 3 unassigned eventually because of its skills ? I didn't imagine that at first. But we can transmit the skills from service 1 to service 2 and then from service 2 to service 3.

Have this propagation within vrp.create seems to be the good place

senhalil: Yes I know it is like that that's why they are skipped -- except sequence (because of the decision I mentioned).

I have high expectations :) that if we received pickup/delivery pairs with different skills then we will probably have services in order/sequence/same_route relations with different skills that might lead to this case and the user will "not understand why the service 1 and 2 are not planned" :face_exhaling:

senhalil: I like the idea of accumulating forwards -- that works for `shipment` too (it will not affect the split independent of shipment relations if we accumulate in only one direction)

senhalil: Then instead of doing it for `all_or_none` we can do it for all linking relations.

```ruby    # Relations that link multiple services to be on the same route
    LINKING_RELATIONS = %i[
      order
      same_route
      sequence
      shipment
      same_vehicle
    ].freeze
```

Does this list complete -- at least from the point of view of "forward skill accumulation"? 

I specifically asked from skill accumulation point of view because other relations for which we cannot do skill accumulation (e.g., `meetup`, `min/max_dur/day_lapse`) we still need a refacto

braktar: The list seems complete, i see no other need for a skills accumulation :+1:

.rubocop.yml
```diff
@@ -32,8 +37,11 @@ Layout/EmptyLines:
...
-Layout/IndentationWidth:
+Layout/IndentationConsistency:
   Severity: warning
 
-Layout/IndentationConsistency:
+Layout/IndentationWidth:
   Severity: warning
+
+# Layout/LineLength:
+#   Enabled: false
```
braktar: Are these extra comment lines necessary ?

senhalil: :shrug: I just fixed the alphabetical order -- assumed it is necessary since someone left it as a comment instead of removing it but we can remove it.

senhalil:
```suggestion
```

senhalil: Done

braktar (approved):

PR #381, opened 2022-03-25
Delete routes with empty mission_ids

braktar:


senhalil (commented):
Can we move the test to wrapper_test.rb since the logic is in wrapper.rb? Otherwise :rocket:

senhalil (approved):

PR #384, opened 2022-03-29
Simplify shouldn't raise in case of activities

braktar:


wrappers/wrapper.rb
```diff
@@ -1360,7 +1360,8 @@ def simplify_service_setup_duration_and_vehicle_setup_modifiers(vrp, solution =
         }
       when :rewind
         # take it back in case in dicho and there will be re-optimization
-        return nil unless vrp.services.any?{ |s| s.activity[:simplified_setup_duration] }
+        return nil if vrp.services.any?{ |s| s.activities.any? } ||
```
senhalil, line 1363: We need to do the same skip in `:patch_solution` stage as well, no?

test/wrapper_test.rb
```diff
@@ -2834,6 +2834,14 @@ def test_simplify_constraints_simplifies_pause
     refute vrp.vehicles.first.rests.none?, 'Should not have removed this rest because there is cost_late_multiplier'
   end
 
+  def test_simplify_activities_generates_no_raise
+    problem = VRP.toy
+    problem[:services].first[:activities] = [problem[:services].first.delete(:activity)] + [{ point_id: 'p1' }]
+
+    vrp = TestHelper.create(problem)
+    OptimizerWrapper.config[:services][:demo].simplify_constraints(vrp)
+  end
```
senhalil: The test doesn't test this fix I think. Are you sure it fails when you reverse the fix?

senhalil: For reiwnd you need to call `patch_and_rewind_simplified_constraints`

senhalil (changes requested):

wrappers/wrapper.rb
```diff
@@ -1360,7 +1360,8 @@ def simplify_service_setup_duration_and_vehicle_setup_modifiers(vrp, solution =
...
+                      vrp.services.none?{ |s| s.activity[:simplified_setup_duration] }
```
senhalil, line 1364: If I understand it correctly we don't need to modify `:patch_solution` at the moment because the services only have one activity in the solution already.. 

However, I am afraid this modification of the `:patch_solution` condition has introduced a bug during model::solution implementation.

Screenshot from 2022-03-31 11-04-04
   
The check was on `vrp.services` but it got moved to `solution.routes.any?.stops`. 

There are two problems if I am not mistaken:
1) If the solution is empty, we forget to patch the unassigned services
2) I am not even sure the `stop.activity` object can have the `simplified_setup_duration` member coming from the original service.activity automatically (unless they are the same exact object, but they are not, right? we do generate stop.activity). I just check and we do manually copy this info to the stop.activity in stop_parser but this violates the locality of the simplification function and adds an outside logic to stop_parser.rb.

I think we need move the `:patch_solution` verification to the same verification of `:rewind` -- i.e., 
`return nil if vrp.services.any?{ |s| s.activities.any? } || vrp.services.none?{ |s| s.activity[:simplified_setup_duration] }`. What do you say? If we do it in this commit `Simplify shouldn't raise in case of activities` I can fix the conflict during cherry-pick because on proid `:patch_solution` is still on the buggy if condition

wrappers/wrapper.rb
braktar: A solution stop only reference a single activity from the original service, the one which is performed

senhalil (approved):

PR #386, opened 2022-03-31
Various operators

braktar:


senhalil (changes requested):
Do you think the solver can return a solution in which the empty operation is first and there is a travel time between the depot and its location -- and there is a service/setup duration on the empty operation?

You wanted to make sure that this is the case (i.e. there is a travel/active setup time) and it is not happening because the empty operation is at the same location as the depot (i.e., no service and travel times). Are we sure it is not the latter?

Or-tools putting this empty operation to the very beginning doesn't make any sense if it degrades the solution quality this obviously -- cost-wise and lost-timewise.

----

Unfortunately, if this is the case, this means there is a possibility that there is service and travel time, then we need to fix the `shift_route_times` function in this PR because that function supposes that there are no time windows (since it was originally written for `simplify_vehicle_pause` and only used at places where there is no risk of shifting a service before its TW.begin or after its TW.end.

I feel like this is going to be a mess... but since we are deleting/removing a service here, it might be okay (inserting is very complicated), here we just need to improve the `shift_route_times` so that it has a similar mechanism to `waiting_resorb` but it is due to TW.begin (i.e., we shift the stops coming after the current one a bit less to the left if we hit the TW limit of the current stop). 

Does this make sense @braktar? Let me know if you prefer to attack this together given that we need to deploy it before Tuesday :grimacing:

models/solution/route.rb
```diff
@@ -77,6 +115,19 @@ def shift_route_times(shift_amount, shift_start_index = 0)
...
+      def shift_route_distance(shift_amount, shift_start_index = 0)
+        return if shift_amount == 0
+
+        raise 'Cannot shift the route, there are not enough stops' if shift_start_index > self.stops.size
+
+        self.info.total_distance += shift_amount
+        self.stops.each.with_index{ |stop, index|
+          next if index <= shift_start_index
+
+          stop.info.current_distance += shift_amount
```
senhalil: A check against too much shift?
```suggestion
          next if index <= shift_start_index

          raise 'Cannot shift the route, shift_amount is too small -- it would lead to negative current_distance` if shift_amount < -stop.info.current_distance

          stop.info.current_distance += shift_amount
```

models/solution/route.rb
```diff
@@ -55,6 +55,34 @@ def insert_stop(vrp, stop, index, idle_time = 0)
         shift_route_times(idle_time + stop.activity.duration, index)
       end
 
+      def remove_stop(vrp, stop, index)
```
senhalil: unsued parameter?
```suggestion
      def remove_stop(vrp, index)
```

lib/cleanse.rb
```diff
@@ -70,31 +66,41 @@ def self.same_fill_units(capacities, previous, current)
...
           true
         else
           previous_activity = stop if previous_activity.nil? || stop.service_id
           false
         end
       }
+
+      empty_index = get_empty_index_if_first_position(capacity_unit_ids, route.stops)
+      if empty_index
+        removed_stop = route.remove_stop(vrp, route.stops[empty_index], empty_index)
```
senhalil: unused function argument
```suggestion
        removed_stop = route.remove_stop(vrp, empty_index)
```

models/solution/route.rb
```diff
@@ -55,6 +55,34 @@ def insert_stop(vrp, stop, index, idle_time = 0)
...
+        removed_stop = stops.delete_at(index)
+        if index != 0 && index != stops.size
```
senhalil: Why don't we need to do any corrections if the deleted stop is the first or the last?

I think we need to at least correct, total_time and total_travel_time. Actually, we need to do correct every stop.info.current_time/distance etc as well.

In fact, we can use the same code with minor modifications. if the index is 0, then there is no previous_stop (i.e., `nil`) and when previous_stop is nil then `new_travel_time` is 0 and previous_end is removed_stop.info.begin_time (i.e., `previous_end = previous_stop&.info.end_time || removed_stop.info.begin_time`). A similar modification is needed for previous_matrix_index.

Similarly if removed_stop is the last stop, then we need to correct the totals, but instead of doing them manually we can apply the above logic and this time base it on the existence of  `next_stop`/`next_matrix_index` etc.

Let me know what you think.

models/solution/route.rb
```diff
@@ -55,6 +55,34 @@ def insert_stop(vrp, stop, index, idle_time = 0)
...
+          previous_stop = stops[index - 1]
+          next_stop = stops[index]
+
+          next_start = next_stop.info.begin_time
+          next_matrix_index = stops[index..-1].find{ |n_stop| n_stop.activity.point }&.activity&.point&.matrix_index
```
senhalil: Why can't we do following?
```suggestion
          next_matrix_index = next_stop.activity.point.matrix_index
```

models/solution/route.rb
```diff
@@ -69,6 +97,16 @@ def shift_route_times(shift_amount, shift_start_index = 0)
             waiting_resorb = [stop.info.waiting_time - current_shift, 0].max
             current_shift = [current_shift - waiting_resorb, 0].max
             stop.info.waiting_time -= waiting_resorb
+          else
+            active_tw = stop.active_timewindow
+            if active_tw
+              current_shift =
+                if shift_amount >= 0
+                  [current_shift, active_tw.safe_end(stop.activity.lateness_allowed?) - stop.info.begin_time].min
```
senhalil: We cannot do this, this is a `raise` situation because we already shift the previous services X amount to the right but here you realise you cannot shift the current stop X amount. But then the end of the previous stop and the beginning of the current stop would overlap, right? Or, did I miss something.

models/solution/route.rb
```diff
@@ -69,6 +97,16 @@ def shift_route_times(shift_amount, shift_start_index = 0)
...
+                else
+                  [current_shift, active_tw.start - stop.info.begin_time].max
```
senhalil: This review made me realize that with this "logic" we can lead to unnecessary idle time. Imagine a route without any idle time and we deleted a stop `s` -- so everything is shifted `s.duration` to the left. The first few services were able to shift to the left, okay but at some point, we hit a time window and shift amount became `0` -- i.e., we introduced intermediary idle times :clap: 

I don't have a solution.

As I said, when I first write this function, I only had the "no-TW" case in mind and it does the job for that case. However, when the time windows are involved we are not sure if it is better to shift everything following the deleted stop to the left or everything before it to the right or everything to the middle direction... I am afraid we are beating a dead horse. What do you think?

models/solution/stop.rb
```diff
@@ -94,6 +94,21 @@ def build_skills
...
+        if info.waiting_time > 0
+          active_tw = activity.timewindows.select{ |tw| tw.start > info.begin_time }.min_by{ |tw|
+            tw.start - info.begin_time
+          }
+          return active_tw
+        end
+
+        activity.timewindows.find{ |tw|
+          tw.start <= info.begin_time && info.begin_time <= tw.safe_end(activity.lateness_allowed?)
+        }
```
senhalil, line 115: It is better to check first without "lateness" first (and then check it with lateness only if the first doesn't find a TW) because with lateness activated, time windows might overlap and we can identify an incorrect time window.

models/solution/route.rb
braktar: The next may be a rest which have no point :/

models/solution/route.rb
braktar: You are correct !

models/solution/route.rb
braktar: I see two cases 

- Force start: We need to stick to the route begin_time
- Without force start, there is no need to shift the end of the route to the left, but to move the beginning of the route to the right

Perhaps shift route times and distance should take an interval of indices to update which gives implicitly the direction to update

models/solution/route.rb
senhalil: :disappointed_relieved: bien-vu

models/solution/route.rb
senhalil: Maybe we can pre-calculate how much actually we "can" shift and then shift everything that amount (i..e, we run the algo first without modifying the values and then use the final current_shift_amount to shift everyhign (and maybe raise a warning if the calculated maximum shift_amount is different than the original shift_amount) but I am not sure if we are on the right track with this function.

models/solution/route.rb
braktar: With the new case of removing the empty mission in first position, i can't be sure there is no timewindow :/

senhalil: I have no idea. This is getting more and more complicated. Maybe we should just delete the "empty" service and re-launch ortools? :D

senhalil: > Without force start, there is no need to shift the end of the route to the left, but to move the beginning of the route to the right

Even without force_start in the presence of tw.end, we might want to push left :/

senhalil: Maybe the best is to add as many raises/checks as necessary to this function (so that it is only used for the original no TW case, and for the erasuse of empty-service -- which is always the first stop).

senhalil: Do you have a concrete example/instance that we can launch to observe the use of unnecessary (and clearly sub-optimal) empty service as the first stop? Maybe we can fix that?

models/solution/route.rb
```diff
@@ -55,6 +55,46 @@ def insert_stop(vrp, stop, index, idle_time = 0)
         shift_route_times(idle_time + stop.activity.duration, index)
       end
 
+      def remove_stop(vrp, index)
+        removed_stop = stops.delete_at(index)
+        previous_stop = stops[0..index - 1].reverse.find{ |n_stop| n_stop.activity.point } if index != 0
+        next_stop = stops[index..-1].find{ |n_stop| n_stop.activity.point } if index != stops.size
+        passed_stop = false
+        non_located_stops = stops[0..index - 1].reverse.select{ |stop|
+          passed_stop = true if stop == removed_stop
```
senhalil: if stop is deleted from stops (`removed_stop = stops.delete_at(index)`)

how can it still be in the `stops` array (`if stop == removed_stop`)

models/solution/route.rb
```diff
@@ -55,6 +55,46 @@ def insert_stop(vrp, stop, index, idle_time = 0)
...
+        next_start = next_stop&.info&.begin_time || 0
+        previous_matrix_index = previous_stop&.activity&.point&.matrix_index
+        next_matrix_index = next_stop&.activity&.point&.matrix_index
+        previous_end = previous_stop.info.end_time
+        route_matrix = vrp.matrices.find{ |matrix| matrix.id == vehicle.matrix_id }
+        if route_matrix.time
+          new_travel_time =
+            previous_matrix_index && next_matrix_index &&
+            route_matrix.time[previous_matrix_index][next_matrix_index] || 0
+          shift_route_times_backward(previous_end + intermediate_duration - next_start - new_travel_time, index)
```
senhalil: But (like you said) if there is force-start or if there are tw.end and/or tw.begin it might be preferable to push stuff to the left or right or both... how can we decide? 

here we assume, either we should add check at the beginning of this function so that the "simplificaiton"/"cleaning" is only done under strict (simple) conditions that we can handle and then we can improve this function (and relax the check) along the road. How does that sound?

We need to decide which simple case to handle first and do the realise with that case next week. Do we know the clients use case? Lets talk this Monday morning..

senhalil (commented):
I am not done but cannot look at this before Monday afternoon so submitting it as is.

models/solution/route.rb
braktar: it should be previous_stop ;)

models/solution/route.rb
braktar: In the context of the customer, we have both force_start and a timewindow on a service which stands for a located rest

braktar:
I've removed all the filter and shift logic as it was trying to fix a "false problem" or at least a problem that could be addressed with a good definition of the problem

senhalil (commented):
Do we need any checks for assumptions of `shift_route_distance` and can we add a warning (or raise) for the assumptions of `shift_route_times` that we discussed last week?

senhalil (approved):

PR #391, opened 2022-04-05
Fix zip computed times & change condition (rests)

braktar:


senhalil (approved):

optimizer_wrapper.rb
```diff
@@ -591,7 +590,7 @@ def self.apply_zones(vrp)
   end
 
   def self.clique_cluster(vrp, cluster_threshold, force_cluster)
-    @zip_condition = cluster_threshold && vrp.matrices.any? && vrp.services.all?(&:activity) &&
+    @zip_condition = cluster_threshold && vrp.matrices.any? && vrp.services.all?(&:activity) && vrp.rests.empty? &&
```
senhalil, line 593: I just saw this condition. Why do we want to skip zip_cluster because of a vehicle with pause?

braktar: Zip cluster ignore the timewindows. The upper level route is build with compressed zip_clusters. So when the solution is then expended with unzip_cluster and the result of the TSPs the times are shifted. We cannot ensure that the rests stays feasible

senhalil: Cheers for the explanation

PR #392, opened 2022-04-08
Fix activities and relations

braktar:
- [x] needs https://github.com/Mapotempo/optimizer-ortools/pull/35
- [x] and Docker.yml and etc modification

### DIFF
--- a/models/concerns/validate_data.rb
+++ b/models/concerns/validate_data.rb
@@ -203,17 +203,25 @@ module ValidateData
         @hash[:services].find{ |service| service[:id] == linked_id }
       }
       previous_service = nil
+      previous_activities = []
       services.each{ |service|
-        if previous_service && forbidden_position_pairs.include?([previous_service[:activity][:position], service[:activity][:position]])
-          inconsistent_position_services << [previous_service[:id], service[:id]]
-        end
+        activities = service[:activity] ? [service[:activity]] : service[:activities]
+
+        previous_activities.each{ |previous_activity|
+          activities.each{ |activity|
+            next unless forbidden_position_pairs.include?([previous_activity[:position], activity[:position]])
+
+            inconsistent_position_services << [previous_service[:id], service[:id]]
+          }
+        }
         previous_service = service
+        previous_activities = activities
       }
     }
 
     return unless inconsistent_position_services.any?
 
-    raise OptimizerWrapper::DiscordantProblemError.new("Inconsistent positions in relations: #{inconsistent_position_services}")
+    raise OptimizerWrapper::DiscordantProblemError.new("Inconsistent positions in relations: #{inconsistent_position_services.uniq}")
   end
 
   def calculate_day_availabilities(vehicles, timewindow_arrays)
@@ -405,6 +413,7 @@ module ValidateData
     check_position_relation_specificities
     check_relation_consistent_ids
     check_sticky_relation_consistency
+    check_relation_compatibility_with_alternatives
 
     if periodic_heuristic
       incompatible_relation_types = @hash[:relations].collect{ |r| r[:type] }.uniq - %i[force_first never_first force_end same_vehicle]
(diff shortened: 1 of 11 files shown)


wrappers/wrapper.rb
```diff
@@ -671,7 +671,8 @@ def check_timewindow_inconsistency(vrp, unfeasible, service)
 
         max_earliest_arrival = 0
         relation.linked_services.map{ |service_in|
-          next unless service_in.activity.timewindows.any?
+          # TODO: Should consider alternatives
```
senhalil: can you add a log warning as well? So that we have a way to see how many times we ignore activities.

models/solution/parsers/stop_parser.rb
```diff
@@ -36,7 +37,7 @@ def self.parse(service, options)
         pickup_shipment_id: service.type == :pickup ? (service.original_id || service.id) : nil,
         delivery_shipment_id: service.type == :delivery ? (service.original_id || service.id) : nil,
         type: service.type,
-        alternative: options[:index],
+        alternative: alternative_index,
```
senhalil: Isn't it better to return `alternative: nil` which will be eliminated by solution parsing so that the user has a way to distinguish between services with alternative and without one?

models/concerns/validate_data.rb
```diff
@@ -484,6 +491,25 @@ def check_sticky_relation_consistency
...
+      services = @hash[:services].select{ |service| relation[:linked_service_ids]&.include?(service[:id]) }
+      not_handled_relations << relation if services.any?{ |service|
+        (service[:activities]&.size || 0) > 1
+      }
+    }
+
+    return unless not_handled_relations.any?
+
+    raise OptimizerWrapper::UnsupportedProblemError.new(
+      "The following relations contains alternatives but didn't handle them: ",
```
senhalil: The following relations are not compatible with alternative services or something like that?

models/relation.rb
```diff
@@ -20,6 +20,7 @@
 module Models
   class Relation < Base
     ALL_OR_NONE_RELATIONS = %i[shipment meetup].freeze
+    ALTERNATIVE_COMPATIBLE = %i[order same_route sequence shipment force_first never_first force_end]
```
senhalil: ALTERNATIVE_COMPATIBLE_RELATIONS  (all others have _RELATIONS :)

senhalil: Can you .freeze it?

senhalil: Also a test would be nice (we have one for LINKING_RELATIONS and FORCING_RELATIONS for example) so that if we modify it by mistake the test will alert us.

models/concerns/validate_data.rb
```diff
@@ -203,17 +203,23 @@ def check_position_relation_specificities
         @hash[:services].find{ |service| service[:id] == linked_id }
       }
       previous_service = nil
+      previous_activities = []
       services.each{ |service|
-        if previous_service && forbidden_position_pairs.include?([previous_service[:activity][:position], service[:activity][:position]])
-          inconsistent_position_services << [previous_service[:id], service[:id]]
-        end
+        activities = service[:activity] && [service[:activity]] || service[:activities]
```
senhalil: ternary logic is more clear imho, no?
```suggestion
        activities = service[:activity] ? [service[:activity]] : service[:activities]
```

models/concerns/validate_data.rb
```diff
@@ -203,17 +203,23 @@ def check_position_relation_specificities
...
+
+        previous_activities.each{ |previous_activity|
+          activities.each{ |activity|
+            inconsistent_position_services << [previous_service[:id], service[:id]] if forbidden_position_pairs.include?([previous_activity[:position], activity[:position]])
```
senhalil: line too long, rubocop doesn't detect because this cop is probably deactivated on this file :(

senhalil (approved):

PR #393, opened 2022-04-13
Fix infinite loop and import some comments

braktar:


lib/heuristics/periodic_heuristic.rb
```diff
@@ -703,29 +703,31 @@ def insertion_cost_with_tw(timewindow, route_data, service, position)
...
           }.reduce(&:+)
-        while route_exclusion_costs < route_cost_fixed || can_not_insert_more
+        while route_exclusion_costs < route_cost_fixed
           inserted_id, _unlocked_ids = try_to_add_new_point(vehicle_id, started_day)
           impacted_days |= adjust_candidate_routes(vehicle_id, started_day)
           if inserted_id
             route_exclusion_costs += @services_data[inserted_id][:raw].exclusion_cost.to_f
           else
-            can_not_insert_more = true
+            break
```
senhalil, line 722: Next time can you separate the actual fix and the other stuff like the syntax and comment? Here, I accept that it is quite manageable but if we don't have the discipline to separate commits, PRs get nasty quite quickly ;)

senhalil (approved):

PR #396, opened 2022-04-14
Call or-tools after periodic heuristic

braktar:


wrappers/ortools.rb
```diff
@@ -59,7 +59,6 @@ def solver_constraints
         :assert_solver_if_not_periodic,
         :assert_first_solution_strategy_is_possible,
         :assert_first_solution_strategy_is_valid,
-        :assert_clustering_compatible_with_periodic_heuristic,
```
senhalil: I think you excluded the max_split modification from this PR. Did you forget to remove this change?

optimizer_wrapper.rb
```diff
@@ -226,9 +226,14 @@ def self.solve(service_vrp, job = nil, block = nil)
...
         optim_solution = vrp.configuration.preprocessing.heuristic_result if vrp.periodic_heuristic?
+        if vrp.configuration.resolution.solver && vrp.configuration.preprocessing.first_solution_strategy == ['periodic']
+          vrp.configuration.preprocessing.first_solution_strategy = ['global_cheapest_arc']
+        end
       end
 
-      if vrp.configuration.resolution.solver && !vrp.periodic_heuristic?
+      if vrp.configuration.resolution.solver &&
+         (vrp.configuration.preprocessing.partitions.empty? && vrp.configuration.preprocessing&.max_split_size.nil? ||
+           vrp.services.size < (vrp.configuration.preprocessing&.max_split_size || 0))
```
senhalil: Why do we need these extra conditions? Max split and partition logics are handled before `solve()`. 

Is it because you want to limit the size of the general problem that is optimized with or-tools directly as we talked? 

If that is the case what I expected was a more literal condition like this 

```suggestion
      if vrp.configuration.resolution.solver && (
        !vrp.periodic_heuristic? || vrp.services.size < XXX
      )
```

This way we are sure that we do not affect the behaviour of other options. Plus if we the user increase max_split for any reason this would lead to ortools being launched with too many services or if they do not set max_split we might launch an optim with 10K services (which would definitely lead to an OOM).

Since with the above modification I suggested we are sure that this check will be active after periodic_heuristic is already launched and we certainly have an initial solution, I suggest a value like 500 for `XXX`.

senhalil (commented):
There are some errored tests and now that we launch solver after periodic_heuristic the tests might take longer (because we didn't check the tests whether they have an extremely long mini_duration).

senhalil (commented):
I think it is :rocket: if you can make the CI pass without touching the CI timelimits :grimacing:

test/real_cases_periodic_test.rb
```diff
@@ -222,6 +222,7 @@ def test_treatment_site
 
     def test_route_initialisation
       vrp = TestHelper.load_vrp(self)
+      vrp.configuration.resolution.solver = false
```
braktar: This kind of change did not satisfy me :/
I'm currently considering to add a new parameter to allow "vrp resolution" after scheduling

@senhalil What do you think ?

braktar:
As the vrp resolution is performed outside of the context of periodic heuristic, the solution content did not contain info like day_week & co which are necessary for the construction of the csv in the context of `test_csv_headers_compatible_with_import_according_to_language`

test/real_cases_periodic_test.rb
senhalil: Why don't you like this modification?

I mean I don't like it either but my problem with it is the fact that if you turn the solver off `vrp.configuration.resolution.solver = false` the solver is not going to be called even for the sub-problems themselves.

senhalil: I think if we are considering a new parameter, a parameter to limit the "size" of problems that we launch the solver would be better no? We already have a solver=false/true which we can base our decision on.

senhalil:
> the solution content did not contain info like day_week & co which are necessary for the construction of the csv in the context of 

If it is related to CSV how did this test worked before? Possible that I didn't quite get the problem but is it because periodic solution has some extra/special info in the solution and if we launch solver after periodic such info get erased?

braktar:
Yes ! That's it!

senhalil:
> Yes ! That's it!

In that case, I guess we have to make sure that the output is uniform. Do you think it can be done without much hassle?

models/solution/solution.rb
```diff
@@ -96,6 +96,14 @@ def +(other)
...
     end
 
+    def <(other)
+      self_metrics = [self.unassigned_stops.size, self.cost]
+
+      other_metrics = [other.unassigned_stops.size, other.cost]
+
+      self_metrics <=> other_metrics
+    end
+
```
senhalil: I am not sure if this is the best way to handle this because if there are "real" exclusion costs, this might cause a problem. Imagine more unassigned services but less total cost...

(Not related to this PR, I am not sure if we include the real exclusion in the total cost, do you know if we do?)

optimizer_wrapper.rb
```diff
@@ -221,14 +221,23 @@ def self.solve(service_vrp, job = nil, block = nil)
...
         periodic = Interpreters::PeriodicVisits.new(vrp)
         vrp = periodic.expand(vrp, job, &block)
         optim_solution = vrp.configuration.preprocessing.heuristic_result if vrp.periodic_heuristic?
+        if vrp.configuration.resolution.solver && vrp.configuration.preprocessing.first_solution_strategy == ['periodic']
+          vrp.configuration.resolution.duration -= optim_solution.elapsed
+          if vrp.configuration.resolution.minimum_duration && vrp.configuration.resolution.duration
+            vrp.configuration.resolution.minimum_duration *= 1 - optim_solution.elapsed / vrp.configuration.resolution.duration
+          end
+          vrp.configuration.preprocessing.first_solution_strategy = ['global_cheapest_arc']
+        end
```
senhalil: I am not sure if we need this time correction logic because the heuristic is quite fast (especially for small instances)  but also we always argue that the min/max_durations are for "solver time" but this elapsed time includes all the overheads plus it might lead to negative time at this current state
```ruby
vrp.configuration.resolution.duration -= optim_solution.elapsed
if vrp.configuration.resolution.minimum_duration && vrp.configuration.resolution.duration
     vrp.configuration.resolution.minimum_duration *= 1 - optim_solution.elapsed / vrp.configuration.resolution.duration
end
```

Lets not override user's first_solution_strategy choices:
```suggestion
        if vrp.configuration.resolution.solver && vrp.periodic_heuristic?
          first_solution_strategy = vrp.configuration.preprocessing.first_solution_strategy
          first_solution_strategy -= 'periodic'
          first_solution_strategy += 'global_cheapest_arc' if first_solution_strategy.empty?
        end
```

optimizer_wrapper.rb
```diff
@@ -280,6 +289,11 @@ def self.solve(service_vrp, job = nil, block = nil)
     end
 
     if optim_solution # Job might have been killed
+      if periodic_heuristic_flag
+        periodic_solution = vrp.configuration.preprocessing.heuristic_result
+        optim_solution = periodic_solution if optim_solution < periodic_solution
```
senhalil: Why are we overriding optim_solution if it is "better" (i.e., `<` less unplanned, less cost, etc.)?

wrappers/ortools.rb
```diff
@@ -378,6 +379,17 @@ def build_route_stop(vrp, vehicle, problem_services, problem_rests, activity)
...
+    def build_schedule_times(vrp, vehicle)
+      return {} unless vrp.schedule?
+
+      size_weeks = (vrp.configuration.schedule.range_indices[:end].to_f / 7).ceil.to_s.size
+      week = Helper.string_padding(vehicle.global_day_index / 7 + 1, size_weeks)
+      {
+        day_week_num: "#{vehicle.global_day_index % 7}_#{week}",
+        day_week: "#{OptimizerWrapper::WEEKDAYS[vehicle.global_day_index % 7]}_#{week}"
+      }
+    end
```
senhalil: Is this the "custom" logic done by periodic_heuristic? If so, can you export this logic to a function on the periodic heuristics side and call it here so that there is one function. Maybe you can add a TODO here saying that this should be handled by the `Solution` class so that it is automatic for all solvers.

braktar: This method is now part of Solution::Stop::Info class and is called by the Solution::Route initializer or once a stop is inserted into a route

models/solution/solution.rb
braktar: exclusion cost is currently part of cost if it is defined for the unassigned nodes. So it is indeed not relevant to only rely on the number of unassigned stops in these cases. Moreover it would depend on priority as well. 

I will provide another formula to compare two solutions :+1:

models/solution/parsers/route_parser.rb
```diff
@@ -33,6 +33,7 @@ def self.parse(route, vrp, matrix, options = {})
                     @route.stops.any?(&:service_id)
 
       route_data ||= route_info(vrp)
+      route.stops.each{ |stop| stop.info.set_schedule(vrp, route.vehicle) }
```
senhalil: If you have this general function now, can't we remove the logic from the periodic heuristic as well? I do not see such removal in this PR, did I miss it?

optimizer_wrapper.rb
```diff
@@ -221,14 +221,27 @@ def self.solve(service_vrp, job = nil, block = nil)
...
+          vrp.configuration.resolution.duration -= optim_solution.elapsed
+          if vrp.configuration.resolution.minimum_duration && vrp.configuration.resolution.duration
+            vrp.configuration.resolution.minimum_duration *= 1 - optim_solution.elapsed / vrp.configuration.resolution.duration
+          end
+
+          if vrp.configuration.resolution.solver && vrp.periodic_heuristic?
+            vrp.configuration.preprocessing.first_solution_strategy  -= ['periodic']
+            vrp.configuration.preprocessing.first_solution_strategy  += ['global_cheapest_arc'] if vrp.configuration.preprocessing.first_solution_strategy.empty?
+          end
+        end
```
senhalil: I don't think we should modify the time. The time passed in periodic has nothing to do with solver time -- because it is not controlled by minimum_and maximum_duration and as we do not modify the minimum/maximum duration due to other preprocessing as Split/Dicho time passed inside vroom/clustering, we should leave this time be.

The intermediary optimization time is not controlled by the user and can be excluded from the calculations. Otherwise, the user has no control over how much time will be spent inside the last optimisation. 

It looks like my previous suggestion to have an intermediary variable failed due to the assignment operator not modifying the original array :face_exhaling:  sorry about that. What about the one down below, it is a lot more compact and doesn't generate a future Rubocop violation :heart: 
```suggestion
        if vrp.periodic_heuristic?
          optim_solution = vrp.configuration.preprocessing.heuristic_result

          if vrp.configuration.resolution.solver
            first_solution_strategy = vrp.configuration.preprocessing.first_solution_strategy
            first_solution_strategy.delete('periodic')
            first_solution_strategy << 'global_cheapest_arc' if first_solution_strategy.empty?
          end
        end
```

optimizer_wrapper.rb
senhalil: negative time happened already :) 
Screenshot from 2022-04-21 11-59-46

optimizer_wrapper.rb
braktar: Oh, my bad ! I was pretty sure I had removed the time reduction

senhalil (approved):

PR #407, opened 2022-05-03
Forbid zip cluster with timewindows And used capacities 

braktar:


optimizer_wrapper.rb
```diff
@@ -612,17 +612,22 @@ def self.apply_zones(vrp)
     }
   end
 
-  def self.clique_cluster(vrp, cluster_threshold, force_cluster)
+  def self.set_zip_condition(vrp, cluster_threshold)
     @zip_condition = cluster_threshold && vrp.matrices.any? && vrp.services.all?(&:activity) && vrp.rests.empty? &&
```
senhalil: While you are at it, can you remove the `@zip_condition` variable, please? It is used only locally in the `clique_cluster` function. 

Then we can name this function to `clique_cluster_candidate?` (we have this convention for dicho and split_solve) and its result can be assigned to a local variable `apply_clique_cluster` which can replace all `@zip_condition` inside `clique_cluster`.

api/v01/entities/vrp_input.rb
```diff
@@ -157,7 +157,6 @@ module VrpConfiguration
     optional(:partition_metric, type: Symbol, documentation: { hidden: true }, desc: '[ DEPRECATED : use partitions structure instead ]')
     optional(:kmeans_centroids, type: Array[Integer], documentation: { hidden: true }, desc: '[ DEPRECATED : use partitions structure instead ]')
     optional(:cluster_threshold, type: Float, desc: 'Regroup close points which constitute a cluster into a single geo-located point')
-    optional(:force_cluster, type: Boolean, desc: 'Force to cluster visits even if containing timewindows and quantities')
```
senhalil, line 160: I think this is a very good idea because i) we don't really support it ii) it can lead to infeasible solutions and iii) no one uses it... However, we still need to do it the right way :/

Can you mark it "Deprecated" and "Hidden: true" here and then in validate_data.rb add a condition  to the `check_configuration` function that raises and `UnsupportedProblemError` if force_cluster is true. Otherwise, we silently ignore a deprecated parameter :see_no_evil: :hear_no_evil: :speak_no_evil:

CHANGELOG.md
```diff
@@ -20,6 +20,7 @@
 - Mapotempo github wiki is now directly part of the project [#351](https://github.com/Mapotempo/optimizer-api/pull/351)
 - Empty/Fill behavior changed so that the loads are naturally managed within optimizer-ortools [#370](https://github.com/Mapotempo/optimizer-api/pull/370)
 - Allow multiple rests with the same day index in periodic heuristic algorithm (`first_solution_strategy='periodic'`) [#273](https://github.com/Mapotempo/optimizer-api/pull/273)
+- Conditions to benefit of `cluster_threshold` are more stricts. This avoids shifts out of feasible timewindows and capacity excesses. [#407](https://github.com/Mapotempo/optimizer-api/pull/407)
 
 ### Removed
```
senhalil:
```suggestion
- Conditions to benefit from `cluster_threshold` parameter are more strict -- to avoid time-window and capacity constraint violations [#407](https://github.com/Mapotempo/optimizer-api/pull/407)

### Removed

- `force_cluster` option is removed -- `cluster_threshold` is applied only on supported cases [#407](https://github.com/Mapotempo/optimizer-api/pull/407)
```

senhalil (commented):
minor corrections but otherwiese :rocket:

api/v01/entities/vrp_input.rb
braktar: As `force_cluster` has no more behavior in our problem, i don't really agree that it should be kept in the input model even hidden. This was the case as well for the `type` field of services. Some other fields were marked Removed in the changelog, but the behavior was kept through another parameter (ex `solver_parameter`). If the behavior is kept in a way or in an other, from my point of view it justifies to keep it. Otherwise, it can be dumped. Moreover, that is a field which has only be used by me in 2016. No one used it since then.

Tell me if that's ok for you

senhalil: I mean, I get you but it was in our documentation since 2016 and since "we" are changing the behaviour, we need to advertise the change properly. 

If a user (this could even be you) has an old JSON file with `force_cluster: true` we need to be able to refuse that file since force clustering was once supported but it is not supported anymore. Otherwise, we have no way to check since declared_params remove unknown fields -- and for obvious reasons, we cannot refuse all undeclared params -- and the user has no way to understand the clustering is not done.

In fact, we need to raise not only if force_cluster is true but even if it is false -- so if the config hash has force_cluster key. Does it make sense for you?

braktar: Raising an error for field which have no more purpose seems a bit dangerous for me. It it was a parameter used by a user in production, we would not block the solve. What if i only generate a warning sentry on our side ?

senhalil: I mean yeah, warning sentry would work too :+1: 

To be honest, I think completely ignoring an option is far more dangerous than blocking the user -- it is us changing the behaviour, we need to properly communicate it -- so I don't understand exactly why you are okay with the first but not with the latter. 

Anyways, in the sentry warning, can you mention that this parameter is deleted and we need to explain to the user that it is ignored and they should stop sending it.

senhalil (approved):

PR #409, opened 2022-05-11
Fix self selection with activities

braktar:


lib/interpreters/compute_several_solutions.rb
```diff
@@ -259,7 +259,10 @@ def self.select_best_heuristic(vrp)
               }
         end
       }
-      size_mtws = services.count{ |service| service.activity.timewindows.size > 1 }
+      size_mtws = services.count{ |service|
+        (service.activity&.timewindows&.size || 0) > 1 ||
+         service.activities.flat_map{ |act| act.timewindows.size }.size > 1
```
senhalil: Is this really what we want -- i.e., counting a service with multiple activities each with a single time window as a service with multiple time windows?

If I am not mistaken, if there is only one activity with multiple time windows inside `activities` then this count fails, right?

senhalil (changes requested):

lib/interpreters/compute_several_solutions.rb
braktar: Multiple activities with one timewindow each is similar to one activity with multiple timewindows (specially in the case of lateness authorized).

I've changed a bit the condition, as your remarks make me think, that no timewindow is similar to an infinite timewindow. then we still need to count it correctly.

lib/interpreters/compute_several_solutions.rb
```diff
@@ -259,7 +259,10 @@ def self.select_best_heuristic(vrp)
               }
         end
       }
-      size_mtws = services.count{ |service| service.activity.timewindows.size > 1 }
+      size_mtws = services.count{ |service|
+        (service.activity&.timewindows&.size || 1) > 1 || # No timewindow means an implicit infinite timewindow
+          service.activities.flat_map{ |act| [act.timewindows.size, 1].max }.sum > 1
```
senhalil: We can simplify as follows I think. Does it look okay to you? 
```suggestion
        service.activity&.timewindows&.size.to_i > 1 || 
          # No timewindow means an implicit infinite timewindow
          service.activities.sum{ |act| [act.timewindows.size, 1].max } > 1
```

senhalil (approved):

PR #410, opened 2022-05-12
Doc update

braktar:
First batch of doc edits

### DIFF
--- a/docs/Activity.md
+++ b/docs/Activity.md
@@ -1,3 +1,5 @@
+# Activity
+
 Describe where an activity take place, when it can be performed and how long it last.
 ```json
   "activity": {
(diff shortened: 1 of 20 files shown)


senhalil (approved):
Can't we generate (at least a good chunk of) this wiki/doc from the grape-swagger documentation? It feels like we have a lot of duplicated information and as was the case before this PR, it will be fast outdated.

senhalil:
@braktar for example this [repo](https://github.com/peter-evans/swagger-github-pages) does something like that -- it puts swagger documentation on github -- or [this repo](https://github.com/drg-adaptive/swagger-wiki) might be a better fit -- it generates wiki from swagger? What do you think?

PR #415, opened 2022-05-23
Fix avancement for max split

braktar:
Simple fix before entire rework

### DIFF
--- a/lib/interpreters/split_clustering.rb
+++ b/lib/interpreters/split_clustering.rb
@@ -231,8 +231,12 @@ module Interpreters
           ss_data[:cannot_split_further] = true
         else
           service_vrp[:split_level] = split_level + 1
-          service_vrp[:split_denominators] << 2**service_vrp[:split_level]
-          service_vrp[:split_sides] << 0
+          if !service_vrp[:split_denominators].include?(2**service_vrp[:split_level])
+            service_vrp[:split_denominators] << 2**service_vrp[:split_level]
+            service_vrp[:split_sides] << 0
+          else
+            service_vrp[:split_sides][split_level + 1] = 0
+          end
         end
 
         # TODO: In some case one side might be significantly smaller than the other side.
@@ -271,7 +275,7 @@ module Interpreters
         }
         ss_data[:cannot_split_further] = false
         log "<-- split_solve_core level: #{split_level}"
-
+        service_vrp[:split_level] = split_level
         solutions.reduce(&:+)
       else # Stopping condition -- the problem is small enough
         # Finally generate the sub-vrp without hard-copy and solve it


senhalil (changes requested):
This is not the proper fix -- `service_vrp[:split_level] = split_level` assignment is at an incorrect place. Because of this, the right branch will use the split_level of the last children of the left branch (when we are backtracking) -- which leads to higher split_levels on the right-hand side branch. service_vrp[:split_level] is modified during recursion. 

I already have the fix -- I remember telling you that I am on it already and you shouldn't bother.

lib/interpreters/split_clustering.rb
```diff
@@ -271,7 +275,7 @@ def self.split_solve_core(service_vrp, job = nil, &block)
         }
         ss_data[:cannot_split_further] = false
         log "<-- split_solve_core level: #{split_level}"
-
+        service_vrp[:split_level] = split_level
```
braktar, line 278: @senhalil This line allows to fix the constant increasing of the split_level by reverting the split_level to its value before the dive to a deeper level.

lib/interpreters/split_clustering.rb
```diff
@@ -231,8 +231,12 @@ def self.split_solve_core(service_vrp, job = nil, &block)
           ss_data[:cannot_split_further] = true
         else
           service_vrp[:split_level] = split_level + 1
-          service_vrp[:split_denominators] << 2**service_vrp[:split_level]
-          service_vrp[:split_sides] << 0
+          if !service_vrp[:split_denominators].include?(2**service_vrp[:split_level])
```
braktar, line 234: This condition allows to initialize a new level never explored until now.

lib/interpreters/split_clustering.rb
```diff
@@ -231,8 +231,12 @@ def self.split_solve_core(service_vrp, job = nil, &block)
...
+            service_vrp[:split_denominators] << 2**service_vrp[:split_level]
+            service_vrp[:split_sides] << 0
+          else
+            service_vrp[:split_sides][split_level + 1] = 0
```
braktar, line 238: When arriving to a level already explored by another branch we reset the counter of this level. But it should be applied to every subsequent levels too

lib/interpreters/split_clustering.rb
braktar:
```suggestion
          if service_vrp[:split_denominators].size < service_vrp[:split_level]
```

lib/interpreters/split_clustering.rb
braktar:
```suggestion
            (split_level + 1..-1).map{ |level| service_vrp[:split_sides][level] = 0 }
```

lib/interpreters/split_clustering.rb
senhalil: it is not at the correct place

PR #416, opened 2022-05-24
Dicho uses splitclustering split to respect relations

senhalil:
Respects relations and initial solutions
Splitclustering.split is several magnitudes faster since it works on representative vrp
Clustering quality increase
Final solution quality improvement -- from ~50% unplanned services -> to ~10% unplanned services


lib/interpreters/split_clustering.rb
```diff
@@ -246,6 +247,9 @@ def self.split_solve_core(service_vrp, job = nil, &block)
 
         # RECURSIVELY CALL split_solve_core AND MERGE SOLUTIONS
         solutions = sides.collect.with_index{ |side, index|
+          # This is a recursive function and service_vrp[:split_level] needs to be corrected because
+          # it is incremented within the children of the other branch ("left-side") of this level
+          service_vrp[:split_level] = split_level + 1 # This needs to be here!
```
braktar, line 251: In a test case, by the end of the resolution, the split_denominators array : `[1, 2, 4, 8, 16, 4, 8, 16, 8]` but in the logic I had in mind it should be `[1, 2, 4, 8, 16]`. Which was the purpose of this PR: https://github.com/Mapotempo/optimizer-api/pull/415

Regardless of the position of the `split_level` increment

lib/heuristics/dichotomous_approach.rb
```diff
@@ -111,30 +114,30 @@ def self.dichotomous_heuristic(service_vrp, job = nil, &block)
...
-            transfer_unused_vehicles(service_vrp, solution, sub_service_vrps) if index.zero? && solution
+            transfer_unused_vehicles(solution, sub_service_vrps) if index.zero? && solution
 
             solution
           }
           solution = solutions.reduce(&:+)
           log "dicho - level(#{service_vrp[:dicho_level]}) before remove_bad_skills unassigned rate " \
               "#{solution.unassigned_stops.size}/#{service_vrp[:vrp].services.size}: " \
-              "#{(solution.unassigned_stops.size.to_f / service_vrp[:vrp].services.size * 100).round(1)}%", level: :debug
+              "#{(solution.unassigned_stops.size.to_f / service_vrp[:vrp].services.size * 100).round(1)}%"
```
braktar, line 115: Is it necessary to move these logs from :debug to :info ? The logs defined as :info are already barely readable due to their high quantity

lib/interpreters/split_clustering.rb
senhalil: We already chat about this on dicord and decided to postpone the dev -- since the progress bar increaments in this PR are really "beautiful", we decided to leave it for the main advancement ticket (which will probably be based on something else then number of subproblems).

lib/heuristics/dichotomous_approach.rb
senhalil: I asked the same question :thinking: I ended up activating them because
- without the logs, it is impossible to see what is wrong (and dicho split is random so not easy to reproduce)
- most of the logs I activated are either once or twice per job or per sub-problem -- so we won't have 100s/1000s of extra lines of log because of them.

braktar (approved):

PR #418, opened 2022-06-01
Prepare release

braktar:


CHANGELOG.md
```diff
@@ -1,6 +1,6 @@
 # Changelog
 
-## [v1.9.0-dev] - Unreleased
```
senhalil, line 3: Can you add `## [v1.10.0-dev] - Unreleased`

api/api_v01.rb
```diff
@@ -65,7 +65,7 @@ class ApiV01 < Grape::API
 
 Its purpose is to provide a complete chain for the resolution. From a provided VRP, it requires a distance matrix, solve the problem and prepare a self sufficient result.
 
-Please check the Github Wiki for more details: [docs](docs/Home.md)
+Please check the projet documentation for more details: [docs/Home](https://github.com/Mapotempo/optimizer-api/blob/dev/docs/Home.md)
```
senhalil: Do we want `blob/dev` or `master`? At the moment master doesn't exist but it will be after the merge :)

CHANGELOG.md
braktar: I'll create a separate commit/PR for this once the version is released ;)

senhalil:
I just added the "dicho" item to the changelog

api/api_v01.rb
braktar: master will be a fixed version of the doc. But I was thinking that the one which read the doc expects an up to date documentation. What to do you think ?

senhalil: I was thinking if the doc moves on the dev branch and master points to dev then there will be a mismatch :/ 

I suspect that we do not really have to give the full link but a relative one (so that dev branch will show dev/doc and master will show the master/doc). 

Can you check if GitHub docs mention such full/relative path/link or if any other big repo that we use does something similar? If not possible we can go either way but tbh I prefer the master since that is the public/stable version.

braktar: Ruby have multiple way to link the doc :

- Static link to master : https://github.com/ruby/ruby/blob/master/doc/maintainers.rdoc#label-Platform+Maintainers
- Link to an external generated documentation

Rails use relative links and links to an external generated documentation

In our use case, the generated doc is generated manually, and rarely up to date, but this is the main entrance for software editors. We need to keep the link to the detailed doc correct in this context.

It's ok for me to point to master, but this implies to update master at each release and probably even fix versions ;)


N.B: Another subject, not related, the [swagger doc](http://swagger.mapotempo.com/) use an old version of swagger-ui which generates incorrectly our documentation.

senhalil: Sh*t, I just realised that this is the documentation which will not be on github.com. Is that it?

In that case, of course we need the full URL :/

If they are using the stable version they need the master, in any case, they can click on the branch button to switch to dev branch if they want.

swagger-ui -> unfortunately we are blocked by mt-web :( If we still don't have a ticket, can you create one in the next milestone or in the one after that  and tag @Lalensia so that he can plan the switch on both side. Effectively our docs (and mt-web-api docs) are not sexy nor correct in our case.

senhalil (approved):

api/api_v01.rb
senhalil: @braktar did you check if we have a ticket? You now the issue better than me, can you create a ticket if we don't have one please?

braktar: There is no ticket on Optim-API side as there is nothing to do to migrate swagger-ui from our perspective. On MWeb side, here is the issue : https://gitlab.com/woopit/mapotempo/mapotempo-web/-/issues/830

PR #421, opened 2022-06-03
Qual: fix all rubocop issues (except metrics)

senhalil:


lib/heuristics/concerns/periodic_end_phase.rb
```diff
@@ -423,12 +458,16 @@ def reaffect_prohibiting_partial_assignment
       break if adapted_still_removed.empty? || (adapted_still_removed - banned).empty?
 
       while most_prio_and_frequent.empty?
-        most_prioritary = (adapted_still_removed - banned).group_by{ |removed| @services_data[removed.first][:raw].priority }.min_by{ |priority, _set| priority }[1]
-        most_prio_and_frequent = most_prioritary.group_by{ |removed| removed[1] }.max_by{ |visits_number, _set| visits_number }[1]
+        most_prioritary = (adapted_still_removed - banned).group_by{ |removed|
```
braktar, line 461: Shouldn't `Layout/MultilineAssignmentLayout` apply here ?

lib/heuristics/dichotomous_approach.rb
```diff
@@ -283,7 +294,7 @@ def self.insert_unassigned_by_skills(service_vrp, unassigned_services, unassigne
       # However, it is not very easy to find a generic and effective way.
 
       sub_solutions = []
-      vehicle_count = (skills.empty? && !vrp.routes.empty?) ? [vrp.routes.size, 6].min : 3
+      vehicle_count = skills.empty? && !vrp.routes.empty? ? [vrp.routes.size, 6].min : 3
```
braktar: It has no impact on the evaluation result, but I would prefer to keep the parenthesis around the ternary operator condition

test/lib/heuristics/periodic_functions_test.rb
```diff
@@ -179,7 +182,8 @@ def test_clean_stops_small_lapses
           10 => {
             stops: [{ id: 'service_1', point_id: 'point_0' }], vehicle: vrp.vehicles.first.id, capacity: []
           }
-      })
+      }
+      )
```
braktar, line 186: Seems odd to me that consecutive `}` and `)` are on different line at the same column

util/job_manager.rb
```diff
@@ -115,7 +134,7 @@ def perform
       begin
         Result.set(self.uuid, p)
       rescue Redis::BaseError => e
-        log "Could not set the last result due to the following error (will try again after sleep): #{e}", level: :warning
+        log "Couldn't set the last result due to the following error (tring again in 2 minutes): #{e}", level: :warning
```
braktar:
```suggestion
        log "Couldn't set the last result due to the following error (trying again in 2 minutes): #{e}", level: :warning
```

lib/heuristics/dichotomous_approach.rb
senhalil: It is the rubocop-mapotempo :/

senhalil: Otherwise, I agree that complex ternary with parenthesis `()`  is more legible.

lib/heuristics/concerns/periodic_end_phase.rb
senhalil: If I understand it correctly, it doesn't apply the rule when the block is chained. I decide to leave it to the automation, though since it would be loosing battle to fix them manually :D

test/lib/heuristics/periodic_functions_test.rb
senhalil: Agreed, it should have indented the `}` :/ I'll fix it and if it doesn't take it back when re-launch, submit it with the fix

lib/heuristics/concerns/periodic_end_phase.rb
senhalil: If I understand it correctly, it doesn't apply the rule when the block is chained. I decide to leave it to the automation, though since it would be loosing battle to fix them manually :D

test/lib/heuristics/periodic_functions_test.rb
senhalil: fixed it manually and rubocop didn't take it back :shrug: don't know why it doesn't fix it in the first place (rubocop bug?)

lib/heuristics/dichotomous_approach.rb
senhalil: What do you think, shall we change this cop -- it is already enabled but it is on `EnforcedStyle: require_no_parentheses (default)` and I am thinking `EnforcedStyle: require_parentheses_when_complex`

braktar: Seems good to me !

braktar (approved):

PR #426, opened 2022-06-14
Consider alternative in points collection in split_clustering

braktar:


lib/interpreters/split_clustering.rb
```diff
@@ -331,7 +331,7 @@ def self.create_sub_vrp(split_solve_data)
       sub_vrp.services.concat ss_data[:transferred_empties_or_fills]
 
       # only necessary points -- because compute_matrix doesn't check the difference
-      sub_vrp.points = sub_vrp.services.map{ |s| s.activity.point } |
+      sub_vrp.points = sub_vrp.services.flat_map{ |s| s.activity ? s.activity.point : s.activities.map(&:point) } |
```
senhalil:
```suggestion
      sub_vrp.points = sub_vrp.services.flat_map{ |s| s.activity&.point || s.activities.map(&:point) } |
```

senhalil (commented):
I wish we could have got rid of this "activity/activities" duality before but 🤷

lib/interpreters/split_clustering.rb
```diff
@@ -294,13 +294,15 @@ def self.split_solve_core(service_vrp, job = nil, &block)
         ss_data[:cannot_split_further] = false
         log "<-- split_solve_core level: #{split_level}"
 
-        solutions.reduce(&:+)
+        solution = solutions.reduce(&:+)
+        Cleanse.cleanse_duplicate_empties_fills(service_vrp[:vrp], solution)
```
senhalil: Is there a specific reason for calling this here instead of adding it to the `self.cleanse` function so that it is called systematically?

braktar: the cleanse method is called before the merge of the solutions :/

senhalil: I don't understand, why this forces us to call it here. Can't we keep the duplicate empty/fill till the end of solve (where we call the cleanse method -- if I am not mistaken). If we cannot then, maybe the location of the cleanse method call is not correct either..

braktar: Cleanse was in the `solve` method, this implies that it was only applied right after the call of the solver.

I've moved the cleanse method to `define_process`. This way, every solution whatever the resolution process (split/dicho/solver) is correct.

senhalil (approved):

senhalil:
@braktar can you resolve the sentry and share the link of this PR there please?

PR #428, opened 2022-06-16
Expose the effective setup_time

braktar:
@senhalil Any suggestion for the naming ? i am not really fan of `arrival_time`

### DIFF
--- a/api/v01/entities/vrp_result.rb
+++ b/api/v01/entities/vrp_result.rb
@@ -62,6 +62,7 @@ module Api
       expose :travel_time, documentation: { type: Integer, desc: 'Travel time from previous point (in s)' }
       expose :travel_value, documentation: { type: Integer, desc: 'Travel value from previous point' }
       expose :waiting_time, documentation: { type: Integer, desc: 'Idle time (in s)' }
+      expose :setup_time, documentation: { type: Integer, desc: 'Effective setup duration (in s)'}
       expose :begin_time, documentation: { type: Integer, desc: 'Time visit starts' }
       expose :end_time, documentation: { type: Integer, desc: 'Time visit ends' }
       expose :departure_time, documentation: { type: Integer, desc: '' }
--- a/wrappers/wrapper.rb
+++ b/wrappers/wrapper.rb
@@ -1578,6 +1578,8 @@ module Wrappers
         overall_total_travel_time_correction = 0
         solution.routes.each{ |route|
           vehicle = vehicles_grouped_by_vehicle_id[route[:vehicle_id]].first
+          vehicle.coef_setup = vehicle[:simplified_coef_setup]
+          vehicle.additional_setup = vehicle[:simplified_additional_setup]
           total_travel_time_correction = 0
           previous_point_id = nil
           route.stops.each{ |stop|
@@ -1602,6 +1604,10 @@ module Wrappers
 
           overall_total_travel_time_correction += total_travel_time_correction
           route.info.total_travel_time -= total_travel_time_correction.round
+
+          # Solution patcher must let the problem untouched
+          vehicle.coef_setup = 1
+          vehicle.additional_setup = 0
         }
         solution.info.total_travel_time -= overall_total_travel_time_correction.round
 
--- a/models/solution/route.rb
+++ b/models/solution/route.rb
@@ -20,6 +20,7 @@ require './models/base'
 module Models
   class Solution < Base
     class Route < Base
+      MISSION_TYPES = [:service, :pickup, :delivery].freeze
       field :geometry
 
       has_many :stops, class_name: 'Models::Solution::Stop'
@@ -38,6 +39,7 @@ module Models
         options[:vehicle] = vehicle
         hash = super(options)
         hash['activities'] = hash.delete('stops')
+        compute_setup_times(hash)
         hash['cost_details'] = hash.delete('cost_info')
         hash['detail'] = hash.delete('info')
         hash.merge!(info.vrp_result(options))
@@ -51,6 +53,7 @@ module Models
           stop = Models::Solution::Stop.new(stop) if stop.is_a? Hash
           stop.info.router_mode ||= vehicle&.router_mode
           stop.info.speed_multiplier ||= vehicle&.speed_multiplier
+          stop.compute_info_end_time(vehicle: vehicle)
           stop
         }
       end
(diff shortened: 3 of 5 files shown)


senhalil:
I thought we decided to drop this PR/dev -- since arrival time can mean different things with respect to waiting time/setup duration/etc and the user can easily calculate whichever the value they prefer using begin time, departure time, travel time and setup duration that exist in the solutions we return

braktar:
I just though about it, but the actual setup_duration return is not ponderated with the route vehicle properties :/

senhalil:
Hmmm yeah that complicates things, I would like to say let's do that -- i.e., return the actual/effectif setup duration instead of a nominative one -- but I am not sure of it. Do we return a nominative service duration or the effectif one?

braktar:
We currently return the nominative duration, in order to have the same behavior between assigned and unassigned activities

senhalil:
> We currently return the nominative duration, in order to have the same behavior between assigned and unassigned activities

If this is the only reason, we don't have to IMO because assigned services become "stops" they are different objects (they have timing and other related info) but unassigned services stay just "services". 

I find it more informative to show the actual setup duration spent for the stop for the assigned services otherwise that information is useless -- i.e., we show X minutes setup duration for a stop but it is not the first service of this location so the setup duration is not applied but to figure this out the user has to compare point_id's which is very clumsy and because of this calculating arrival_time is very complicated on the user side. 

That is the fact that we didn't do the right choice for setup-duration information now forces us to add another parameter. My suggestion is to fix the existing one instead of keeping it broken and adding more parameters on top of it

senhalil:
@braktar can you tell me your plan about this dev/PR? Did you talk to the client, are they okay with doing the calc on their side, if/once we correct the setup_duration (or add effective setup duration)?

braktar:
What came out after the discussion with the client  

1. Before v1.9.0, in some cases we were returning the field "arrival_time", they had the intention to use it. Some other client may have done the same.  
2. The detail fields are expected to be the same than in input (in the future, those fields should disappear).  
3. Compute the proper timing is not an easy thing with only the input data. We should return, as you say, someway or the effective setup_duration. Or the associated timing beside the other timings.

With these elements, introduce back a pre-existing field seems to me the best thing to do in this case. What do you think ?

Moreover it should be part of the CSV output

senhalil:
In fact, we removed it with a discussion that "arrival_time" doesn't have a predefined meaning (it can mean before break/wait/setup or after them or any combination) and that we shouldn't force the user or choose for them...

I do not think there are other users because we haven't got any complaints (yet but since it has been sometime now, I think we are good except for this client :crossed_fingers:) 

Hold on, I think I wasn't clear before but my suggestion is not putting the effective setup duration in the `:details` (I agree that details will go since it is duplicated data) but giving it as a part of the solution like we do for travel_time/waiting_time/etc between each activity (since effective setup duration is not duplicated data, it is part of the solution and dependent on vehicle and order of the activities). So under `VrpResultSolutionRouteActivities` we would have `setup_duration` and this value like travel_time/waiting_time/etc would be 0 if it is 0 in reality, or a value that is scaled with the vehicle coefficient and shifted with additional setup_duration if needed (like the way travel_time is scaled with speed multiplier in the background and the user doesn't fetch the matrix and multiply it with speed_mutiplier, the user will not do an operation to see the actual/applied/effective setup_duration).

senhalil:
Closed it by mistake while sending the message :face_exhaling:

senhalil:
What do you think after [my last message](https://github.com/Mapotempo/optimizer-api/pull/428#issuecomment-1161788151)?

I forgot to mention the CSV, (so that the files stay uploadable to mt-web) we can continue to put the "nominal" setup duration on its own column). But as we saw, there are other users who are expecting (or at least interpreting) effective setup duration and cannot understand why there is soo much difference. Unfortuanlity we can have the effective setup duration in the CSV only as an extra "informative" column (which becomes invalid if the user changes the orders of the activities).

api/v01/entities/vrp_result.rb
```diff
@@ -62,6 +62,8 @@ class VrpResultSolutionRouteActivities < Grape::Entity
       expose :travel_time, documentation: { type: Integer, desc: 'Travel time from previous point (in s)' }
       expose :travel_value, documentation: { type: Integer, desc: 'Travel value from previous point' }
       expose :waiting_time, documentation: { type: Integer, desc: 'Idle time (in s)' }
+      expose :setup_time, documentation: { type: Integer, desc: 'Effective setup duration (in s)'}
+      expose :arrival_time, documentation: { type: Integer, desc: 'Time visit setup duration starts'}
```
senhalil: Is this `:arrival_time` forgotten?

senhalil:
@braktar can you rename the commit as well so that git-log/blame/etc wouldn't be incorrect

wrappers/wrapper.rb
```diff
@@ -1578,6 +1578,8 @@ def simplify_service_setup_duration_and_vehicle_setup_modifiers(vrp, solution =
         overall_total_travel_time_correction = 0
         solution.routes.each{ |route|
           vehicle = vehicles_grouped_by_vehicle_id[route[:vehicle_id]].first
+          vehicle.coef_setup = vehicle[:simplified_coef_setup]
+          vehicle.additional_setup = vehicle[:simplified_additional_setup]
```
senhalil, line 1582: This is done by `:rewind` logic, do we need it?

wrappers/wrapper.rb
```diff
@@ -1602,6 +1604,10 @@ def simplify_service_setup_duration_and_vehicle_setup_modifiers(vrp, solution =
 
           overall_total_travel_time_correction += total_travel_time_correction
           route.info.total_travel_time -= total_travel_time_correction.round
+
+          # Solution patcher must let the problem untouched
+          vehicle.coef_setup = nil
+          vehicle.additional_setup = nil
```
senhalil: This operation is incorrect I think.

The `:simplify` operation sets these values to 1 and 0
```ruby
vehicle.coef_setup = 1
vehicle.additional_setup = 0
```
and `:rewind` takes these modifications back. Why do we need to set them to `nil` here?

wrappers/wrapper.rb
braktar: The `:rewind` is applied once the problem is over. We need to have correct intermediate solutions too

senhalil: I was mainly confused because of the `nil`s but I see that those are gone now.

senhalil (approved):

PR #452, opened 2022-07-22
Remove sticky_vehicles from periodic_heuristic

Pierre-Graber:


lib/interpreters/compute_several_solutions.rb
```diff
@@ -15,7 +15,6 @@
 # along with Mapotempo. If not, see:
 # <http://www.gnu.org/licenses/agpl.html>
 #
-
```
braktar: Let the line

braktar (approved):

PR #469, opened 2022-12-29
add new parameter to use a new Localsearch algorithm to find an initial solution

Pierre-Graber:


api/v01/entities/vrp_input.rb
```diff
@@ -152,6 +152,7 @@ module VrpConfiguration
   end
 
   params :vrp_request_preprocessing do
+    optional(:unconstrainted_initialization, type: Boolean, desc: 'Use experimental algorithm to find an initial solution')
```
alain-andre, line 155: Why is it called `unconstrainted_initialization` when the PR calls about an initial solution to find ? 
Shouldn't it be called `initialise_solution` ?

config/environments/development.rb
```diff
@@ -19,6 +19,7 @@
 require './wrappers/demo'
 require './wrappers/vroom'
 require './wrappers/ortools'
+require './wrappers/localsearch'
```
alain-andre: Shouldn't all modifications in `development.rb` be done in `production.rb` too?

wrappers/localsearch.rb
```diff
@@ -0,0 +1,28 @@
...
+#
+# You should have received a copy of the GNU Affero General Public License
+# along with Mapotempo. If not, see:
+# <http://www.gnu.org/licenses/agpl.html>
+#
+require './wrappers/wrapper'
+require './wrappers/ortools_vrp_pb'
+require './wrappers/ortools_result_pb'
+module Wrappers
+  class Localsearch < Ortools
```
alain-andre: I'm wondering about the name `Localsearch`. 

This class is called `Localsearch`, is used if `unconstrainted_initialization` is true and uses a `initial_solution`. 

Is there not a misunderstanding in the names?

api/v01/entities/vrp_input.rb
guillaumemaran: There are many initial solution calculation methods. This one is just a different one, which is unconstrainted... But you right, we should call it this way everywhere.

config/environments/production.rb
```diff
@@ -64,6 +64,7 @@ module OptimizerWrapper
       demo: DEMO,
       vroom: VROOM,
       ortools: ORTOOLS,
+      unconstrainted_initialization: UNCONSTRAINTED_INITIALIZATION,
```
alain-andre, line 68: The initialization of the variable is missing I think

`UNCONSTRAINTED_INITIALIZATION = Wrappers::UnconstraintedInitialization.new(tmp: TMP_DIR)`

alain-andre (approved):

PR #474, opened 2023-01-04
Add test for unconstrained_initialization + change Dockerfile

Pierre-Graber:

PR #479, opened 2023-03-20
Blackboard

Pierre-Graber:


docker/Dockerfile
```diff
@@ -20,6 +20,30 @@ ENV LANG C.UTF-8
 # Set correct environment variables.
 ENV HOME /root
 
+RUN add-apt-repository ppa:deadsnakes/ppa
```
alain-andre: Can be a lot simplest like in this [PR](https://github.com/Mapotempo/optimizer-api/pull/474/files#diff-f34da55ca08f1a30591d8b0b3e885bcc678537b2a9a4aadea4f190806b374ddc)

PR #481, opened 2023-05-03
Max value cluster

Pierre-Graber:


lib/interpreters/split_clustering.rb
```diff
@@ -1473,6 +1473,20 @@ def add_duration_from_and_to_depot(vrp, data_items)
           end
           log "matrix computed in #{(Time.now - tic).round(2)} seconds"
 
+          max_time_from_depot = 0
```
guillaumemaran, line 1476: Why only from depot and to depot ?

